PR #32, opened 2015-01-19
(build) Adding TeamCity Build Status Shield

gep13:


ferventcoder:
:+1:

ferventcoder:
Does that show the current state of the master branch only?

ferventcoder:
I fixed up the appveyor one to exclude the PRs

gep13:
Are you referring to this line of the appveyor.yml?
nuget:
  disable_publish_on_pr: true
If no, what else are you referring to?  The appveyor build is building Pull Requests at the minute.
Gary
Date: Mon, 19 Jan 2015 16:02:39 -0800

gep13:
In all honesty, I am not sure.
Based on the samples that I saw here:
http://shields.io/
And pulling from the existing shield that was on chocolatey/chocolatey it looks like this is the case, yes.

Date: Mon, 19 Jan 2015 16:01:49 -0800

gep13:
> Does that show the current state of the master branch only?

Given that the build status badge is now green, I think that this is the case, and I guess this makes sense, and it matches what the build status badge from AppVeyor is doing as well.

ferventcoder:
Sweet :)

Issue #132, opened 2015-02-28
[API] Expose Package Results through the API


ferventcoder:

Jaykul:
Do you want to change or overload the "Run()" method so it can return output? 
Maybe a generic `List<T> Run<T>()` overload? 

I mean this is going to need to include most of the commands where the user expects to see output, at least:
list, search, source list, ...

ferventcoder:
Actually it's more about adding other methods like List, Install, etc instead of just Run(). Run is fire and forget.

ferventcoder:
Unfortunately this one just broke almost every configuration manager that uses choco - https://github.com/chocolatey/choco/commit/5782a217fb576cbc01caa9d1edc7d134e5019ca9#commitcomment-13577152

PR #137, opened 2015-03-02
Start exposing lists and package results via the API #132

Jaykul:
This is mostly so you can review and provide feedback ;-)

### DIFF
--- a/src/chocolatey/GetChocolatey.cs
+++ b/src/chocolatey/GetChocolatey.cs
@@ -213,6 +213,15 @@ namespace chocolatey
 
             AssemblyFileExtractor.extract_all_resources_to_relative_directory(_container.GetInstance<IFileSystem>(), Assembly.GetAssembly(typeof(ChocolateyResourcesAssembly)), ApplicationParameters.InstallLocation, folders, ApplicationParameters.ChocolateyFileResources);
         }
+
+        public IEnumerable<T> List<T>()
+        {
+            extract_resources();
+            var configuration = create_configuration(new List<string>());
+            configuration.RegularOuptut = true;
+            var runner = new GenericRunner();
+            return runner.list<T>(configuration, _container, isConsole: false, parseArgs: null);
+        }
     }
 
     // ReSharper restore InconsistentNaming
--- a/src/chocolatey/infrastructure.app/commands/ChocolateySourceCommand.cs
+++ b/src/chocolatey/infrastructure.app/commands/ChocolateySourceCommand.cs
@@ -28,7 +28,7 @@ namespace chocolatey.infrastructure.app.commands
 
     [CommandFor(CommandNameType.sources)]
     [CommandFor(CommandNameType.source)]
-    public sealed class ChocolateySourceCommand : ICommand
+    public sealed class ChocolateySourceCommand : IListCommand<ChocolateySource>
     {
         private readonly IChocolateyConfigSettingsService _configSettingsService;
 
@@ -138,6 +138,11 @@ Chocolatey will allow you to interact with sources.
                     _configSettingsService.source_enable(configuration);
                     break;
             }
+        }
+
+        public IEnumerable<ChocolateySource> list(ChocolateyConfiguration configuration)
+        {
+            return _configSettingsService.source_list(configuration);
         }
     }
 }
--- a/src/chocolatey/infrastructure.app/configuration/ChocolateySource.cs
+++ b/src/chocolatey/infrastructure.app/configuration/ChocolateySource.cs
@@ -1,0 +1,25 @@
+﻿// Copyright © 2015 - Present RealDimensions Software, LLC
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// 
+// You may obtain a copy of the License at
+// 
+// 	http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace chocolatey.infrastructure.app.configuration
+{
+    public class ChocolateySource
+    {
+        public string Id { get; set; }
+        public string Value { get; set; }
+        public bool Disabled { get; set; }
+        public bool Authenticated { get; set; }
+    }
+}
--- a/src/chocolatey/infrastructure.app/runners/GenericRunner.cs
+++ b/src/chocolatey/infrastructure.app/runners/GenericRunner.cs
@@ -13,10 +13,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+
 namespace chocolatey.infrastructure.app.runners
 {
     using System;
     using System.Linq;
+    using System.Collections.Generic;
     using SimpleInjector;
     using adapters;
     using attributes;
@@ -28,7 +30,7 @@ namespace chocolatey.infrastructure.app.runners
 
     public sealed class GenericRunner
     {
-        public void run(ChocolateyConfiguration config, Container container, bool isConsole, Action<ICommand> parseArgs)
+        private ICommand find_command(ChocolateyConfiguration config, Container container, bool isConsole, Action<ICommand> parseArgs)
         {
             var commands = container.GetAllInstances<ICommand>();
             var command = commands.Where((c) =>
@@ -83,7 +85,7 @@ Chocolatey is not an official build (bypassed with --allow-unofficial).
  If you are seeing this message and it is not expected, your system may 
  now be in a bad state. Only official builds are to be trusted.
 "
-                         );
+                        );
 
                     }
                 }
@@ -96,14 +98,38 @@ Chocolatey is not an official build (bypassed with --allow-unofficial).
                     }
 
                     command.noop(config);
+                    return null;
                 }
-                else
+            }
+            return command;
+        }
+
+        public void run(ChocolateyConfiguration config, Container container, bool isConsole, Action<ICommand> parseArgs)
+        {
+            var command = find_command(config, container, isConsole, parseArgs);
+            if(command != null)
+            {
+                this.Log().Debug("_ {0}:{1} - Normal Run Mode _".format_with(ApplicationParameters.Name, command.GetType().Name));
+                command.run(config);
+            }
+        }
+
+        public IEnumerable<T> list<T>(ChocolateyConfiguration config, Container container, bool isConsole, Action<ICommand> parseArgs)
+        {
+            var command = find_command(config, container, isConsole, parseArgs) as IListCommand<T>;
+            if (command == null)
+            {
+                if (!string.IsNullOrWhiteSpace(config.CommandName))
                 {
-                    this.Log().Debug("_ {0}:{1} - Normal Run Mode _".format_with(ApplicationParameters.Name, command.GetType().Name));
-                    command.run(config);
+                    throw new Exception("The implementation of '{0}' does not support listing '{1}'".format_with(config.CommandName, typeof(T).Name));
                 }
+                return new List<T>();
+            }
+            else
+            {
+                this.Log().Debug("_ {0}:{1} - Normal List Mode _".format_with(ApplicationParameters.Name, command.GetType().Name));
+                return command.list(config);
             }
         }
     }
-
 }
--- a/src/chocolatey/infrastructure.app/services/ChocolateyConfigSettingsService.cs
+++ b/src/chocolatey/infrastructure.app/services/ChocolateyConfigSettingsService.cs
@@ -16,6 +16,7 @@
 namespace chocolatey.infrastructure.app.services
 {
     using System;
+    using System.Collections.Generic;
     using System.Linq;
     using configuration;
     using infrastructure.services;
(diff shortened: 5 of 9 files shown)


gep13:
@ferventcoder I gave the Travis build a kick, as it seemed to time out when restoring nuget packages

Jaykul:
Let me know what you think -- I have at least one more "list" method to add (on the list command) and probably two (on the install command). 

Also -- I stuck _your_ copyright statement on the tops of the files.  Whilst not technically correct, I don't care.

Also -- I noticed all your copyright headers say 2011 (except the three that I just added, ha).

ferventcoder:
> Also -- I stuck your copyright statement on the tops of the three files I had to add, but I noticed all your copyright headers say 2011 (except the three that I just added, ha).

They say 2011 - Present.

ferventcoder:
Let's talk about your commit messages - https://github.com/chocolatey/choco/blob/master/CONTRIBUTING.md#prepare-commits

src/chocolatey/infrastructure.app/services/ChocolateyConfigSettingsService.cs
```diff
@@ -43,14 +46,27 @@ public void noop(ChocolateyConfiguration configuration)
...
-                this.Log().Info(() => "{0}{1} - {2}".format_with(source.Id, source.Disabled ? " [Disabled]" : string.Empty, source.Value));
+                if (logResults) { 
+                    this.Log().Info(() => "{0}{1} - {2}".format_with(source.Id, source.Disabled ? " [Disabled]" : string.Empty, source.Value));
+                }
+                yield return new ChocolateySource {
+                    Id = source.Id,
+                    Value = source.Value,
+                    Disabled =  source.Disabled,
+                    Authenticated = string.IsNullOrWhiteSpace(source.Password)
+                };
```
ferventcoder: One can logresults AND return. What benefits do you gain in yielding?

src/chocolatey/infrastructure.app/services/ChocolateyConfigSettingsService.cs
```diff
@@ -43,14 +46,27 @@ public void noop(ChocolateyConfiguration configuration)
...
             }
         }
 
+        public IList<ConfigFileSourceSetting> source_list()
+        {
+            return configFileSettings.Sources.ToArray();
+        }
+
```
ferventcoder: Remove this.

src/chocolatey/infrastructure.app/commands/ChocolateySourceCommand.cs
```diff
@@ -28,7 +28,7 @@ namespace chocolatey.infrastructure.app.commands
 
     [CommandFor(CommandNameType.sources)]
     [CommandFor(CommandNameType.source)]
-    public sealed class ChocolateySourceCommand : ICommand
+    public sealed class ChocolateySourceCommand : IListCommand<ChocolateySource>
```
ferventcoder: What benefit does this change have?

ferventcoder: I mean if you implemented a separate command, what are you gaining here?

src/chocolatey/infrastructure.app/runners/ListRunner.cs
```diff
@@ -0,0 +1,113 @@
...
+                }
+
+                this.Log().Debug(() => "Configuration: {0}".format_with(config.ToString()));
+
+                if (isConsole && config.HelpRequested)
+                {
+#if DEBUG
+                    Console.WriteLine("Press enter to continue...");
+                    Console.ReadKey();
+#endif
```
ferventcoder: I'm hoping this was just because you copied over.

src/chocolatey/infrastructure/commands/IListCommand.cs
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+namespace chocolatey.infrastructure.commands
+{
+    using System.Collections.Generic;
+    using app.configuration;
+
+    public interface IListCommand<out T> : ICommand
+    {
+        IEnumerable<T> list(ChocolateyConfiguration config, bool logResults = true);
```
ferventcoder: Okay, after looking at this. I could totally see ICommand be enhanced a bit instead of introducing something new.

`IEnumerable<T> run<T>(ChocolateyConfiguration config)`

Also add `public bool LogResults { get; set; }` to the configuration and stop passing it around everywhere. ;)

src/chocolatey/infrastructure.app/services/ChocolateyConfigSettingsService.cs
Jaykul: Nothing much -- I'm outputting the members as I receive them. It's a habit from writing PowerShell pipeline stuff, and it also means I don't have to actually create an IEnumerable to return. 

It's fine to cache the new objects and output them all at once if you prefer.

src/chocolatey/infrastructure.app/commands/ChocolateySourceCommand.cs
Jaykul: This extends ICommand by adding a *list* function without forcing every command to implement list.  I could make it an additional interface instead of an inherited one, but this seemed natural to me. Is there some objection?

src/chocolatey/infrastructure.app/runners/ListRunner.cs
Jaykul: All except the last few lines of this are copy/paste. I thought about just adding stuff to the existing GenericRunner, but at the time when I did this I couldn't imagine a way to extend the configuration without creating dozens of combinations which would be incorrect...

src/chocolatey/infrastructure/commands/IListCommand.cs
Jaykul: You can't add an overload of a method where the only difference is the return value.

If I **change** the signature, then I have to modify the run method for every command to (at least) return null (or an empty list), and there will be no way of telling which ones actually return something and which don't.

Jaykul: I liked the idea of putting LogResults on the configuration, but ... Should it go on the options for this command, or the global options? Either way it ends up requiring us to rewrite other things to respect that new setting, or else the setting doesn't do anything most of the time -- both options seemed undesirable to me.

src/chocolatey/infrastructure.app/runners/ListRunner.cs
ferventcoder: Right, I just don't think you need most of the items here.

src/chocolatey/infrastructure/commands/IListCommand.cs
ferventcoder: > You can't add an overload of a method where the only difference is the return value, so you'd have to redefine run -- either way, if you put it on the ICommand instead of on a derived interface, then you have to implement it on every command even if there is no benefit.

So basically we are saying redefine run. I think by the time we are done, most of them will likely return something. Perhaps keep it separate for now though.

ferventcoder: > I liked the idea of putting LogResults on the configuration, but ... Should it go on the options for this command, or the global options?

Global Options

ferventcoder: > Either way it ends up requiring us to rewrite other things to respect that new setting, or else the setting doesn't do anything most of the time -- both options seemed undesirable to me.

There is only one or two places where this would be used right now, all related to lists. As far as what can call the setting, most folks are going to be exposed to what the command interfaces tell them they can set, not the configuration.

src/chocolatey/infrastructure.app/commands/ChocolateySourceCommand.cs
ferventcoder: I think I am starting to understand it, I'm just trying to understand it here.

src/chocolatey/infrastructure.app/commands/ChocolateySourceCommand.cs
```diff
@@ -123,7 +123,7 @@ public void run(ChocolateyConfiguration configuration)
             switch (configuration.SourceCommand.Command)
             {
                 case SourceCommandType.list:
-                    _configSettingsService.source_list(configuration);
+                    _configSettingsService.source_list(configuration, true);
```
ferventcoder: Shouldn't this just call the list function below?

src/chocolatey/infrastructure/commands/IListCommand.cs
Jaykul: >>    You can't add an overload of a method where the only difference is the return value, so you'd have to redefine run -- either way, if you put it on the ICommand instead of on a derived interface, then you have to implement it on every command even if there is no benefit.

> So basically we are saying redefine run. I think by the time we are done, most of them will likely return something. Perhaps keep it separate for now though.

I think very few of them will return anything, honestly. For instance, in the sources command we have:

* sources list
* sources add
* sources remove
* sources enable
* sources disable

I doubt any of them _other than list_ have any reason to return anything ... unless you start supporting piping objects with PowerShell's CLIXML and add a -passthru switch. ;-)

ferventcoder:
So much prettier

ferventcoder:
Okay, you can rebase and drop 524f217

src/chocolatey/infrastructure.app/configuration/ChocolateySource.cs
```diff
@@ -0,0 +1,25 @@
...
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace chocolatey.infrastructure.app.configuration
+{
+    public sealed class ChocolateySource
```
ferventcoder: Since this is presented to the outside, would you think sealed or inheritable?

src/chocolatey/infrastructure.app/services/IChocolateyConfigSettingsService.cs
```diff
@@ -16,12 +16,13 @@
...
 {
     using System;
+    using System.Collections.Generic;
     using configuration;
 
     public interface IChocolateyConfigSettingsService
     {
         void noop(ChocolateyConfiguration configuration);
-        void source_list(ChocolateyConfiguration configuration);
+        IEnumerable<ChocolateySource> source_list(ChocolateyConfiguration configuration);
```
ferventcoder: :+1:

src/chocolatey/infrastructure.app/services/ChocolateyConfigSettingsService.cs
```diff
@@ -43,12 +44,22 @@ public void noop(ChocolateyConfiguration configuration)
...
+                    this.Log().Info(() => "{0}{1} - {2}".format_with(source.Id, source.Disabled ? " [Disabled]" : string.Empty, source.Value));
+                }
+                list.Add(new ChocolateySource {
+                    Id = source.Id,
+                    Value = source.Value,
+                    Disabled =  source.Disabled,
+                    Authenticated = string.IsNullOrWhiteSpace(source.Password)
+                });
+            }
+            return list;
```
ferventcoder: :+1:

src/chocolatey/infrastructure.app/runners/GenericRunner.cs
```diff
@@ -82,27 +83,47 @@ public void run(ChocolateyConfiguration config, Container container, bool isCons
...
+
+        public void run(ChocolateyConfiguration config, Container container, bool isConsole, Action<ICommand> parseArgs)
+        {
+            var command = find_command(config, container, isConsole, parseArgs);
+            if(command != null)
+            {
+                this.Log().Debug("_ {0}:{1} - Normal Run Mode _".format_with(ApplicationParameters.Name, command.GetType().Name));
+                command.run(config);
+            }
+        }
```
ferventcoder: Need a space below this.

src/chocolatey/infrastructure.app/runners/GenericRunner.cs
```diff
@@ -82,27 +83,47 @@ public void run(ChocolateyConfiguration config, Container container, bool isCons
...
                 {
-                    this.Log().Debug("_ {0}:{1} - Normal Run Mode _".format_with(ApplicationParameters.Name, command.GetType().Name));
-                    command.run(config);
+                    throw new Exception("The implementation of '{0}' does not support listing '{1}'".format_with(config.CommandName, typeof(T).Name));
                 }
+                return new List<T>();
+            }
+            else
+            {
+                this.Log().Debug("_ {0}:{1} - Normal Run Mode _".format_with(ApplicationParameters.Name, command.GetType().Name));
```
ferventcoder: I'd go with `Normal List Mode` here.

src/chocolatey/infrastructure.app/configuration/ChocolateySource.cs
Jaykul: sure, I'll unseal it.

src/chocolatey/infrastructure.app/runners/GenericRunner.cs
Jaykul: Check.

Jaykul:
Do you want to merge this one and deal with the others separately?

src/chocolatey/GetChocolatey.cs
```diff
@@ -213,7 +213,16 @@ private void extract_resources()
...
         }
+
+        public IEnumerable<T> List<T>()
+        {
+            extract_resources();
+            var configuration = create_configuration(new List<string>());
+            configuration.RegularOuptut = true;
+            var runner = new GenericRunner();
+            return runner.list<T>(configuration, _container, isConsole: false, parseArgs: null);
+        }
```
ferventcoder: :+1:

ferventcoder: I tend to push return statements to have an empty line before. This looks good though.

ferventcoder:
Looks great! Next time use a topic branch ;)

Jaykul:
Yeah, you know ... ever since my first github pull request (when I learned that additional commits to the branch _become_ part of the pull request), I usually do that.  I don't remember why I didn't in this case, other than the whole "I'm just fixing this one thing" mentality ;-)

Issue #143, opened 2015-03-03
ToList on Nuget list command is bad, mmkay?


ferventcoder:
https://github.com/chocolatey/choco/commit/226e9982644b975678800a01e15faaf856c6128c removed yield returns so one must wait for the entire list to return before seeing output.

PR #146, opened 2015-03-04
(GH-143) Make NuGet calls yield return, so we stream output as it comes

Jaykul:
Also stop sorting things into the order they're already in anyway, because sorting blocks output.

### DIFF
--- a/src/chocolatey.tests/infrastructure.app/commands/ChocolateyPinCommandSpecs.cs
+++ b/src/chocolatey.tests/infrastructure.app/commands/ChocolateyPinCommandSpecs.cs
@@ -337,10 +337,12 @@ namespace chocolatey.tests.infrastructure.app.commands
                 configuration.Sources = ApplicationParameters.PackagesLocation;
                 configuration.ListCommand.LocalOnly = true;
                 configuration.AllVersions = true;
-                var packageResults = new ConcurrentDictionary<string, PackageResult>(); 
-                packageResults.GetOrAdd("regular", new PackageResult(package.Object, null));
-                packageResults.GetOrAdd("pinned", new PackageResult(pinnedPackage.Object, null));
-                nugetService.Setup(n => n.list_run(It.IsAny<ChocolateyConfiguration>(), false)).Returns(packageResults);
+                var packageResults = new []
+                {
+                    new PackageResult(package.Object, null),
+                    new PackageResult(pinnedPackage.Object, null)
+                };
+                nugetService.Setup(n => n.list_run(It.IsAny<ChocolateyConfiguration>(), true)).Returns(packageResults);
                 configuration.PinCommand.Command = PinCommandType.list;
             }
 
@@ -412,7 +414,7 @@ namespace chocolatey.tests.infrastructure.app.commands
                 configuration.PinCommand.Command = PinCommandType.list;
                 command.run(configuration);
 
-                nugetService.Verify(n => n.list_run(It.IsAny<ChocolateyConfiguration>(), false), Times.Once);
+                nugetService.Verify(n => n.list_run(It.IsAny<ChocolateyConfiguration>(), true), Times.Once);
             }
             
             [Pending("NuGet is killing me with extension methods. Need to find proper item to mock out to return the package object.")]
--- a/src/chocolatey/infrastructure.app/nuget/NugetList.cs
+++ b/src/chocolatey/infrastructure.app/nuget/NugetList.cs
@@ -22,7 +22,7 @@ namespace chocolatey.infrastructure.app.nuget
 
     // ReSharper disable InconsistentNaming
 
-    public sealed class NugetList
+    public static class NugetList
     {
         public static IEnumerable<IPackage> GetPackages(ChocolateyConfiguration configuration, ILogger nugetLogger)
         {
@@ -31,7 +31,8 @@ namespace chocolatey.infrastructure.app.nuget
 
             if (configuration.AllVersions)
             {
-                return results.Where(PackageExtensions.IsListed).OrderBy(p => p.Id).ToList();
+                // just trust the server order, don't sort because that's a blocking operation
+                return results.Where(PackageExtensions.IsListed);
             }
 
             if (configuration.Prerelease && packageRepository.SupportsPrereleasePackages)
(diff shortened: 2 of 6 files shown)


ferventcoder:
You would think they were in order... maybe they are again.

src/chocolatey.tests/infrastructure.app/commands/ChocolateyPinCommandSpecs.cs
```diff
@@ -337,10 +337,12 @@ public override void Context()
...
-                var packageResults = new ConcurrentDictionary<string, PackageResult>(); 
-                packageResults.GetOrAdd("regular", new PackageResult(package.Object, null));
-                packageResults.GetOrAdd("pinned", new PackageResult(pinnedPackage.Object, null));
-                nugetService.Setup(n => n.list_run(It.IsAny<ChocolateyConfiguration>(), false)).Returns(packageResults);
+                var packageResults = new []
+                {
+                    new PackageResult(package.Object, null),
+                    new PackageResult(pinnedPackage.Object, null)
+                };
+                nugetService.Setup(n => n.list_run(It.IsAny<ChocolateyConfiguration>(), true)).Returns(packageResults);
```
ferventcoder: :+1:

src/chocolatey.tests/infrastructure.app/commands/ChocolateyPinCommandSpecs.cs
```diff
@@ -412,7 +414,7 @@ public void should_call_nuget_service_list_run_when_command_is_list()
                 configuration.PinCommand.Command = PinCommandType.list;
                 command.run(configuration);
 
-                nugetService.Verify(n => n.list_run(It.IsAny<ChocolateyConfiguration>(), false), Times.Once);
+                nugetService.Verify(n => n.list_run(It.IsAny<ChocolateyConfiguration>(), true), Times.Once);
```
ferventcoder: ?

src/chocolatey/infrastructure.app/nuget/NugetList.cs
```diff
@@ -43,11 +44,12 @@ public static IEnumerable<IPackage> GetPackages(ChocolateyConfiguration configur
                 results = results.Where(p => p.IsLatestVersion);
             }
 
-            return results.OrderBy(p => p.Id)
-                          .AsEnumerable()
-                          .Where(PackageExtensions.IsListed)
-                          .Where(p => configuration.Prerelease || p.IsReleaseVersion())
-                          .distinct_last(PackageEqualityComparer.Id, PackageComparer.Version).ToList();
```
ferventcoder: No idea but the nuget.exe uses the distinct_last, that's where I got that from.

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -71,36 +71,31 @@ public void list_noop(ChocolateyConfiguration config)
...
+        public IEnumerable<PackageResult> list_run(ChocolateyConfiguration config, bool logResults)
         {
-            var packageResults = new ConcurrentDictionary<string, PackageResult>();
-
-            var packages = NugetList.GetPackages(config, _nugetLogger).ToList();
-
-            foreach (var package in packages.or_empty_list_if_null())
+            foreach (var package in NugetList.GetPackages(config, _nugetLogger))
             {
+                var pkg = package; // for lamda access
```
ferventcoder: Actually I'd prefer that you use pkg in the for each loop and set this `var package = pkg`

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -71,36 +71,31 @@ public void list_noop(ChocolateyConfiguration config)
...
                     }
                 }
                 else
                 {
-                    this.Log().Debug(() => "{0} {1}".format_with(package.Id, package.Version.to_string()));
+                    this.Log().Debug(() => "{0} {1}".format_with(pkg.Id, pkg.Version.to_string()));
                 }
 
-                packageResults.GetOrAdd(package.Id, new PackageResult(package, null));
+                yield return new PackageResult(pkg, null);
```
ferventcoder: This could produce really interesting results in any commands that use "all"

ferventcoder: Like `upgrade all` - I see that as needing a `.ToList()` on it - I had some interesting results before whatever changes I made - it may not need the change.

src/chocolatey.tests/infrastructure.app/commands/ChocolateyPinCommandSpecs.cs
Jaykul: Aw foey. You know, I should have double-checked how that was, originally.  In the process of the edits I _removed_ that boolean parameter in favor of the new `config.LogOutput` property, then decided I should do that later, separately.

I think that's going to be a problem, because I'm pretty sure the tests didn't pass with it set to false. I'll investigate.

src/chocolatey/infrastructure.app/nuget/NugetList.cs
Jaykul: Well, I'm pretty sure you can do the same thing with the sort/group/first, but the output appears to be correct without the distinct_last and I honestly think it would be a big problem if more than one item in the feed was marked as `IsLatestVersion` so I don't see the point ;-)

Maybe the original (nuget) code was just paranoid...

ferventcoder: Perhaps it was?

src/chocolatey/infrastructure.app/services/NugetService.cs
Jaykul: Fair enough, I didn't even think about it.

ferventcoder: Ping @Jaykul

src/chocolatey.tests/infrastructure.app/commands/ChocolateyPinCommandSpecs.cs
ferventcoder: What did you come up with here?

ferventcoder: Ping.

Jaykul: This one I actually know, now.  After the changes to make things stream output, I had to set this to TRUE so the results would get enumerated (and the method under test would get called).
Of course, it goes away in the next commit, because we got rid of the boolean and moved it to the super-all-knowing-config-object.

src/chocolatey/infrastructure.app/services/NugetService.cs
Jaykul: I'm so confused -- I'd forgotten all about this pull request -- did I resubmit all this last night? I *definitely did* change this.

src/chocolatey/infrastructure.app/services/NugetService.cs
Jaykul: Yeah, any call to the LIST method MUST enumerate the results, or else nothing happens (and hey, it happens REALLY, REALLY FAST!)

Jaykul:
Let's toss this one and use #224 (ok?)

PR #196, opened 2015-03-23
Added Remove-BinFile Helper

### LINKED ISSUE #195: Add Remove-BinFile
Now that we have decided to add Generate-BinFile to Choco vNext, we really need to add the corresponding Remove-BinFile method.

gep13:
Previously, Chocolatey had an internal helper function
called Remove-BinFile.  While this was never part of the
public API, many packages take a dependency on it.

Without this change, a number of packages will fail to
install.

Closes #195


gep13:
@ferventcoder not sure exactly "why" the whitespace changes appeared when I edited the csproj file.  I am guessing that this was EditorConfig playing with the file.  I only put the additional EmbeddedResource into the same itemgroup, as Visual Studio was putting it into it's own one, I thought it made same to put them in the same one.

Issue #181, opened 2015-03-20
When prompting for a user yes/no answer, use a short [y/n] representation


christianrondeau:
Here is how yes/no/skip answers are displayed:

```bash
Do you want to run the script?
NOTE: If you choose not to run the script, the installation will
fail.
Skip is an advanced option and most likely will never be wanted.

1) yes
2) no [Default - Press Enter]
3) skip
```

Here is a suggestion of how the prompt could look like:

```bash
Do you want to run the script? (Yes/No/Skip): y
```

In other words: when asking for a yes/no question, use the shorthand, similar to other windows command line tools.

Example:

* xcopy: `Overwrite C:\path\file.txt (Yes/No/All)?`

For more information: https://groups.google.com/forum/#!topic/chocolatey/UVncL7PxXRg

(Moved from https://github.com/chocolatey/chocolatey/issues/689)

christianrondeau:
Hi,

This has been almost fully implemented, but a single detail. See: https://github.com/christianrondeau/choco/blob/22af53bac95cd701cfcc81c7a2a6ef39ceb4cfbb/src/chocolatey/infrastructure/commandline/InteractivePrompt.cs#L58 (Line 58 and 60)

There is a `Log().Info()`, and then a `Console.ReadLine`. That results in two lines. I don't see anywhere in the stack trace a line break, it seems to be within `log4net` that this happens... so in order to achieve what I'm trying to do here, the only way I see I can do it is by:

1. Either I bypass `log4net`, and output the prompt directly to the console, and/or;
2. I then output both the prompt AND the answer in `log4net`; to do this, I'll have to be able to disable console output for that second line (which might be messy), and/or;
3. We skip completely the logging of prompts, and instead we make sure to include in the line *after* the prompt the selection (e.g. `Are you sure? (yes/no): n` then `Selection: no`)

If you know a better way to achieve this, let me know.

christianrondeau:
Well it looks like it's impossible with log4net.

I will then do the direct to console approach, and output the selection after, i.e. (simplified):

Console.Write("prompt: ");
String result = Console.ReadLine();
Log().Info("You have selected " + result");

This means that either 1. We don't see the prompt message in the log, or b. We see it twice in the console.

Unless we create a "prompt" logger, which would behave like the normal logger, without the console appender. Is this worth the additional complexity?

ferventcoder:
maybe...

christianrondeau:
[...]

PR #25, opened 2015-01-18
Add a Gitter chat badge to README.md

gitter-badger:
### chocolatey/choco now has a Chat Room on Gitter

@gep13 has just created a chat room. You can visit it here: [https://gitter.im/chocolatey/choco](https://gitter.im/chocolatey/choco?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&content=body_link).

This pull-request adds this badge to your README.md:

[Gitter](https://gitter.im/chocolatey/choco?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=body_badge)

Happy chatting.

PS: [Click here](https://gitter.im/settings/badger/opt-out) if you would prefer not to receive automatic pull-requests from Gitter in future.

PR #199, opened 2015-03-24
#185 Remove default choice from `InteractivePrompt.prompt_for_confirmation` and all associated tests

### LINKED ISSUE #185: Do not offer a default option when prompting for a user choice
I suggest not offering a default option when offering choices to the end user. This will avoid cases when typing too fast ends up with <kbd>Enter</kbd> is not being preceeded by the desired number. It also shortens the output and makes less stuff to read.

There are currently at least two cases where a default option is offered.

``` bash
Which version of vagrant would you like to uninstall?
 1) None [Default - Press Enter]
 2) 1.6.5
 3) All versions
```

Also:

``` bash
Do you want to run the script?
 NOTE: If you choose not to run the script, the installation will
 fail.
 Skip is an advanced option and most likely will never be wanted.

 1) yes
 2) no [Default - Press Enter]
 3) skip
```

For more details: https://groups.google.com/forum/#!topic/chocolatey/UVncL7PxXRg

(Moved from https://github.com/chocolatey/chocolatey/issues/693)

christianrondeau:
Resolves #185 (Do not offer a default option when prompting for a user choice).

Note that it simply removes the code in `interactivePrompt` for `defaultValue` (first commit) and `requireValue` (second commit) since nowhere in the code other than in tests was the value actually optional. Most of the changes other than that are tests that don't make sense anymore (e.g. all tests for optional/default prompts).

This is my first pull request for this project, so feel free to suggest changes to how I make these (e.g. leave all tests and ask), since I plan on doing a few more. All tests are green!

### DIFF
--- a/src/chocolatey.console/Program.cs
+++ b/src/chocolatey.console/Program.cs
@@ -194,7 +194,7 @@ Chocolatey detected you are not running from an elevated command shell
  elevated shell. When you open the command shell, you should ensure 
  that you do so with ""Run as Administrator"" selected.
 
- Do you want to continue?", new[] {"yes", "no"}, "no", requireAnswer: true);
+ Do you want to continue?", new[] { "yes", "no" }, defaultChoice: null, requireAnswer: true);
 
                 if (selection.is_equal_to("no"))
                 {
--- a/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
+++ b/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
@@ -618,7 +618,7 @@ ATTENTION: You must take manual action to remove {1} from
             var rollback = true;
             if (config.PromptForConfirmation)
             {
-                var selection = InteractivePrompt.prompt_for_confirmation(" Unsuccessful operation for {0}.{1}  Do you want to rollback to previous version (package files only)?".format_with(packageResult.Name, Environment.NewLine), new[] { "yes", "no" }, "yes", requireAnswer: true);
+                var selection = InteractivePrompt.prompt_for_confirmation(" Unsuccessful operation for {0}.{1}  Do you want to rollback to previous version (package files only)?".format_with(packageResult.Name, Environment.NewLine), new[] { "yes", "no" }, defaultChoice: null, requireAnswer: true);
                 if (selection.is_equal_to("no")) rollback = false;
             }
 
--- a/src/chocolatey/infrastructure.app/services/NugetService.cs
+++ b/src/chocolatey/infrastructure.app/services/NugetService.cs
@@ -816,7 +816,7 @@ packages as of version 1.0.0. That is what the install command is for.
                             choices.Add(allVersionsChoice);
                         }
 
-                        var selection = InteractivePrompt.prompt_for_confirmation("Which version of {0} would you like to uninstall?".format_with(packageName), choices, abortChoice, true);
+                        var selection = InteractivePrompt.prompt_for_confirmation("Which version of {0} would you like to uninstall?".format_with(packageName), choices, defaultChoice: null, requireAnswer: true);
 
                         if (string.IsNullOrWhiteSpace(selection)) continue;
                         if (selection.is_equal_to(abortChoice)) continue;
--- a/src/chocolatey/infrastructure.app/services/PowershellService.cs
+++ b/src/chocolatey/infrastructure.app/services/PowershellService.cs
@@ -210,7 +210,7 @@ Do you want to run the script?
  fail.
  Skip is an advanced option and most likely will never be wanted.
 "
-                                                 , new[] {"yes", "no", "skip"}, "no", requireAnswer: true);
+                                                 , new[] { "yes", "no", "skip" }, defaultChoice: null, requireAnswer: true);
                     if (selection.is_equal_to("yes")) shouldRun = true;
                     if (selection.is_equal_to("no"))
                     {


gep13:
@christianrondeau can I get you to have a read of our Contributing document, specifically this section:

https://github.com/chocolatey/choco/blob/master/CONTRIBUTING.md#prepare-commits

Thanks

christianrondeau:
@gep13 I did, and carefully :) I was unsure of my pull request especially because of all removed associated tests. I guess this is why you are referring me to the `CONTRIBUTING.md` file.

Since the issue was about removing the defaults everywhere, I thought of removing the *functionality* instead of changing only usages; many tests were written specifically to handle defaults; I did a minimal cleanup (which happens to be quite a lot). I also saw that required (which is associated with defaults) was actually not used, and was also phantom code. I made a separate commit for that, which can be rolled back.

So, I had two choices: either I only changed the *calls* to `prompt_for_confirmation` and passed `null` everywhere (leaving lots phantom code and tests), or I cleaned up after myself and removed code that was not used anymore.

I usually preferred the latter, but if you prefer the former, I can certainly reissue a pull request and do that in the future. Let me know!

gep13:
The main thing that I was referring to was this:

>The first line of the commit message should be a short description around 50 characters in length and be prefixed with the GitHub issue it refers to with parentheses surrounding that. If the GitHub issue is #25, you should have (GH-25) prefixed to the message.

As I noticed that the commit message was wrapping around:

image

I am not yet clear on how stringent @ferventcoder is going to be about things like this :smile_cat:

gep13:
I hadn't actually looked at the contents of the PR from a review perspective, I was just trying to pre-empt a possible issue that you might be called on :smile_cat:

ferventcoder:
My biggest issue is the removing of code from infrastructure. The infrastructure section is general purpose items that could be pulled out and into other things or their own things.
Infrastructure.app is application-specific, tailored things for the application in question, where these specific changes would have made more sense to destroy features.

I'd rather see adding additional items to the general purpose area. Almost all of the work here could have been done by simple changes to the parameters passed.

One could add an application specific prompt setup that the app would call instead that will call the general purpose one with the specific setup, if that makes sense.

christianrondeau:
@gep: Thanks, I'll re-issue the commits with shorter names.

@ferventcoder: So you are saying: keep the "unused" code, as it is infrastructure code and may be used for other purposes in the future. Is this correct? I would not create an application-specific wrapper just for this then, I'd simply pass `null` to default value and `false` to required, it will make a much smaller commit :)

I'll try to avoid "cleaning up" after myself in future pull requests when the code is in the `infrastructure` section, but still do it for code in the app-specific portion of the code. Let me know if I misunderstood anything, let's make this "simple" PR perfect before I start piling up other PRs :)

PR coming up in a few hours (when I have some free time)

ferventcoder:
> So you are saying: keep the "unused" code, as it is infrastructure code and may be used for other purposes in the future. Is this correct? I would not create an application-specific wrapper just for this then, I'd simply pass null to default value and false to required, it will make a much smaller commit :)

Yes. I think you'd still want `required: true`, as that is what it was before and that's how you were setting it up with the code removal.

christianrondeau:
@ferventcoder Correct for `required`, that's what I meant - I typed too fast :)

christianrondeau:
Here is another (much simpler) pull request. Note that because there was no tests for default options, I didn't write any. I did verify it worked on a build though :)

I kept the two commits with the tests cleanup, just in case: https://github.com/christianrondeau/choco/tree/185-no-prompt-default-failed-pr-1

Let me know if that's better. If you'd prefer that I add tests, let me know.

ferventcoder:
Always prefer tests :)

On Tuesday, March 24, 2015, Christian Rondeau <notifications@github.com>
wrote:

> Here is another (much simpler) pull request. Note that because there was
> no tests for default options, I didn't write any. I did verify it worked on
> a build though :)
>
> I kept the two commits with the tests cleanup, just in case:
> https://github.com/christianrondeau/choco/tree/185-no-prompt-default-failed-pr-1
>
> Let me know if that's better. If you'd prefer that I add tests, let me
> know.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/chocolatey/choco/pull/199#issuecomment-85769325>.
>


-- 

Rob
"Be passionate in all you do"

http://devlicio.us/blogs/rob_reynolds
http://ferventcoder.com
http://twitter.com/ferventcoder

christianrondeau:
@ferventcoder I guessed so, I do too :+1: 

Though in that case I'm not sure how to proceed. Since there was *no* test verifying that the "default" did indeed work, and since I only *removed* a feature, not only I have no new tests to write, but I also have no tests to update or remove :stuck_out_tongue:. The behavior of `null` default choice is already tested... so, what's your take on this?

ferventcoder:
I didn't have any tests verifying using the default? Gah, happens when you
are on a deadline - technical debt. :/

On Tuesday, March 24, 2015, Christian Rondeau <notifications@github.com>
wrote:

> @ferventcoder <https://github.com/ferventcoder> I guessed so, I do too [image:
> :+1:]
>
> Though in that case I'm not sure how to proceed. Since there was *no*
> test verifying that the "default" did indeed work, and since I only
> *removed* a feature, not only I have no new tests to write, but I also
> have no tests to update or remove [image: :stuck_out_tongue:]. The
> behavior of null default choice is already tested... so, what's your take
> on this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/chocolatey/choco/pull/199#issuecomment-85775207>.
>


-- 

Rob
"Be passionate in all you do"

http://devlicio.us/blogs/rob_reynolds
http://ferventcoder.com
http://twitter.com/ferventcoder

src/chocolatey.console/Program.cs
```diff
@@ -194,7 +194,7 @@ private static void warn_when_admin_needs_elevation(ChocolateyConfiguration conf
  elevated shell. When you open the command shell, you should ensure 
  that you do so with ""Run as Administrator"" selected.
 
- Do you want to continue?", new[] {"yes", "no"}, "no", requireAnswer: true);
+ Do you want to continue?", new[] { "yes", "no" }, defaultChoice: null, requireAnswer: true);
```
ferventcoder: :+1:

src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
```diff
@@ -618,7 +618,7 @@ private void rollback_previous_version(ChocolateyConfiguration config, PackageRe
             var rollback = true;
             if (config.PromptForConfirmation)
             {
-                var selection = InteractivePrompt.prompt_for_confirmation(" Unsuccessful operation for {0}.{1}  Do you want to rollback to previous version (package files only)?".format_with(packageResult.Name, Environment.NewLine), new[] { "yes", "no" }, "yes", requireAnswer: true);
+                var selection = InteractivePrompt.prompt_for_confirmation(" Unsuccessful operation for {0}.{1}  Do you want to rollback to previous version (package files only)?".format_with(packageResult.Name, Environment.NewLine), new[] { "yes", "no" }, defaultChoice: null, requireAnswer: true);
```
ferventcoder: I almost prefer this one to keep the default choice.

ferventcoder:
I'm +1 on merging this, with possibly fixing the rollback prompt. That's one that seems like it would have a default choice. Thoughts?

christianrondeau:
My humble opinion would be to avoid defaults altogether. Here is my rationale:

1. If you're asking something, require a response. In the end, you still have to type something, whether it's just <kbd>Enter</kbd> or <kbd>y</kbd> <kbd>Enter</kbd>
2. Personally, I see two choices: rollback (and do something I might not understand/trust), or don't rollback (and let me repair manually). I would probably choose not to rollback, especially when failing an update. Since it's not a clear-cut choice, a default may not make sense.
3. If however rollback should be the de facto choice, then it should not be an option at all.
4. The prompt #181 will shorten the display. I'm not sure how to clearly identify a default option without adding noise to the prompt message.

This being said, those are my only arguments :smile:. If you still believe the default value to be preferable, let me know and I'll change the PR to bring back the default option for this specific case.

ferventcoder:
Merged into stable at https://github.com/chocolatey/choco/commit/e992c4f5ad489330561cbe3127e8fd8f1c133142 and will be released in 0.9.9.5.

PR #224, opened 2015-04-05
(GH-143) Make NuGet calls yield return, so we stream output as it comes

Jaykul:
Fix NuGet so it streams output.

WARNING NOTE: This makes it VERY IMPORTANT that SOMEONE must enumerate the results, or NOTHING will happen. ;-)

### DIFF
--- a/src/chocolatey.tests/infrastructure.app/commands/ChocolateyPinCommandSpecs.cs
+++ b/src/chocolatey.tests/infrastructure.app/commands/ChocolateyPinCommandSpecs.cs
@@ -337,10 +337,12 @@ namespace chocolatey.tests.infrastructure.app.commands
                 configuration.Sources = ApplicationParameters.PackagesLocation;
                 configuration.ListCommand.LocalOnly = true;
                 configuration.AllVersions = true;
-                var packageResults = new ConcurrentDictionary<string, PackageResult>(); 
-                packageResults.GetOrAdd("regular", new PackageResult(package.Object, null));
-                packageResults.GetOrAdd("pinned", new PackageResult(pinnedPackage.Object, null));
-                nugetService.Setup(n => n.list_run(It.IsAny<ChocolateyConfiguration>(), false)).Returns(packageResults);
+                var packageResults = new []
+                {
+                    new PackageResult(package.Object, null),
+                    new PackageResult(pinnedPackage.Object, null)
+                };
+                nugetService.Setup(n => n.list_run(It.IsAny<ChocolateyConfiguration>(), true)).Returns(packageResults);
                 configuration.PinCommand.Command = PinCommandType.list;
             }
 
@@ -412,7 +414,7 @@ namespace chocolatey.tests.infrastructure.app.commands
                 configuration.PinCommand.Command = PinCommandType.list;
                 command.run(configuration);
 
-                nugetService.Verify(n => n.list_run(It.IsAny<ChocolateyConfiguration>(), false), Times.Once);
+                nugetService.Verify(n => n.list_run(It.IsAny<ChocolateyConfiguration>(), true), Times.Once);
             }
             
             [Pending("NuGet is killing me with extension methods. Need to find proper item to mock out to return the package object.")]
--- a/src/chocolatey/infrastructure.app/nuget/NugetList.cs
+++ b/src/chocolatey/infrastructure.app/nuget/NugetList.cs
@@ -22,7 +22,7 @@ namespace chocolatey.infrastructure.app.nuget
 
     // ReSharper disable InconsistentNaming
 
-    public sealed class NugetList
+    public static class NugetList
     {
         public static IEnumerable<IPackage> GetPackages(ChocolateyConfiguration configuration, ILogger nugetLogger)
         {
@@ -31,7 +31,8 @@ namespace chocolatey.infrastructure.app.nuget
 
             if (configuration.AllVersions)
             {
-                return results.Where(PackageExtensions.IsListed).OrderBy(p => p.Id).ToList();
+                // just trust the server order, don't sort because that's a blocking operation
+                return results.Where(PackageExtensions.IsListed);
             }
 
             if (configuration.Prerelease && packageRepository.SupportsPrereleasePackages)
@@ -43,11 +44,12 @@ namespace chocolatey.infrastructure.app.nuget
                 results = results.Where(p => p.IsLatestVersion);
             }
 
-            return results.OrderBy(p => p.Id)
-                          .AsEnumerable()
-                          .Where(PackageExtensions.IsListed)
-                          .Where(p => configuration.Prerelease || p.IsReleaseVersion())
-                          .distinct_last(PackageEqualityComparer.Id, PackageComparer.Version).ToList();
+            // just trust the server order, don't sort because that's a blocking operation
+            // also don't worry about multiple versions, considering Is*LatestVersion only applies to one
+            return results.Where(PackageExtensions.IsListed)
+                          .Where(p => configuration.Prerelease || p.IsReleaseVersion());
+                       // .OrderBy(p => p.Id).ThenByDescending(p => p.Version)
+                       // .GroupBy(p => p.Id).Select(g => g.First());
         }
     }
 
--- a/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
+++ b/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
@@ -80,7 +80,7 @@ namespace chocolatey.infrastructure.app.services
                 var list = _nugetService.list_run(config, logResults: true);
                 if (config.RegularOutput)
                 {
-                    this.Log().Warn(() => @"{0} packages {1}.".format_with(list.Count, config.ListCommand.LocalOnly ? "installed" : "found"));
+                    this.Log().Warn(() => @"{0} packages {1}.".format_with(list.Count(), config.ListCommand.LocalOnly ? "installed" : "found"));
 
                     if (config.ListCommand.LocalOnly && config.ListCommand.IncludeRegistryPrograms)
                     {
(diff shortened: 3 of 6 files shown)


ferventcoder:
Does this supersede #146 ?

Jaykul:
They are in order, each one should have stacked, but I think they ACTUALLY overlap. (sorry much!)

Jaykul:
What are we waiting for on this, @ferventcoder?

src/chocolatey.tests/infrastructure.app/commands/ChocolateyPinCommandSpecs.cs
```diff
@@ -412,7 +414,7 @@ public void should_call_nuget_service_list_run_when_command_is_list()
                 configuration.PinCommand.Command = PinCommandType.list;
                 command.run(configuration);
 
-                nugetService.Verify(n => n.list_run(It.IsAny<ChocolateyConfiguration>(), false), Times.Once);
+                nugetService.Verify(n => n.list_run(It.IsAny<ChocolateyConfiguration>(), true), Times.Once);
```
ferventcoder: same question here. What changed this from false to true.

src/chocolatey/infrastructure.app/nuget/NugetList.cs
```diff
@@ -31,7 +31,8 @@ public static IEnumerable<IPackage> GetPackages(ChocolateyConfiguration configur
 
             if (configuration.AllVersions)
             {
-                return results.Where(PackageExtensions.IsListed).OrderBy(p => p.Id).ToList();
+                // just trust the server order, don't sort because that's a blocking operation
+                return results.Where(PackageExtensions.IsListed);
```
ferventcoder: How many different ways did you test this?

ferventcoder: Based on that I mean with what diff types of sources?

src/chocolatey/infrastructure.app/nuget/NugetList.cs
```diff
@@ -43,11 +44,12 @@ public static IEnumerable<IPackage> GetPackages(ChocolateyConfiguration configur
                 results = results.Where(p => p.IsLatestVersion);
             }
 
-            return results.OrderBy(p => p.Id)
-                          .AsEnumerable()
-                          .Where(PackageExtensions.IsListed)
-                          .Where(p => configuration.Prerelease || p.IsReleaseVersion())
-                          .distinct_last(PackageEqualityComparer.Id, PackageComparer.Version).ToList();
+            // just trust the server order, don't sort because that's a blocking operation
+            // also don't worry about multiple versions, considering Is*LatestVersion only applies to one
```
ferventcoder: The problem here is when it is a file folder list. If you have multiple versions in that source, none of them consider themselves the latest either. That's why the sorting and grabbing distinct last.

src/chocolatey.tests/infrastructure.app/commands/ChocolateyPinCommandSpecs.cs
Jaykul: Same answer: The test is "if I call X does Y get called" ....

After the changes to make things stream output, you have to set this to TRUE so the results would get enumerated (and the method under test would get called). Of course, this parameter *goes away* in the next commit, because we got rid of the boolean and moved it to the super-all-knowing-config-object.

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -71,36 +71,31 @@ public void list_noop(ChocolateyConfiguration config)
...
+        public IEnumerable<PackageResult> list_run(ChocolateyConfiguration config, bool logResults)
         {
-            var packageResults = new ConcurrentDictionary<string, PackageResult>();
-
-            var packages = NugetList.GetPackages(config, _nugetLogger).ToList();
-
-            foreach (var package in packages.or_empty_list_if_null())
+            foreach (var package in NugetList.GetPackages(config, _nugetLogger))
             {
+                var pkg = package; // for lamda access
```
ferventcoder: My thoughts about switching `package` to `pkg` on line 76 and making this lamdba-safe var named `package` still apply here.

src/chocolatey.tests/infrastructure.app/commands/ChocolateyPinCommandSpecs.cs
ferventcoder: Ah true.

src/chocolatey/infrastructure.app/nuget/NugetList.cs
Jaykul: Well, to be honest, Chocolatey and Nuget. However, the determination I made is that it's better to output in "source order" fast than to output in "Id Order" much, much, slower (not to mention memory impact). 

Point being that if we _do_ need to sort for certain sources,  we should do so MUCH further back up the stack (e.g. in the apps, where we enumerate the collection, not in the core library).

src/chocolatey/infrastructure.app/services/NugetService.cs
Jaykul: Yeah, I made that change in the next pull request which touches these files :-( I tried and tried to get move that change into this commit but I couldn't get git/diff to cooperate.

src/chocolatey/infrastructure.app/nuget/NugetList.cs
Jaykul: Hrm. Well, we have to either:
1. Do that in the apps which call this API, or 
2. Slow down the results for all callers.

Jaykul:
Perhaps we should add a switch to the API (config) like `StreamNuGetSearchResults` which defaults to false, and if it's false, do the sorting and filtering in the API (as before), and only stream the results unsorted when someone explicitly asks for it?

src/chocolatey/infrastructure.app/nuget/NugetList.cs
ferventcoder: I'm not sure it actually does slow the thing down, but what do I know. If it works in Nuget just fine, it should also work here as the code was written looking at what they are doing.

ferventcoder:
@Jaykul yes perhaps. I want to get this thing moving along. I just want to make sure we test all avenues. I think we'll want to add an integration spec for each scenario and see what happens.

RichiCoder1:
:+1: for getting this moving along ;)

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -896,13 +891,11 @@ private void set_package_names_if_all_is_specified(ChocolateyConfiguration confi
                 var input = config.Input;
                 config.Input = string.Empty;
 
-                var localPackages = list_run(config, logResults: false);
-
+                config.PackageNames = list_run(config, false).Select(p => p.Name).@join(ApplicationParameters.PackageNamesSeparator);
```
ferventcoder: Good, enumerate. :+1:

src/chocolatey/infrastructure.app/nuget/NugetList.cs
ferventcoder: Something is a bit weird now. I am getting all versions of local packages.

ferventcoder: I'm trying to determine what it is, but I think the distinct_last is what ensures only one if not all versions

ferventcoder: Yep, that was it. Works like a dream with it back in there.

ferventcoder: Maybe.

ferventcoder:
Okay, this is completely weird. Let's look at the two scenarios we have. One is with `OrderById` and `distinct_last`, we'll call that one equal to NuGet.CommandLine scenario, and the other is this PR as is.

For memory/time consumption, I ran against the full list from chocolatey.org. The actual impact of time was negligent. Both took around 20 seconds total to output the results. It was about 2 seconds slower to implement OrderBy (but I didn't see it as a blocking operation, unsure why not). However the number of packages was different - I came up with 2780-ish packages with NuGet.Commandline adjustments and only 2172-ish with the PR as is.

The time to first result from query was also kind of interesting. In the NuGet.CommandLine scenario, it's about 500 milliseconds to first result. In the original PR scenario, it doesn't appear to stream appropriately. I'm guessing the difference here has to do with logging the output. 

@Jaykul what kind of results where you seeing?

ferventcoder:
Merged into stable at https://github.com/chocolatey/choco/commit/73ba21f0caf181544843ee97a6e10b627bfb6bf6 and master at https://github.com/chocolatey/choco/commit/79ed3af65aeb7bbcbe500c64036a6f3a6401ca60

src/chocolatey/infrastructure.app/nuget/NugetList.cs
Jaykul: LOL -- yes, that is what ensures only one not all (for file based repos)

It's a blocking filter, which means that when you're calling web API based repositories (where it's not necessary), you're going to have to wait for ALL results to be received and parsed before you can output any results. It's dramatically slower to first output (and thus _feels_ a lot slower), but **critical*** for file folder repositories.

ferventcoder: I'm not sure that it turned out to not stream. But I'm not testing from the API perspective.

ferventcoder: If that is true about the all, it only take 2 seconds to do so from chocolatey.org

Jaykul: Well, distinct_last doesn't inherently block streaming: it returns the last element of one set as soon as it sees the first element of the next set -- but it requires sorting. Unless the sorting is done server side, you have to block to sort.

RichiCoder1: Nuget Server, iirc, uses OData under the covers, which does support server side sorting. Not sure if chocolatey preserves this detail, or if the chocolatey lib uses it.

ferventcoder: yes, it preserves

Jaykul: NuGet Server _currently_ uses OData under the covers, but it's only one possible provider. The FileSystem implementation doesn't, for instance (although it's presumably so much faster nobody cares).  Anyway. Like I said, totally up to you.

ferventcoder: Currently, yes there is that. :)

src/chocolatey/infrastructure.app/services/NugetService.cs
ste10k41: 447928594327

src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
```diff
@@ -80,7 +80,7 @@ public void list_run(ChocolateyConfiguration config, bool logResults)
                 var list = _nugetService.list_run(config, logResults: true);
                 if (config.RegularOutput)
                 {
-                    this.Log().Warn(() => @"{0} packages {1}.".format_with(list.Count, config.ListCommand.LocalOnly ? "installed" : "found"));
```
ste10k41: Leave a line 📝

PR #273, opened 2015-05-07
(GH-274) Allow unattended installs as non-admin user

### LINKED ISSUE #274: Unattended installation of Chocolatey without admin rights aren't possible.
`chocoInstall` calls `choco unpackself` which [throws up a prompt if the shell is not elevated or running as an admin user](https://github.com/chocolatey/choco/blob/3f04fc06ecd59948183ebf0e20a6e7abff0007c4/src/chocolatey/infrastructure.app/runners/GenericRunner.cs#L148).

This blocks the installation on user input, which makes unattended installations without admin rights impossible at the moment (which is important for CI like environments or scripts).

smspillaz:
Previously, if a non-admin user attempted to install chocolatey,
a message indicating that non-admin installs were for advanced
users only would appear and block the installation.

This runs contrary to chocolatey policy, so just pass -y during
the unpackself command. The installation should never block.

This closes #274

### DIFF
--- a/nuget/chocolatey/tools/chocolateysetup.psm1
+++ b/nuget/chocolatey/tools/chocolateysetup.psm1
@@ -4,7 +4,7 @@ $sysDrive = $env:SystemDrive
 $tempDir = $env:TEMP
 $defaultChocolateyPathOld = "$sysDrive\Chocolatey"
 #$ErrorActionPreference = 'Stop'
-$debugMode = $false
+$debugModeParams = ""
 
 function Initialize-Chocolatey {
 <#
@@ -34,17 +34,13 @@ param(
   Write-Debug "Initialize-Chocolatey"
 
   if ($env:ChocolateyEnvironmentDebug -eq 'true') {
-    $debugMode = $true
+    $debugModeParams = "dv"
   }
 
   Install-DotNet4IfMissing
 
   $chocoNew = Join-Path $thisScriptFolder 'chocolateyInstall\choco.exe'
-  if ($debugMode) {
-    & $chocoNew unpackself -fdvy
-  } else {
-    & $chocoNew unpackself -fy
-  }
+  & $chocoNew unpackself -fy$debugModeParams
 
   $installModule = Join-Path $thisScriptFolder 'chocolateyInstall\helpers\chocolateyInstaller.psm1'
   Import-Module $installModule -Force
@@ -313,11 +309,7 @@ param(
   $chocoExeDest = Join-Path $chocolateyPath 'choco.exe'
   Copy-Item $chocoExe $chocoExeDest -force
 
-  if ($debugMode) {
-    & $chocoExeDest unpackself -fdvy
-  } else {
-    & $chocoExeDest unpackself -fy
-  }
+  & $chocoExeDest unpackself -fy$debugModeParams
 }
 
 function Ensure-ChocolateyLibFiles {
--- a/nuget/chocolatey/tools/chocolateysetup.psm1
+++ b/nuget/chocolatey/tools/chocolateysetup.psm1
@@ -41,9 +41,9 @@ param(
 
   $chocoNew = Join-Path $thisScriptFolder 'chocolateyInstall\choco.exe'
   if ($debugMode) {
-    & $chocoNew unpackself -fdv
+    & $chocoNew unpackself -fdvy
   } else {
-    & $chocoNew unpackself -f
+    & $chocoNew unpackself -fy
   }
 
   $installModule = Join-Path $thisScriptFolder 'chocolateyInstall\helpers\chocolateyInstaller.psm1'
@@ -314,9 +314,9 @@ param(
   Copy-Item $chocoExe $chocoExeDest -force
 
   if ($debugMode) {
-    & $chocoExeDest unpackself -fdv
+    & $chocoExeDest unpackself -fdvy
   } else {
-    & $chocoExeDest unpackself -f
+    & $chocoExeDest unpackself -fy
   }
 }
 


ferventcoder:
Howdy - have you had a chance to review CONTRIBUTING.md?

ferventcoder:
What version of choco were you running before? We've removed the admin requirement for a bunch of commands, including unpackself.

smspillaz:
@ferventcoder Hey, I had a brief look - my understanding was that PRs were fine for small changes. Was there anything else I needed to do?

As for which version I'm running - I'm running 0.9.9.5.

I think the main problem here isn't so much the admin requirement as much as the interactive prompt that the command throws up. It blocks fully automated installations, which is important for a continuous integration script I'm working on.

ferventcoder:
Then I think it should pass `-y` by default, don't you?

ferventcoder:
I didn't think unpackself threw up any prompts (this is where I could be wrong).

ferventcoder:
> Hey, I had a brief look - my understanding was that PRs were fine for small changes. Was there anything else I needed to do?

Yes, there is an extensive section on the git commit message. We need to be sticklers there because it makes searching later down the line much easier if commits are formatted correctly with the correct tags on them.

smspillaz:
@ferventcoder Indeed, `unpackself` prompts about lack of admin access.

Potentially we could pass `-y` by default.

My understanding is that `-y` also auto-accepts a lot of other stuff too ([simple code search here](https://github.com/chocolatey/choco/search?utf8=%E2%9C%93&q=PromptForConfirmation&type=Code)), which is why I added the environment variable (off by default so that any potential damage is limited.

I'll update the commit message - thanks for the pointer :)

ferventcoder:
@smspillaz in the case of unpackself, there is no reason not to have it `-y`

ferventcoder:
Ah ha - https://github.com/chocolatey/choco/blob/master/src/chocolatey/infrastructure.app/commands/ChocolateyUnpackSelfCommand.cs#L108-L111

ferventcoder:
Apparently I set that. But it needs to run silently, otherwise it breaks choco conventions, that is why it should go with `-y` every time.

nuget/chocolatey/tools/chocolateysetup.psm1
```diff
@@ -39,11 +39,17 @@ param(
...
+  if ($env:ChocolateyEnvironmentAllowGlobalConfirmation -eq 'true') {
+    $allowGlobalConfirmation = '-y'
+  } else {
+    $allowGlobalConfirmation = $null
+  }
+
   $chocoNew = Join-Path $thisScriptFolder 'chocolateyInstall\choco.exe'
   if ($debugMode) {
-    & $chocoNew unpackself -fdv
+    & $chocoNew unpackself -fdv $allowGlobalConfirmation
```
ferventcoder: This makes me think we should set $debugMode as `'-dv'` and only have one command there instead of two. But also that we should always use `-y`. If the script stops for any reason, it is breaking Chocolatey conventions.

smspillaz:
Filed issue #274 for the commit message.

nuget/chocolatey/tools/chocolateysetup.psm1
smspillaz: Cool. Shall I just add `-y` everywhere then?

ferventcoder:
Sweet, let me know when you update the commit and repush it, plus the fixes I requested. :)

ferventcoder:
It's a start, but the actual commit messages themselves need prepended with `(GH-274)` and squashed down as one commit. :)

ferventcoder:
I love your first commit message - https://github.com/chocolatey/choco/pull/273/commits - well formed but needs to remove the word `nuget:` in the summary.

smspillaz:
@ferventcoder Cool. I've squashed down to two commits (one for the change, one for the cleanup recommended) ...... aaaannd CI just failed.

Do you want those two in separate commits or just one?

smspillaz:
hahaha, it *was* in the `nuget` subdirectory :P

ferventcoder:
Haha, I guess that is true, the nuget package. :)

nuget/chocolatey/tools/chocolateysetup.psm1
```diff
@@ -314,9 +310,9 @@ param(
   Copy-Item $chocoExe $chocoExeDest -force
 
   if ($debugMode) {
-    & $chocoExeDest unpackself -fdv
+    & $chocoExeDest unpackself -fdvy
```
smspillaz: Oh weird, I missed this. Hang on.

ferventcoder:
Nah, those two things are separate ideas. I don't believe in squashing separate distinct things into one commit.

smspillaz:
Oh, I need to update the commit message for the first commit. Hang tight.

smspillaz:
(you can cancel all the backed up builds on appveyor by the way, there's probably quite a few O.o)

ferventcoder:
Yeah, I better do that. :)

smspillaz:
I didn't see debugMode used anywhere else, no

smspillaz:
@ferventcoder Looks like its all passing (travis-ci is broken by a bad mono package is seems).

ferventcoder:
@smspillaz travis-ci has issues sometimes.

smspillaz:
@ferventcoder What's the timeline for getting these changes merged in? I'm waiting to start work on one of my own projects that will use choco under the hood.

ferventcoder:
This is going into 0.9.9.6, which will be releasing pretty soon.

ferventcoder:
Merged into stable at https://github.com/chocolatey/choco/commit/ae68a690f73767ef8cf52c101687f3c1ffd223cc to be released with 0.9.9.6.

Issue #295, opened 2015-05-30
Installing chocolatey removes $env:PSModulePath changes for current powershell session


chrissimon-au:
When installing in a powershell process with something like:

    iex ((new-object net.webclient).DownloadString('https://chocolatey.org/install.ps1'))

After the install, $env:PSModulePath has reverted to the Machine level environment variable.

This is causing problems for me as the Machine level variable only has:

     C:\Windows\system32\WindowsPowerShell\v1.0\Modules\;C:\Program Files (x86)\AWS Tools\PowerShell\

But as part of establishing a powershell session, $env:PSModulePath looks like:

    C:\Users\Administrator\Documents\WindowsPowerShell\Modules;C:\Program Files\WindowsPowerShell\Modules;C:\Windows\system32\WindowsPowerShell\v1.0\Modules\;C:\Program Files (x86)\AWS Tools\PowerShell\

After installing chocolatey, it reverts to:

     C:\Windows\system32\WindowsPowerShell\v1.0\Modules\;C:\Program Files (x86)\AWS Tools\PowerShell\

And the absence of the Documents and Program Files paths causes problems for other things I'm trying to do in the same script.

Seems likely this is being done by Update-SessionEnvironment.  The need for Update-SessionEnvironment is clear, but could it possible merge with the existing $env:PSModulePath rather than assuming that the process has not yet modified it?

ferventcoder:
To be clear this is only during the install.

Where are the other values resolved from?

chrissimon-au:
Yes, I've only noticed this during install.  

Not sure where those other values are coming from.  https://msdn.microsoft.com/en-us/library/dd878326%28v=vs.85%29.aspx suggests the Documents path is a default ($home\Documents\WindowsPowerShell\Modules) - perhaps it's being added by Powershell itself - but doesn't explain the Program Files path.  I'm still hunting that down - will let you know if/when I find it.

But either way - shouldn't in-process changes to $env:PSModulePath be respected, wherever they've come from?  If they were set by the script that's installing chocolatey, then you wouldn't be able to resolve them from anywhere other than $env:PSModulePath.

ferventcoder:
The problem is that Update-Session is somewhat naive, it only loads the values from the registry.

ferventcoder:
By the way, this is a duplicate of #129 - let's continue discussion there.

ferventcoder:
Actually it is not a duplicate, it is only related.

ferventcoder:
[...]

Issue #129, opened 2015-02-27
Update-SessionEnvironment doesn't update $env:PSMODULEPATH


eavonius:
I have a PowerShell Module (PowerDelivery) that I've created an MSI with the Wix Toolset for. When I install this with Install-ChocolateyInstallPackage (it's embeded as a .msi in the nuget package installed by chocolatey) I see that the environment variable gets appended to (PSMODULEPATH, system variable) but after calling Update-SessionEnvironment in chocolateyInstall.ps1 the current PowerShell environment doesn't have this set.

My chocolateyInstall.ps1:

$installDir = $(Split-Path -parent $MyInvocation.MyCommand.Definition)
$msiFile = Join-Path $installDir PowerDelivery.msi

Install-ChocolateyInstallPackage 'powerdelivery3' 'msi' '/quiet' $msiFile

Update-SessionEnvironment

This fails, executing $env:PSMODULEPATH shows the old path 
even though checking this in the Windows System Environment 
Variables control panel applet shows the path to the module has 
been appended:

Import-Module PowerDelivery -Force

ferventcoder:
Can I ask if this is INSIDE the package or once you come back out to your shell?

eavonius:
All these lines of example PowerShell are in my chocolateyInstall.ps1. When I call $env:PSMODULEPATH after chocolatey runs, this is after the chocolatey install fails.

If I omit the line to Import-Module there are no errors, but the module isn't loaded etc.

ferventcoder:
Ah okay, right on. Can you verify whether this is new choco only or also present with the older version?

eavonius:
I'm using 0.9.8.32 (I see that 0.9.8.33 is available). Would you like me to upgrade to 0.9.8.33 and test again? I haven't tried the new chocolatey yet - can these run side by side? Thanks for looking at this Rob.

ferventcoder:
Okay great, so this is an existing issue. :+1:

ferventcoder:
I mean not great that its a bug for you. But good that it is not something new choco introduced.

eavonius:
Thanks for the clarification. I'm getting ready to release a new version of PowerDelivery that's a complete rewrite myself - so I figured I might as well use the latest chocolatey. 

[...]

PR #327, opened 2015-06-14
adds helpers/functions/Install-PSModulePath.ps1

ctaggart:
This fixes chocolatey/chocolatey#292. I simply copied `Install-ChocolateyPath.ps1` and made minimal changes. I'm using it like this from `chocolateyInstall.ps1`:

```
$tools = "$env:chocolateyPackageFolder\tools"
. "$tools\Install-PSModulePath.ps1"
Install-PSModulePath -pathToInstall $tools -pathType ([System.EnvironmentVariableTarget]::Machine)
```

I would think `Machine` would be a better default, but I left it as `User`. It would be great to have an `Uninstall-PSModulePath.ps1` too.

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Install-PSModulePath.ps1
+++ b/src/chocolatey.resources/helpers/functions/Install-PSModulePath.ps1
@@ -1,0 +1,56 @@
+﻿# Copyright 2011 - Present RealDimensions Software, LLC & original authors/contributors from https://github.com/chocolatey/chocolatey
+# 
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# 
+#     http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+function Install-PSModulePath {
+param(
+  [string] $pathToInstall,
+  [System.EnvironmentVariableTarget] $pathType = [System.EnvironmentVariableTarget]::User
+)
+  Write-Debug "Running 'Install-PSModulePath' with pathToInstall:`'$pathToInstall`'";
+  $originalPathToInstall = $pathToInstall
+
+  #get the PSModulePath variable
+  Update-SessionEnvironment
+  $envPath = $env:PSModulePath
+  if (!$envPath.ToLower().Contains($pathToInstall.ToLower()))
+  {
+    Write-Host "PSModulePath environment variable does not have $pathToInstall in it. Adding..."
+    $actualPath = Get-EnvironmentVariable -Name 'PSModulePath' -Scope $pathType
+
+    $statementTerminator = ";"
+    #does the path end in ';'?
+    $hasStatementTerminator = $actualPath -ne $null -and $actualPath.EndsWith($statementTerminator)
+    # if the last digit is not ;, then we are adding it
+    If (!$hasStatementTerminator -and $actualPath -ne $null) {$pathToInstall = $statementTerminator + $pathToInstall}
+    if (!$pathToInstall.EndsWith($statementTerminator)) {$pathToInstall = $pathToInstall + $statementTerminator}
+    $actualPath = $actualPath + $pathToInstall
+
+    if ($pathType -eq [System.EnvironmentVariableTarget]::Machine) {
+      if (Test-ProcessAdminRights) {
+        Set-EnvironmentVariable -Name 'PSModulePath' -Value $actualPath -Scope $pathType
+      } else {
+        $psArgs = "Install-PSModulePath -pathToInstall `'$originalPathToInstall`' -pathType `'$pathType`'"
+        Start-ChocolateyProcessAsAdmin "$psArgs"
+      }
+    } else {
+      Set-EnvironmentVariable -Name 'PSModulePath' -Value $actualPath -Scope $pathType
+    }
+
+    #add it to the local path as well so users will be off and running
+    $envPSPath = $env:PSModulePath
+    $env:PSModulePath = $envPSPath + $statementTerminator + $pathToInstall
+  }
+}
+
+# [System.Text.RegularExpressions.Regex]::Match($Path,[System.Text.RegularExpressions.Regex]::Escape('locationtoMatch') + '(?>;)?', '', [System.Text.RegularExpressions.RegexOptions]::IgnoreCase)


src/chocolatey.resources/helpers/functions/Install-PSModulePath.ps1
```diff
@@ -0,0 +1,56 @@
+﻿# Copyright 2011 - Present RealDimensions Software, LLC & original authors/contributors from https://github.com/chocolatey/chocolatey
```
ferventcoder: This will be a little different since it originates here. The `& original...` should be removed.

src/chocolatey.resources/helpers/functions/Install-PSModulePath.ps1
```diff
@@ -0,0 +1,56 @@
...
+    if ($pathType -eq [System.EnvironmentVariableTarget]::Machine) {
+      if (Test-ProcessAdminRights) {
+        Set-EnvironmentVariable -Name 'PSModulePath' -Value $actualPath -Scope $pathType
+      } else {
+        $psArgs = "Install-PSModulePath -pathToInstall `'$originalPathToInstall`' -pathType `'$pathType`'"
+        Start-ChocolateyProcessAsAdmin "$psArgs"
+      }
+    } else {
+      Set-EnvironmentVariable -Name 'PSModulePath' -Value $actualPath -Scope $pathType
+    }
```
ferventcoder: Most of line 39-48 repeats https://github.com/chocolatey/choco/blob/master/src/chocolatey.resources/helpers/functions/Install-ChocolateyEnvironmentVariable.ps1#L57-L66

You should just call that instead.

src/chocolatey.resources/helpers/functions/Install-PSModulePath.ps1
```diff
@@ -0,0 +1,56 @@
...
+
+    #add it to the local path as well so users will be off and running
+    $envPSPath = $env:PSModulePath
+    $env:PSModulePath = $envPSPath + $statementTerminator + $pathToInstall
+  }
+}
+
+# [System.Text.RegularExpressions.Regex]::Match($Path,[System.Text.RegularExpressions.Regex]::Escape('locationtoMatch') + '(?>;)?', '', [System.Text.RegularExpressions.RegexOptions]::IgnoreCase)
```
ferventcoder: Remove this since it was copy paste from another function.

ferventcoder:
On the actual commit message itself, please create an issue here in chocolatey/choco, link it back to that issue and then format your commit message according to CONTRIBUTING.md. Once you've done that please force push your branch to update the PR. Thanks!

src/chocolatey.resources/helpers/functions/Install-PSModulePath.ps1
ferventcoder: I do get that this was a copy, but the original probably needs updated as well, so... :)

ferventcoder:
From https://github.com/chocolatey/chocolatey/issues/292#issuecomment-47425437 :

>One nuance I ran across while retrieving the PATH variable with [System]::GetEnvironmentVariable (specifying the machine scope) is that it still returned the user + machine scope, so I had to also handle if the user-scoped path was there or not. (I found this out the hard way and ended up breaking the Chocolatey prefetch in the puppet module because I did not properly remove the path from $PSModulePath).

@JimFicarra how did you end up handling the split of user from machine scope?

JimFicarra:
Hey Rob,

Are you having this issue?  I remember writing this and having the issue, but I don't remember the conditions that triggered it.  I looked back at the code I wrote and I don't see me handling it other than using the Machine scope (EnvironmentVariableTarget Enumeration) so perhaps I was not using the proper parameter for targeting the machine scope at the time I wrote it.. :)  

In any case, below are snippets of code I wrote back then in the ChocolateyInstall.ps1 to handle the addition of the variable to machine scope.  Hope they are helpful.
```
$installDir = "C:\Program Files\Common Files\MyCompany\PowerShellModules"
$modPath = "C:\Program Files\Common Files\MyCompany\PowerShellModules"
$psModulePath = [Environment]::GetEnvironmentVariable("PSModulePath", "Machine")

#Add the path to the machine scoped PSModulePath environment variable only if not there
if (-not ($psModulePath -match [regex]::Escape($modPath)))
	{
		Write-Host "Adding $installDir to PSModulePath"
		Write-Debug "Adding $installDir to PSModulePath"
		$psModulePath += ";" + $installDir
		[Environment]::SetEnvironmentVariable("PSModulePath",$psModulePath, "Machine")
	}
else
	{
		Write-Host "Machine Environment Variable PSModulePath already contains $installDir.  Variable not updated"
		Write-Debug "Machine Environment Variable PSModulePath already contains $installDir.  Variable not updated"
	}
```
And here's a function I wrote in the chocolateyuninstall.ps1 that removes the desired path from the PSModulePath env variable.  It handles the cases where the path is either at the beginning, end, or middle of the entire semicolon-delimited path string.  

In my implementation, this function is called in my uninstall ONLY IF the module path I'm removing from the PSModulePath env variable doesn't contain any other modules after removing the specific modules within this uninstall.  (I have other company-specific modules that I write and put in this path, hence the check before removing from the PSModulePath - I don't want to break usage for other modules I put in my common path)

```
#See if the installDir is empty, if so remove from PSModulePath
		if ((Get-ChildItem $installDir) -eq $NULL)
		{
			RemoveFrom-PSModulePath $installDir
		}
		else
		{
			Write-Host "$installDir is not empty - not removing path from from PSModulePath as it may break other modules installed in this path"
			Write-Debug "$installDir is not empty - not removing path from from PSModulePath as it may break other modules installed in this path"
		}
```

Below is the function called.

```
Function RemoveFrom-PSModulePath
{

Param(
	[Parameter(Mandatory=$True)]
	[string]$modPath

)
	####################################################
	##Remove path from System Env Variable PSModulePath#
	####################################################
	
	#We remove only if the path is in the env variable AND there are no more folders in the installation directory	
	$psModulePath = [Environment]::GetEnvironmentVariable("PSModulePath", "Machine")
	
	#Remove the path to the machine scoped PSModulePath environment veriable only if not there
	if (($psModulePath -match [regex]::Escape($modPath)))
	{
		Write-Host "Removing $installDir from PSModulePath"
		Write-Debug "Removing $installDir from PSModulePath"
		# If it's at the end of the path we remove the preceeding ';' as well.
		if ($psModulePath.Split(';')[-1] -match [regex]::Escape($modPath))
		{
			$psModulePath = $psModulePath -replace [regex]::Escape(";$modPath"), $NULL
			[Environment]::SetEnvironmentVariable("PSModulePath",$psModulePath, "Machine")
		}
		# Else if it's at the front we remove the trailing ';'
		elseif ($psModulePath.Split(';')[0] -match [regex]::Escape($modPath))
		{
			$psModulePath = $psModulePath -replace [regex]::Escape("$modPath;"), $NULL
			[Environment]::SetEnvironmentVariable("PSModulePath",$psModulePath, "Machine")
		}
		#Else if it's in the middle we replace with a ';'
		elseif ($psModulePath.Split(';') -match [regex]::Escape($modPath))
		{
			$psModulePath = $psModulePath -replace [regex]::Escape(";$modPath;"), ";"
			[Environment]::SetEnvironmentVariable("PSModulePath",$psModulePath, "Machine")
		}		
		
	}
	else
	{
		Write-Host "System Environment Variable PSModulePath does not contain $installDir.  Variable not updated"
		Write-Debug "System Environment Variable PSModulePath does not contain $installDir.  Variable not updated"
	}
}
```

ctaggart:
@JimFicarra It looks like you've done much more work here. Would you like to replace this PR with a better one?

JimFicarra:
@ctaggart - I'd be happy to contribute!  Let me clean up the code a bit and I'll submit a PR.

eavonius:
Any update on this? I've been trying to set PSMODULEPATH in two modules where one is dependent on the other. If they both get installed in the same choco session, the second one (dependent on the other) fails since Update-SessionEnvironment, and setting $env and [Environment]::SetEnvironmentVariable don't work across the installs. No, I can't combine the modules - on some machines users will install one, on some both.

ferventcoder:
@eavonius last I knew I believe @JimFicarra was going to fix up his code and submit a superseding PR for this feature.

eavonius:
Thanks rob. @JimFicarra any update? That was 5 months ago ;). Hopefully this is still working code, I could sure use it! :)

ferventcoder:
> Any update on this? I've been trying to set PSMODULEPATH in two modules where one is dependent on the other. If they both get installed in the same choco session, the second one (dependent on the other) fails since Update-SessionEnvironment, and setting $env and [Environment]::SetEnvironmentVariable don't work across the installs. No, I can't combine the modules - on some machines users will install one, on some both.

This has been fixed in 0.9.10.

ferventcoder:
@JimFicarra any updates on submitting those fixes?

jberezanski:
> > Any update on this? I've been trying to set PSMODULEPATH in two modules where one is dependent on the other. If they both get installed in the same choco session, the second one (dependent on the other) fails since Update-SessionEnvironment, and setting $env and [Environment]::SetEnvironmentVariable don't work across the installs. No, I can't combine the modules - on some machines users will install one, on some both.

Cross-referencing chocolatey/chocolatey#475.

> This has been fixed in 0.9.10.

Glad to hear!

eavonius:
Revisiting this - how has this been fixed in 0.9.10? I just tried on Windows 10 using 0.9.9.12 (current version) installing my two powershell modules that have chocolatey packages again. I'm still running into the issue where when the dependent package installs, the next package doesn't see the change to PSMODULEPATH even though I add it to $env:PSMODULEPATH, the registry, and call Update-SessionEnvironment -Full.

Here's the chocolateyInstall from PowerDelivery3Node (PowerDelivery3 package depends on it). When I install PowerDelivery3 locally off the built nuget package, it does find the dependency to PowerDelivery3Node, installs it first, and there are no errors - but at the end of both packages being installed the current PowerShell session doesn't have the new cmdlets available.

https://github.com/eavonius/powerdelivery/blob/master/Scripts/Chocolatey/PowerDeliveryNode/ChocolateyInstall.ps1

Should I just show a message that the person has to reload their shell to use it? I'm using this as part of infrastructure automation and if I call chocolatey on a remote node to install PowerDeliveryNode, I want the calling script to be able to access its cmdlets immediately after installation.

ferventcoder:
> Revisiting this - how has this been fixed in 0.9.10? I just tried on Windows 10 using 0.9.9.12 (current version) installing my two powershell modules that have chocolatey packages again. It appears that when the dependent package installs, the next package does now see the change to PSMODULEPATH which is good, however the calling process doesn't see the new modules.

0.9.10.0 versus 0.9.9.12. 0.9.10 is still in beta. I'm not sure if you made that determination or not.

ferventcoder:
@eavonius in 0.9.10 a user can refresh their shell with `refreshenv`, which calls Update-SessionEnvironment`. 

refresh env

ferventcoder:
Note a couple of really important things there - Chocolatey tells the user that they need to refresh their environment when it detects changes to the environment values.

ferventcoder:
You can call `refreshenv` after the installation (or use `Update-SessionEnvironment`) outside of Chocolatey scripts - if the installation is able to put it on the profile. I'm not completely sure how it works with SYSTEM, I'd love some testing on that if you are able to.

ferventcoder:
https://github.com/chocolatey/choco/issues/295

eavonius:
Sorry Rob I didn't notice I was reading the version wrong. Is there a way
to install this version without getting source?

I'll do some testing, really want to get these packages up there!

Thanks for your support,

Jayme

ferventcoder:
choco upgrade chocolatey --pre

On Tuesday, May 31, 2016, Jayme Edwards <notifications@github.com> wrote:

> Sorry Rob I didn't notice I was reading the version wrong. Is there a way
> to install this version without getting source?
>
> I'll do some testing, really want to get these packages up there!
>
> Thanks for your support,
>
> Jayme
>
> On Mon, May 30, 2016 at 11:08 AM, Rob Reynolds <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>>
> wrote:
>
> > #295 <https://github.com/chocolatey/choco/issues/295>
> >
> > —
> > You are receiving this because you were mentioned.
> > Reply to this email directly, view it on GitHub
> > <https://github.com/chocolatey/choco/pull/327#issuecomment-222521552>,
> or mute
> > the thread
> > <
> https://github.com/notifications/unsubscribe/AAWspTXQTKMzjWGu23tZcT--C0kHWv0qks5qGwt0gaJpZM4FCQ8g
> >
> > .
> >
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/chocolatey/choco/pull/327#issuecomment-222852226>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAD4Dss1PwGcN-aWjMfeBP4YR56m96H5ks5qHMVwgaJpZM4FCQ8g>
> .
>


-- 

Rob
"Be passionate in all you do"

http://codebetter.com/robreynolds
http://ferventcoder.com
https://twitter.com/ferventcoder

eavonius:
Darn it. OK thanks Rob, I tried it with the pre package and the packages
install without errors (so the dependent one sees the dependency in its own
process space) but when I do a refreshenv afterwards, the updated system
environment variable for PSMODULEPATH is missing the new paths. I checked
the system environment variables dialog and they are added, but they don't
show up until I close and re-open the terminal :(

ferventcoder:
@eavonius help me understand this a bit more - we don't actually update PSMODULEPATH at all. We hold the original variable and give it back - due to #295. Are you saying you would prefer we update it?

eavonius:
Sorry for the lack of clarity. The chocolatey packages for my two
PowerShell modules append their chocolatey "tools" subdirectories to the
PSMODULEPATH system environment variable during chocolateyInstall.ps1. When
they are done installing, I would like a way for PSMODULEPATH to include
the new paths I've added so any further commands in a scripts that calls
chocolatey as just one part of an overall deployment process have access to
the new cmdlets.

Hopefully that helps.

ferventcoder:
That makes sense, and conflicts with #295 as I understand it. So we will need to figure out a way to make #129 work as well. Perhaps we should continue the discussion there?

eavonius:
That makes sense. I just gave a cursory look at #295.

dragon788:
I think with some other discussion in various places it was suggested this could really just be a specific call to Install-ChocolateyEnvironmentVariable with maybe some logic checks for existing within the environment already, or and is very similar to Install-ChocolateyPath. I'm not sure if just making it an alias or a special case inside the Install-ChocolateyEnvironmentVariable would make sense, detecting when --case-insensitive "PATH"/"PSModulePath" is the target and doing some extra checks like ensuring the ExpandSZ or whatever properties are intact.

ferventcoder:
@ctaggart reviving an old thread. I would love to get something in regarding PS ModulePath. Still interested in moving forward with this?

ctaggart:
@ferventcoder I think it is still a good idea, but I will be unable to contribute.

glachancecmaisonneuve:
Seems like this commit was abandoned. I'm waiting for changes to this commit to go through or be abandoned before sending my own PR.

Issue #369, opened 2015-07-28
Travis build fail: Could not load file or assembly `Microsoft.Build.Utilities`


sideeffffect:
I've noticed that the build on Travis fails because the `.deb` package `mono-gmcs` doesn't exist anymore.
So I removed the installation of `mono-gmcs`, but now the build fails with:

```
[exec] Starting '/usr/bin/xbuild (/home/travis/build/sideeffffect/choco/src/chocolatey.sln /nologo /property:OutputPath='/home/travis/build/sideeffffect/choco/build_output/chocolatey' /property:Configuration=Release /verbosity:detailed /toolsversion:4.0 /property:Platform='Any CPU'  /property:TargetFrameworkProfile='' /property:TargetFrameworkVersion=v4.0 /l:ThoughtWorks.CruiseControl.MSBuild.XmlLogger,"/home/travis/build/sideeffffect/choco/lib/NAnt/ThoughtWorks.CruiseControl.MSBuild.dll";'/home/travis/build/sideeffffect/choco/build_output/build_artifacts/compile/msbuild-mono-4.0-results.xml')' in '/home/travis/build/sideeffffect/choco/build_output'
[exec] 
[exec] Unhandled Exception:
[exec] System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.Build.Utilities, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies.
[exec] File name: 'Microsoft.Build.Utilities, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
[exec]   at (wrapper managed-to-native) System.Reflection.Assembly:InternalGetType (System.Reflection.Assembly,System.Reflection.Module,string,bool,bool)
[exec]   at System.Reflection.MonoAssembly.GetType (System.String name, Boolean throwOnError, Boolean ignoreCase) [0x00000] in <filename unknown>:0 
[exec]   at System.Reflection.Assembly.GetType (System.String name) [0x00000] in <filename unknown>:0 
[exec]   at Mono.XBuild.CommandLine.MainClass.Execute () [0x00000] in <filename unknown>:0 
[exec]   at Mono.XBuild.CommandLine.MainClass.Main (System.String[] args) [0x00000] in <filename unknown>:0 
```

even though the package `libmono-microsoft-build-utilities-v4.0-4.0-cil` is installed
[...]

PR #370, opened 2015-07-29
Fix install stage on traivs

sideeffffect:
this is the first step to fix travis builds
the build still fails though (#369)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: csharp
 install:
-  - sudo apt-get install mono-devel mono-gmcs
   - nuget restore src/chocolatey.sln
 before_script:
   - chmod +x build.sh


ferventcoder:
@sideeffffect Have you had an opportunity to read over our CONTRIBUTING document?

.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: csharp
 install:
-  - sudo apt-get install mono-devel mono-gmcs
```
ferventcoder: Does this mean that mono-devel is already there?

sideeffffect: [indeed there is](https://travis-ci.org/chocolatey/choco/builds/72637212#L953)

sideeffffect:
hi, I've just glossed over it now

 * this seems to be *trivial* change so we shouldn't need CLA; since this project is licensed under a permissive license, I would sign
 * I see I should have made a special name for the commit (`(maint)`), but I don't know how to rename it now

.travis.yml
ferventcoder: :+1:

ferventcoder:
Read through the entire contributing document, it will explain how to fix your commit messages and repush the changes.

ferventcoder:
I actually think we'll close this one for now in preference for #372 (and while we wait on the Xamarin bug to get evaluated). Thank you for researching that! 

I think the next step is to get the specs failure working, which is wide open for either of you to do if you are interested.

ferventcoder:
I pushed the fixes for #372 with fixups to reference the bug you filed against Xamarin in the commit - see https://github.com/chocolatey/choco/commit/6c200e11e5b8b4be428ef6a85e80c5e138a7643a for details.

ferventcoder:
I fixed a few of the specs. Let's see if that means we have a green Travis build.

ferventcoder:
And we are green! 
image

PR #372, opened 2015-07-30
Fix Travis-CI builds by specifying mono version. Fails NUnit tests.

finchd:
Related Travis Doc [here](http://docs.travis-ci.com/user/languages/csharp/#Choosing-Mono-version-to-test-against)

I chose 3.12.0 which actually gets the latest in the 3.12.X line, as
mentioned in [Travis-CI Issue #4130](https://github.com/travis-ci/travis-ci/issues/4130)

Travis Builds still fail, that that is a code problem, not an infrastructure one. ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: csharp
+mono:
+  - 3.12.0
 install:
   - sudo apt-get install mono-devel mono-gmcs
   - nuget restore src/chocolatey.sln


ferventcoder:
Thanks. And now I have two PRs about the same thing. #370.

I have been working with you quite a bit longer than @sideeffffect and you were working on this quite a bit earlier, but the preference is in which option works the best. 

Right now it appears preferring an earlier mono version (as done here) gets things down to the failing at the proper point instead of just at the build.

ferventcoder:
Fixed up the commit wording, including adding notes for the bug that @sideeffffect has filed and rebased onto stable. Merged into stable at https://github.com/chocolatey/choco/commit/d8943f263ab5c5ba088cb8d7b21471cec0c765e0

ferventcoder:
Thanks!

ferventcoder:
For reference, this is the related commit https://github.com/chocolatey/choco/commit/6c200e11e5b8b4be428ef6a85e80c5e138a7643a

ferventcoder:
I fixed a few of the specs. Let's see if that means we have a green Travis build.

ferventcoder:
And we are green! 
image

PR #376, opened 2015-07-30
[WIP - DO NOT MERGE](GH-338) Delete read only files

### LINKED ISSUE #338: Deleting read-only files fails
There seems to be a problem removing files with read-only attribute during upgrade and uninstall. Warnings look like this (in this case, "Clock.ico" is the read-only file):
https://gist.github.com/qzo/e4965ecdbb6e8ca773a3
(upgrade log, the directory was not removed before due to unsuccessful uninstall, see below)
https://gist.github.com/qzo/5e77b1844f87b4bef915
(uninstall log)

How to reproduce:
1. Create a simple package which downloads and extracts an archive containing a file with read-only attribute using Install-ChocolateyZipPackage - an example of such package is here:
   https://github.com/qzo/chocolatey-packages/tree/6b892168b853ce537cf9f1b25aa2009f5f266a5a/packages/manual/farr_m/find-and-run-robot.portable
   (instead of
   http://www.donationcoder.com/Software/Mouser/findrun/downloads/FindAndRunRobotPortable.zip
   which might point to a new version of the application you can use this link:
   https://drive.google.com/uc?export=download&id=0B87Bx4rkw5f6d0NSODBUeUMyOHc
   ).
2. Install the package.
3. Uninstall the package - You get the warnings and read-only files are not removed from either lib or lib-bkp. (And unless you remove them manually, subsequent installation / upgrade won't be able to remove the old files either, causing the same warnings.)

ferventcoder:
This closes #338

### DIFF
--- a/src/chocolatey.tests.integration/chocolatey.tests.integration.csproj
+++ b/src/chocolatey.tests.integration/chocolatey.tests.integration.csproj
@@ -227,6 +227,9 @@
     <Content Include="infrastructure\filesystem\MoveMe.txt">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </Content>
+    <Content Include="infrastructure\filesystem\context\attributes.txt">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
     <Content Include="infrastructure\filesystem\Slipsum.txt">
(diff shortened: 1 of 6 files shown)


ferventcoder:
Wrong branch.

Issue #303, opened 2015-06-02
Install-ChocolateyPath Expands Variables in PATH, Overwriting Preexisting Variables


mnadel:
Install-ChocolateyPath will expand all variables in your PATH, and then write back the expanded values, clobbering any *variables* that existed in PATH prior to Install-ChocolateyPath being called.

e.g. `%PYTHON_HOME%;C:\Utils` -> `C:\Python27;C:\Utils`

ferventcoder:
Clobbering sounds strong - it's really just accidentally expanding the values, right?

ferventcoder:
Okay - I get it about clobbering variables in entries - might need a better way of stating this - I took it wrong on first read.

mnadel:
Sorry about that -- updated the title and added an example.

ferventcoder:
:+1:

mnadel:
This seem like a reasonable approach? Would be happy to open a PR.

http://stackoverflow.com/questions/4598532/powershell-get-tmp-environment-variable-raw-value

ferventcoder:
Yep, seems reasonable - I did similar in Puppet's windows installer and planned to reimplement here at some point. Target stable branch.

ferventcoder:
Also make sure you read and agree to CONTRIBUTING.

ferventcoder:
Bumping this to 0.9.10 while waiting for updates from @mnadel

mnadel:
In reviewing this, I believe the latest commit on 30 Jul contains all changes per our latest [conversation](https://github.com/mnadel/choco/commit/7773cac5695abec967a30b899d488b293ec6bf5d#commitcomment-11558843). Let me know if you've got any other feedback, otherwise I believe it's ready for merging.

ferventcoder:
@mnadel If you can review the PR, it doesn't appear to be updated yet. Example https://github.com/chocolatey/choco/pull/373#discussion_r39821156

DarwinJS:
Not sure if it's still relevant, but here is a link to the code I put together for this today.  The code I did also handles removing duplicates (if the path to remove is in more than once), handles the path whether or not there is a trailing slash versus what is specified on the removal string and it handles removing extra ";' that are found.  The code is *not* generalized to handle both user and machine path.  https://github.com/DarwinJS/ChocoPackages/blob/master/ec2clitools/tools/chocolateyuninstall.ps1

ferventcoder:
This change had a very bad side effect. 

[...]

PR #373, opened 2015-07-30
GH-303 Install-ChocolateyPath Expands Variables in PATH, Overwriting

mnadel:
*This is untested* ... thoughts on how to test this on a Mac?

Closes #303

PR #378, opened 2015-07-31
GH-303 Install-ChocolateyPath Expands Variables in PATH, Overwriting

mnadel:
Preexisting Variables

When adding a new element to PATH, Install-ChocolateyPath reads PATH
(wh ich contains expanded variables), appends the new element, and then
writes the value back to PATH. This results in PATH having its
variables overwritten with its values. Instead of reading $env:PATH to
find the current path, read it from the registry, which will retain the
PATH's original variables.


ferventcoder:
?

mnadel:
Wrong branch name. Repushed w/ original branch name to ensure that all changes are tracked under the same [PR-373](https://github.com/chocolatey/choco/pull/373).

PR #393, opened 2015-08-23
(GH-392) pass 4.0 reference assemblies to ilmerge

### LINKED ISSUE #392: Merging assemblies on a machine running .net 4.5 or higher produces binaries incompatible with .net  4.0
Using either source built binaries on my machine (running 4.5.1) or the binaries provided by nuget result in assembly load errors when executed on a windows 2k8R2 machine after installing .net 4.0.

This can be fixed by either installing .net4.5 on the target machine or, preferably, informing ILMerge of the location of the 4.0 reference assemblies.

mwrock:
Makes chocolatey builds more portable and allows it to be built on .net
4.5+ machines while maintaining compatibility with .net 4.0 clients.

Closes #392

### DIFF
--- a/.build/ilmerge.build
+++ b/.build/ilmerge.build
@@ -23,8 +23,8 @@
   <property name="file.merge.name" value="__REPLACE__" overwrite="false" />
   <!-- exe winexe dll -->
   <property name="merge.target.type" value="__REPLACE__" overwrite="false" />
-  <property name="args.ilmerge" value="/internalize /targetplatform:v4 /target:${merge.target.type} /out:${dirs.merge.to}${path.separator}${file.merge.name} /log:${dirs.merge.log}${path.separator}${log.merge} /xmldocs /ndebug /allowDup ${file.merge.name} " />
-  <property name="args.ilmerge" value="/internalize:${path.file.internalize.ignore} /targetplatform:v4 /target:${merge.target.type} /out:${dirs.merge.to}${path.separator}${file.merge.name} /log:${dirs.merge.log}${path.separator}${log.merge} /xmldocs /ndebug /allowDup ${file.merge.name} " if="${file::exists(path.file.internalize.ignore)}"/>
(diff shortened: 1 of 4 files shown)


.build/ilmerge.build
```diff
@@ -38,8 +38,8 @@
...
-    <property name="args.ilmerge" value="/internalize:${path.file.internalize.ignore} /targetplatform:v4 /target:${merge.target.type} /out:${dirs.merge.to}${path.separator}${file.merge.name} /log:${dirs.merge.log}${path.separator}${log.merge} /ndebug /allowDup ${file.merge.name} " if="${file::exists(path.file.internalize.ignore)}"/>
+    <property name="args.ilmerge" value="/internalize /targetplatform:v4,&quot;${environment::get-variable('ProgramFiles(x86)')}\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.0&quot; /target:${merge.target.type} /out:${dirs.merge.to}${path.separator}${file.merge.name} /log:${dirs.merge.log}${path.separator}${log.merge} /ndebug /allowDup ${file.merge.name} " />
+    <property name="args.ilmerge" value="/internalize:${path.file.internalize.ignore} /targetplatform:v4,&quot;${environment::get-variable('ProgramFiles(x86)')}\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.0&quot; /target:${merge.target.type} /out:${dirs.merge.to}${path.separator}${file.merge.name} /log:${dirs.merge.log}${path.separator}${log.merge} /ndebug /allowDup ${file.merge.name} " if="${file::exists(path.file.internalize.ignore)}"/>
```
ferventcoder: Anything that is changed in .build folder should also be changed over at https://github.com/chucknorris/uppercut/tree/master/product/uppercut.build :)

ferventcoder:
So what you are saying is, to keep targeting 4.0 instead of 4.5, you should do this? :)

mwrock:
Kinda. Otherwise ILMerge targets the actual assemblies of the installed framework and because all post 4.0 framework versions are "in place" upgrades, you end up targetting whatever version is installed on the build box. Targeting the reference assemblies is the way to go.

ferventcoder:
Works for me.

PR #423, opened 2015-09-23
[DO NOT MERGE](maint) Do not assume $env:TEMP is set in powershell helper functions

rhillaardvark:
This is an identical fix to a missing TEMP environment variable in chocolatey/chocolatey.org#139

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Install-ChocolateyPackage.ps1
+++ b/src/chocolatey.resources/helpers/functions/Install-ChocolateyPackage.ps1
@@ -101,6 +101,9 @@ param(
 
   Write-Debug "Running 'Install-ChocolateyPackage' for $packageName with url:`'$url`', args: `'$silentArgs`', fileType: `'$fileType`', url64bit: `'$url64bit`', checksum: `'$checksum`', checksumType: `'$checksumType`', checksum64: `'$checksum64`', checksumType64: `'$checksumType64`', validExitCodes: `'$validExitCodes`' ";
 
+  if ($env:TEMP -eq $null) {
+    $env:TEMP = Join-Path $env:SystemDrive 'temp'
+  }
   $chocTempDir = Join-Path $env:TEMP "chocolatey"
   $tempDir = Join-Path $chocTempDir "$packageName"
   if ($env:packageVersion -ne $null) {$tempDir = Join-Path $tempDir "$env:packageVersion"; }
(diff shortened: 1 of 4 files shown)


ferventcoder:
Why is this necessary?

ferventcoder:
I'm asking mostly because we can assume it is set. We set it https://github.com/chocolatey/choco/blob/master/src/chocolatey/infrastructure.app/services/PowershellService.cs#L199

rhillaardvark:
I am calling into chocolatey.dll from a front end that was not setting the CacheLocation, causing the Start-ChocolateyProcessAsAdmin script to fail. Setting the CacheLocation in the configuration fixes the issue, but the change here would provide a fallback/default value for TEMP if it is not specified in the configuration.

ferventcoder:
> I am calling into chocolatey.dll from a front end that was not setting the CacheLocation, causing the Start-ChocolateyProcessAsAdmin script to fail.

Ah, okay! Can you give me more information on this so I can help set you up for the right calls in? As this is something that should definitely still get called.

rhillaardvark:
Your reference above was enough to point me to the fact that I wasn't setting the configuration up properly via the ConfigurationBuilder (I was constructing my own ChocolateyConfiguration). Looks like I've got the primary functions working properly now (List, Install, Uninstall) via the ChocolateyPackageService.

So yes, this is definitely not needed in a properly set up application. I jumped the gun on the PR

ferventcoder:
Are you coming in from the GetChocolatey API? https://github.com/chocolatey/choco/blob/master/src/chocolatey/GetChocolatey.cs#L168

It should guide you to a fluent setup and call for Run.

rhillaardvark:
I looked at using the GetChocolatey API, but in the case of the list command I want to put the package information into a viewmodel, so it made more sense to me to call into the ChocolateyPackageService directly.

ferventcoder:
> I looked at using the GetChocolatey API, but in the case of the list command I want to put the package information into a viewmodel, so it made more sense to me to call into the ChocolateyPackageService directly.

Any issues you can find there are helpful to move it forward. It's still considered a preview, but really the only approved way folks should interact with Chocolatey since it represents the API. If you move out of using that, wild changes can and will occur and b/c we don't support that, you will be left on your own to figure out what changed and how to fix it. Versus using the approved API, if we break it, we fix it, and expected breakages will follow the semver model.

rhillaardvark:
Understood, thanks for the help!

Issue #427, opened 2015-09-24
Add paging to list/search command


RichiCoder1:
Add the ability to page to the list command and command configuration.

PR #428, opened 2015-09-24
(GH-427) Add paging to ListCommand

RichiCoder1:
See #427.

Adds `--page=#` and `--page-size=#` to `ListCommand`. Page size defaults to 25.

### DIFF
--- a/src/chocolatey/infrastructure.app/commands/ChocolateyListCommand.cs
+++ b/src/chocolatey/infrastructure.app/commands/ChocolateyListCommand.cs
@@ -15,6 +15,7 @@
 
 namespace chocolatey.infrastructure.app.commands
 {
+    using System;
     using System.Collections.Generic;
     using System.Linq;
     using attributes;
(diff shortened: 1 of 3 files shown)


ferventcoder:
Oh wow, I can't wait to play with this just in regular choco.

RichiCoder1:
About the only flaw with this is it currently only works for Nuget. Nuget it is the worst offender, but I'll leave it up to you if you want me to fix the other source providers in this PR, or another.

ferventcoder:
Let's go there appveyor.

ferventcoder:
> only works for Nuget

Like only NuGet sources?

RichiCoder1:
Yup. Nuget is the only source runner that respects the option currently. Cygwin, Python, RubyGems, WebPI, and WindowsFeature currently simply just ignore it.

RichiCoder1:
Though it looks like Cygwin doesn't implement list. Just the others.

PR #430, opened 2015-09-25
(GH-431) Add Count to List Command API

### LINKED ISSUE #431: [API] Add the ability to retrieve package count for a Source
I need to ability to efficiently retrieve the package count for a source that meets the given criteria.

PR at #430

RichiCoder1:
Adds the ability to efficiently retrieve a "count" for a given list command. Usefully for when you want to total number of given packages for a command that meets certain search criteria, but you're also using paging and need a reliable "total". Mostly only relevant to NuGet at this point.

Closes #431

### DIFF
--- a/src/chocolatey/GetChocolatey.cs
+++ b/src/chocolatey/GetChocolatey.cs
@@ -222,6 +222,15 @@ namespace chocolatey
             var runner = new GenericRunner();
             return runner.list<T>(configuration, _container, isConsole: false, parseArgs: null);
         }
+
+        public int Count()
+        {
+            extract_resources();
+            var configuration = create_configuration(new List<string>());
+            configuration.RegularOutput = true;
+            var runner = new GenericRunner();
+            return runner.count(configuration, _container, isConsole: false, parseArgs: null);
+        }
     }
 
     // ReSharper restore InconsistentNaming
--- a/src/chocolatey/GetChocolatey.cs
+++ b/src/chocolatey/GetChocolatey.cs
@@ -185,6 +185,36 @@ namespace chocolatey
             runner.run(args, configuration, _container);
         }
 
+        /// <summary>
+        ///    Run chocolatey after setting the options, and list the results.
+        /// </summary>
+        /// <typeparam name="T">The typer of results you're expecting back.</typeparam>
+        public IEnumerable<T> List<T>()
+        {
+            extract_resources();
+            var configuration = create_configuration(new List<string>());
+            configuration.RegularOutput = true;
+            var runner = new GenericRunner();
+            return runner.list<T>(configuration, _container, isConsole: false, parseArgs: null);
+        }
+
+        /// <summary>
+        ///    Run chocolatey after setting the options,
+        ///    and get the count of items that would be returned if you listed the results.
+        /// </summary>
+        /// <remarks>
+        ///    Is intended to be more efficient then simply calling <see cref="List{T}">List</see> and then Count() on the returned list.
+        ///    It also returns the full count as is ignores paging.
+        /// </remarks>
+        public int ListCount()
+        {
+            extract_resources();
+            var configuration = create_configuration(new List<string>());
+            configuration.RegularOutput = true;
+            var runner = new GenericRunner();
+            return runner.count(configuration, _container, isConsole: false, parseArgs: null);
+        }
+
         private ChocolateyConfiguration create_configuration(IList<string> args)
         {
             var configuration = new ChocolateyConfiguration();
@@ -212,24 +242,6 @@ namespace chocolatey
             };
 
             AssemblyFileExtractor.extract_all_resources_to_relative_directory(_container.GetInstance<IFileSystem>(), Assembly.GetAssembly(typeof(ChocolateyResourcesAssembly)), ApplicationParameters.InstallLocation, folders, ApplicationParameters.ChocolateyFileResources);
-        }
-
-        public IEnumerable<T> List<T>()
-        {
-            extract_resources();
-            var configuration = create_configuration(new List<string>());
-            configuration.RegularOutput = true;
-            var runner = new GenericRunner();
-            return runner.list<T>(configuration, _container, isConsole: false, parseArgs: null);
-        }
-
-        public int Count()
-        {
-            extract_resources();
-            var configuration = create_configuration(new List<string>());
-            configuration.RegularOutput = true;
-            var runner = new GenericRunner();
-            return runner.count(configuration, _container, isConsole: false, parseArgs: null);
         }
     }
 
--- a/src/chocolatey/infrastructure.app/nuget/NugetList.cs
+++ b/src/chocolatey/infrastructure.app/nuget/NugetList.cs
@@ -17,6 +17,7 @@ namespace chocolatey.infrastructure.app.nuget
 {
     using System.Collections.Generic;
     using System.Linq;
+
     using NuGet;
     using configuration;
 
@@ -26,12 +27,41 @@ namespace chocolatey.infrastructure.app.nuget
     {
         public static IEnumerable<IPackage> GetPackages(ChocolateyConfiguration configuration, ILogger nugetLogger)
         {
+            return execute_package_search(configuration, nugetLogger);
+        }
+
+        private static IQueryable<IPackage> execute_package_search(ChocolateyConfiguration configuration, ILogger nugetLogger)
+        {
             var packageRepository = NugetCommon.GetRemoteRepository(configuration, nugetLogger);
+
+            // Whether or not the package is remote determines two things:
+            // 1. Does the repository have a notion of "listed"?
+            // 2. Does it support prerelease in a straight-forward way?
+            // Choco previously dealt with this by taking the path of least resistance and manually filtering out and sort unwanted packages
+            // This result in blocking operations that didn't let service based repositories, like OData, take care of heavy lifting on the server.
+            bool isRemote;
+            var aggregateRepo = packageRepository as AggregateRepository;
+            if (aggregateRepo != null)
+            {
+                isRemote = aggregateRepo.Repositories.All(repo => repo is IServiceBasedRepository);
+            }
+            else
+            {
+                isRemote = packageRepository is IServiceBasedRepository;
+            }
+
             IQueryable<IPackage> results = packageRepository.Search(configuration.Input, configuration.Prerelease);
 
             if (configuration.AllVersions)
             {
-                return results.Where(PackageExtensions.IsListed).OrderBy(p => p.Id);
+                if (isRemote)
+                {
+                    return results.OrderBy(p => p.Id);
+                }
+                else
+                {
+                    return results.Where(PackageExtensions.IsListed).OrderBy(p => p.Id).AsQueryable();
+                }
             }
 
             if (configuration.Prerelease && packageRepository.SupportsPrereleasePackages)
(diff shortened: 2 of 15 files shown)


src/chocolatey/infrastructure.app/nuget/NugetList.cs
```diff
@@ -54,6 +54,33 @@ public static IEnumerable<IPackage> GetPackages(ChocolateyConfiguration configur
...
+                results = results.Where(p => p.IsLatestVersion);
+            }
+
+            return results
+                        .AsEnumerable()
+                        .Where(PackageExtensions.IsListed)
+                        .Where(p => configuration.Prerelease || p.IsReleaseVersion())
+                        .distinct_last(PackageEqualityComparer.Id, PackageComparer.Version)
+                        .Count();
+        }
```
ferventcoder: This method could literally call GetPackages(config, nugetLogger).Count(). Thoughts?

ferventcoder:
This needs a ticket filed on it.

ferventcoder:
So, this is interesting. Two calls out then? :/

RichiCoder1:
Only when paging is used, and when it is, Count will be called once while List will be called multiple times with given constraints.

src/chocolatey/infrastructure.app/nuget/NugetList.cs
RichiCoder1: If they specified paging, then Count() would be inaccurate. The though process here is you could use the same config for both `List` and `Count`, and count would return the correct result.

RichiCoder1:
I also discovered that the way we handle filtering out unlisted packages is straight up wrong. Looking into fixing that.

src/chocolatey/infrastructure.app/nuget/NugetList.cs
ferventcoder: You could change the config momentarily. I'm more of a fan of that than I am duplicating code.

RichiCoder1:
Hrmm. Chocolatey's nuget is out of date, even for v2. They added and explicit switch on `search` to handle delisted packages.

ferventcoder:
> Hrmm. Chocolatey's nuget is out of date, even for v2. They added and explicit switch on search to handle delisted packages.

They also broke other things. That's why we are back on 2.8.2

RichiCoder1:
>>Hrmm. Chocolatey's nuget is out of date, even for v2. They added and explicit switch on search to handle delisted packages.

>They also broke other things. That's why we are back on 2.8.2

Are you up to date with nuget-chocolatey too? It has the extensions I need to make this work efficiently. Failing that, I'm going to have to drop to a lower level and just build the query myself.

ferventcoder:
Take a look at the 2.8.2_adds branch and submit fixes there

RichiCoder1:
Yup, just realized I was looking at the wrong branch.

RichiCoder1:
I was hoping that this wouldn't require a server change but it looks like NuGet's OData doesn't support datetime comparsions in it's IQueryable, and Chocolatey's server claims that IsListed doesn't exist. Server needs to either support an IsListed property or add isListed to it's search.

RichiCoder1:
nuget-chocolatey doesn't support searching against unlisted packages. Why was that check even there?

RichiCoder1:
So the isListed filter was completely superfluous unless I'm missing something. The results from the choco I just pushed return identical to the results from Choco v0.9.9.

Count now returns in an average of 1-2 seconds, and using Page is ever so slightly faster too.

src/chocolatey/infrastructure.app/nuget/NugetList.cs
```diff
@@ -26,12 +31,23 @@ public static class NugetList
     {
         public static IEnumerable<IPackage> GetPackages(ChocolateyConfiguration configuration, ILogger nugetLogger)
         {
+            return SearchPackagesImpl(configuration, nugetLogger);
+        }
+
+
+        public static int GetCount(ChocolateyConfiguration configuration, ILogger nugetLogger)
+        {
+            return SearchPackagesImpl(configuration, nugetLogger).Count();
```
RichiCoder1: If you're wondering why I put `Count()` here, it's because I need IQueryable's Count. Otherwise it pulls down all the results, and gets the count of that.

src/chocolatey/infrastructure.app/nuget/NugetList.cs
```diff
@@ -48,12 +64,11 @@ public static IEnumerable<IPackage> GetPackages(ChocolateyConfiguration configur
                 results = results.Skip(configuration.ListCommand.PageSize * configuration.ListCommand.Page.Value).Take(configuration.ListCommand.PageSize);
             }
 
-            return results.OrderBy(p => p.Id)
-                        .AsEnumerable()
-                        .Where(PackageExtensions.IsListed)
-                        .Where(p => configuration.Prerelease || p.IsReleaseVersion())
```
RichiCoder1: Both the above checks were completely superfluous new as I can tell. The former was redundant with checks the server is already doing, and the later is redundant with the IsPrelease check.

ferventcoder: Please ensure you are testing against multiple NuGet source types.

src/chocolatey/infrastructure.app/nuget/NugetList.cs
```diff
@@ -26,12 +31,23 @@ public static class NugetList
...
+            return SearchPackagesImpl(configuration, nugetLogger);
+        }
+
+
+        public static int GetCount(ChocolateyConfiguration configuration, ILogger nugetLogger)
+        {
+            return SearchPackagesImpl(configuration, nugetLogger).Count();
+        }
+
+        private static IQueryable<IPackage> SearchPackagesImpl(ChocolateyConfiguration configuration, ILogger nugetLogger)
```
ferventcoder: Please do not use `Impl`. :no_bell:

RichiCoder1: Thought you might call that out :). What's the recommended naming here?

src/chocolatey/infrastructure.app/nuget/NugetList.cs
RichiCoder1: Do you have any other recommended NuGet sources I should test against besides Chocolatey?

ferventcoder:
> I was hoping that this wouldn't require a server change but it looks like NuGet's OData doesn't support datetime comparsions in it's IQueryable, and Chocolatey's server claims that IsListed doesn't exist. Server needs to either support an IsListed property or add isListed to it's search.

We should add it.

> nuget-chocolatey doesn't support searching against unlisted packages. Why was that check even there?

I was not aware that nuget could support searching against unlisted packages. Does it allow it?

src/chocolatey/infrastructure.app/nuget/NugetList.cs
ferventcoder: * File source - which you get with list -lo
* NuGet.org itself
* And maybe a myget source if you can get your hands on one.

Earlier when @Jaykul was working on some of this that was similar I could not get the results to match. It wasn't necessarily the count, although it was off as well. It was the order coming back from the query, it was off. https://github.com/chocolatey/choco/pull/224#discussion_r28621540 and https://github.com/chocolatey/choco/pull/224#issuecomment-116647818

ferventcoder: What I'd really love to see before someone starts changing these core bits is some integration tests that really test this out (prior to making the changes).

src/chocolatey/infrastructure.app/nuget/NugetList.cs
ferventcoder: Never `Impl`. It adds zero value to the method in my opinon. Resharper didn't call you out on the naming b/c of where you were for this method. 

My preference is something like `search_source_for_packages` or something like that. Naming is hard. It's easy to see when it is wrong, not always easy to call out what it should look like. ;)

src/chocolatey/infrastructure.app/nuget/NugetList.cs
RichiCoder1: Absolutely agree. Something I was thinking about. Just have to figure out how to reliable intregration test this :)

RichiCoder1: > Earlier when @Jaykul was working on some of this that was similar I could not get the results to match. It wasn't necessarily the count, although it was off as well. It was the order coming back from the query, it was off. #224 (comment) and #224 (comment)

Reviewing them, I'm going to run into the exact same issue, but I'm also unwilling to add distinct_last back. I'd much rather find a way to detect whether if the underlying provider is sourcing from a server than make Search a blocking operation.

RichiCoder1:
>I was not aware that nuget could support searching against unlisted packages. Does it allow it?

Nope. That was me misunderstanding something.

RichiCoder1:
@ferventcoder developed a somewhat intelligent work around. Remote servers, aka OData chocolatey servers, seem to be very uniform and with my changes produce the same results. Other sources however behave....odd. So I now check if all the repositories we're searching against are remote. If they are, I let the server handle the heavy lifting. If not, we run the blocking filters that produce the correct results. The end result is `--page` and `Count()` are blazing fast against OData Servers, and produce the correct behavior elsewhere.

src/chocolatey/infrastructure.app/nuget/NugetList.cs
```diff
@@ -26,12 +27,47 @@ public static class NugetList
...
+        }
+
+        private static IQueryable<IPackage> execute_package_search(ChocolateyConfiguration configuration, ILogger nugetLogger)
+        {
             var packageRepository = NugetCommon.GetRemoteRepository(configuration, nugetLogger);
+
+            // Whether or not the package is remote determines two things:
+            // 1. Does the repository have a notion of "listed"?
+            // 2. Does it support prerelease in a straight-forward way?
+            // Choco previosly dealt with this by taking the path of least resistance and manually filtering out and sort unwanted packages
```
ferventcoder: typo

src/chocolatey/infrastructure.app/nuget/NugetList.cs
```diff
@@ -43,17 +79,25 @@ public static IEnumerable<IPackage> GetPackages(ChocolateyConfiguration configur
...
+            if (!isRemote)
+            {
+                results =
+                    results
+                        .Where(PackageExtensions.IsListed)
+                        .Where(p => configuration.Prerelease || p.IsReleaseVersion())
+                        .distinct_last(PackageEqualityComparer.Id, PackageComparer.Version)
+                        .AsQueryable();
+            }
+
```
ferventcoder: I can live with this :)

ferventcoder:
Do you want to start squashing down/cleaning up your commits and adding the prepend `(GH-431)` and notes to the commits?

ferventcoder:
Example of what I'm looking for with commit summary and message body - https://github.com/chocolatey/choco/commit/e15c3ebccdf23f3bf509ad9d767eb7898d0be4a9

RichiCoder1:
Yup! Wanted to get something we're happy with before cleaning up my commits :)

RichiCoder1:
Squashed.

RichiCoder1:
(and fixed type now that I think of it)

ferventcoder:
I thought I saw at least two distinct commits in there. For squashing, I meant like the commit "Semicolon" and others

RichiCoder1:
Hows that?

ferventcoder:
Looks good! Now if there was some integration testing surrounding this, I wouldn't feel a need to manually test it. I'll pull it in for now and we can adjust if we need to later. :)

src/chocolatey/GetChocolatey.cs
```diff
@@ -222,6 +222,15 @@ public IEnumerable<T> List<T>()
             var runner = new GenericRunner();
             return runner.list<T>(configuration, _container, isConsole: false, parseArgs: null);
         }
+
+        public int Count()
```
ferventcoder: Actually, the name of this throws me. It's not clear what this method returns.

This method needs some xml documentation with it as well (looks like I missed calling that out in the list command PR).

ferventcoder: What about ListCount()?

RichiCoder1: Hrmm. Normally I'd be against it, but I list ListCount works. Or rather `GetListCount()`

RichiCoder1: Done.

ferventcoder: GetListCount sounds fine to. Count isn't clear on what it is doing, that's why I was hoping for more. It would be different if you were calling `GetChocolatey().Setup().List().Count

ferventcoder:
Commit cleanup?

RichiCoder1:
Ohh, and we're all green again :D

ferventcoder:
Thanks!

Issue #396, opened 2015-08-23
Helper for retrieving installed applications list


justdanpo:
About 60 packages use "Get-WmiObject -Class Win32_Product" to check if a package is already installed https://www.google.com/search?q=Win32_Product%20site%3Achocolatey.org

"Win32_Product" class has serious performance issues https://www.google.com/search?q=win32_product%20issues 
It forces Windows Installer to reconfigure every application installed through MSI (see Windows event log).
Depending on installed applications it may take more than 30 minutes to execute a request.
It may be replaced by direct registry access and a required code is large enough to create a suitable helper.

Here is an example of such helper:

```
## Get-Installed
##############################################################################################################
## Returns a list of installed applications.
## Registry based replacement for "Get-WmiObject -Class Win32_Product".
##############################################################################################################

function Get-Installed {
param(
  [string] $NonMSIOption = 'HideNonMSI' # You may use 'ShowNonMSI' option
)

  if(($NonMSIOption -ine 'HideNonMSI') -and ($NonMSIOption -ine 'ShowNonMSI')) {
    throw "Unsupported argument: $NonMSIOption"
  }


  $items = @()

  if(Test-Path "HKLM:SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall") {
    $items += Get-ChildItem "HKLM:SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall"
  }
  if(Test-Path "HKLM:SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall") {
    $items += Get-ChildItem "HKLM:SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall"
  }

  $Win32_Product = @()

  foreach ($app in $items)
  {
    $propitems = $app | Foreach-Object {Get-ItemProperty $_.PsPath}
    $w32item = New-Object System.Object
    $w32item | Add-Member -type NoteProperty -Name "IdentifyingNumber" -Value $app.PsChildname
    $w32item | Add-Member -type NoteProperty -Name "Name"    -Value $propitems.DisplayName
    $w32item | Add-Member -type NoteProperty -Name "Version" -Value $propitems.DisplayVersion
    $w32item | Add-Member -type NoteProperty -Name "Vendor"  -Value $propitems.Publisher
    $w32item | Add-Member -type NoteProperty -Name "Caption" -Value $propitems.DisplayName

    if($w32item.Name) {
[...]

PR #441, opened 2015-10-01
(#396) Get-Installed helper

justdanpo:
https://github.com/chocolatey/choco/issues/396

Direct replacement for "Get-WmiObject -Class Win32_Product".
Win32_Product does lots of unnecessary work and may be VERY slow. Some
of packages use it to check if a package is already installed.

This change adds Get-Installed helper and a piece of code that
automatically fixes install/uninstall scripts which use Win32_Product.


ferventcoder:
This is a good start. However, instead of rewriting files, I would see this as a setting that is for not using the wmi objects, then overriding the function itself and using the setting to determine which way to go.  For an example of overriding the function, see this prior art https://github.com/chocolatey/chocolatey/blob/master/src/helpers/functions/Write-Host.ps1

justdanpo:
But do we really need a setting?

ferventcoder:
You can start out with no setting if you want. The setting is not the important part, it's not rewriting package files.

src/chocolatey.resources/helpers/functions/Get-Installed.ps1
```diff
@@ -0,0 +1,46 @@
+﻿## Get-Installed
+##############################################################################################################
+## Returns a list of installed applications.
+## Registry based replacement for "Get-WmiObject -Class Win32_Product".
+##############################################################################################################
+
+function Get-Installed {
```
ferventcoder: Does this work without specifying any args? That would be a definite improvement on what I was doing before.

src/chocolatey.resources/helpers/functions/Get-WmiObject.ps1
```diff
@@ -0,0 +1,15 @@
+﻿## Get-WmiObject
+##############################################################################################################
+## Wrapper function. Replaces "Win32_Product" class with Get-Installer helper.
+##############################################################################################################
+
+function Get-WmiObject {
+
+  $oc = Get-Command 'Get-WmiObject' -Module 'Microsoft.PowerShell.Management'
```
ferventcoder: Is this the name of where the module is?

src/chocolatey.resources/helpers/functions/Get-WmiObject.ps1
```diff
@@ -0,0 +1,15 @@
+﻿## Get-WmiObject
+##############################################################################################################
+## Wrapper function. Replaces "Win32_Product" class with Get-Installer helper.
+##############################################################################################################
+
+function Get-WmiObject {
```
ferventcoder: Does this work without specifying any params? That would be a definite improvement on what I was doing before.

src/chocolatey.resources/helpers/functions/Get-WmiObject.ps1
justdanpo: C:\15_09_30>powershell -command "get-command get-wmiobject"
    CommandType     Name                                               ModuleName
    -----------     ----                                               ----------
    Cmdlet          Get-WmiObject                                      Microsoft.PowerShell.Management

ferventcoder: :+1:

src/chocolatey.resources/helpers/functions/Get-Installed.ps1
```diff
@@ -0,0 +1,46 @@
+﻿## Get-Installed
+##############################################################################################################
+## Returns a list of installed applications.
+## Registry based replacement for "Get-WmiObject -Class Win32_Product".
+##############################################################################################################
+
+function Get-Installed {
+param(
+  [string] $NonMSIOption = 'HideNonMSI' # You may use 'ShowNonMSI' option
```
ferventcoder: What is this bit? If we are getting installed applications, that's everything.

src/chocolatey.resources/helpers/functions/Get-Installed.ps1
```diff
@@ -0,0 +1,46 @@
...
+  $items = @()
+
+  if(Test-Path "HKLM:SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall") {
+    $items += Get-ChildItem "HKLM:SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall"
+  }
+  if(Test-Path "HKLM:SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall") {
+    $items += Get-ChildItem "HKLM:SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall"
+  }
+
+  $Win32_Product = @()
```
ferventcoder: Naming, lets call it something other than this.

src/chocolatey.resources/helpers/functions/Get-Installed.ps1
```diff
@@ -0,0 +1,46 @@
...
+    $w32item | Add-Member -type NoteProperty -Name "Caption" -Value $propitems.DisplayName
+
+    if($w32item.Name) {
+      if(($NonMSIOption -ieq 'ShowNonMSI') -or ($propitems.WindowsInstaller -eq "1")) {
+        $Win32_Product += $w32item
+      }
+    }
+  }
+
+  return $Win32_Product
```
ferventcoder: Remember this is piped to the next part of the command, so make sure the output is similar.

src/chocolatey.resources/helpers/functions/Get-WmiObject.ps1
justdanpo: Wrapper function may take any amount of arguments. It's just passes them to original function if there are no "Win32_Product". When no arguments are passed, the original function asks a user to input at least "class" argument.

src/chocolatey.resources/helpers/functions/Get-Installed.ps1
justdanpo: When no arguments are passed Get-Installed returns only applications installed by MSI.

    Get-Installed ShowNonMSI

With ShowNonMSI argument it returns extended list which includes applications installed by other installation systems. But some fields for those applications may be empty or set in different format.
Exampes:
MSI

    IdentifyingNumber : {FDAF247B-4ADD-4089-A799-B08206B8E75C}
    Name              : Adobe Flash Player 19 ActiveX
    Version           : 19.0.0.185
    Vendor            : Adobe Systems Incorporated
    Caption           : Adobe Flash Player 19 ActiveX

NonMSI:

    IdentifyingNumber : Декларация 2013
    Name              : Декларация 2013
    Version           : 
    Vendor            : 
    Caption           : Декларация 2013

    IdentifyingNumber : {FC712CA0-A945-11d4-A594-956F6349FC18}
    Name              : True Launch Bar
    Version           : 6.6
    Vendor            : Tordex
    Caption           : True Launch Bar

src/chocolatey.resources/helpers/functions/Get-Installed.ps1
rismoney: win32_product can be a nasty animal.  the general rule is to avoid it if possible.
https://support.microsoft.com/en-us/kb/974524.

src/chocolatey.resources/helpers/functions/Get-Installed.ps1
ferventcoder: It should behave very similarly to Get-WMIObject.

src/chocolatey.resources/helpers/functions/Get-Installed.ps1
ferventcoder: @rismoney yep, that's exactly what @justdanpo is giving us here. An explicit override to not use it at all. :)

src/chocolatey.resources/helpers/functions/Get-Installed.ps1
justdanpo: Wrapped Get-WMIObject behaves similarly. Get-Installed is another thing and it have the option.

src/chocolatey.resources/helpers/functions/Get-WmiObject.ps1
ferventcoder: That's awesome. Wish I would have thought about that with Write-Host, Write-Error and the like. :)

src/chocolatey.resources/helpers/functions/Get-Installed.ps1
ferventcoder: Whatever normally comes out of Get-WmiObject should be what the wrapper that calls this produces.

ferventcoder: By the way, it looks like this is a boolean?

src/chocolatey.resources/helpers/functions/Get-Installed.ps1
```diff
@@ -0,0 +1,46 @@
...
+##############################################################################################################
+
+function Get-Installed {
+param(
+  [string] $NonMSIOption = 'HideNonMSI' # You may use 'ShowNonMSI' option
+)
+
+  if(($NonMSIOption -ine 'HideNonMSI') -and ($NonMSIOption -ine 'ShowNonMSI')) {
+    throw "Unsupported argument: $NonMSIOption"
+  }
```
ferventcoder: Use a boolean/switch instead and you only have true or false. So no need to throw anything here.

src/chocolatey.resources/helpers/functions/Get-Installed.ps1
justdanpo: The only difference is properties order. This doesn't play a role as far as I understand.

ferventcoder: I think the next command usually does the filtering by named properties.

src/chocolatey.resources/helpers/functions/Get-Installed.ps1
justdanpo: I thought "Get-Installed ShowNonMSI" looks more descriptively than "Get-Installed $true" or "Get-Installed $false".

ferventcoder: Name your custom boolean ;)

~~~powershell
$showAll = $true
Get-Installed $showAll
~~~

ferventcoder: So dumb question from Chocolatey's perspective, when would I ever NOT want to see all installed applications?

wjk: @justdanpo: If you want, you can use a PowerShell `[switch]` parameter instead. A PowerShell switch is used just like a boolean value in checks. However, the syntax for supplying the parameter is different. For example, suppose you have a switch parameter called `ShowNonMSI` that defaults to `$false`. Here is how you would use it:

Usage|Result
---------|--------
`Get-Installed` | `$ShowNonMSI` is `$false`
`Get-Installed -ShowNonMSI` | `$ShowNonMSI` is `$true`

You would declare the parameter like this:

```powershell
param(
    [switch]$ShowNonMSI = $false
)
```

Does this help?

justdanpo: @wjk: thank you!
@ferventcoder: wrapped Get-WmiObject should be compatible with original one.
Would it be better to return non-MSI applications by default and suppress them in Get-WmiObject?

ferventcoder:
Would you mind updating that second commit to conform to our contributing guidelines? Unless you are planning to squash that?

justdanpo:
Should I change anything else?

ferventcoder:
Thanks. I think it's ready for a deeper review now.

DarwinJS:
FYI Win32_Product enumeration triggers a some minor processing on every install MSI package.  You can see it if you enabled MSI verbose logging.  I blogged this here: http://csi-windows.com/toolkit/win32product-replacement (the code in the article is not appropriate for this).

src/chocolatey.resources/helpers/functions/Get-Installed.ps1
```diff
@@ -0,0 +1,48 @@
...
+  $productList = @()
+
+  foreach ($app in $items)
+  {
+    $propitems = $app | Foreach-Object {Get-ItemProperty $_.PsPath}
+    $w32item = New-Object System.Object
+    $w32item | Add-Member -type NoteProperty -Name "IdentifyingNumber" -Value $app.PsChildname
+    $w32item | Add-Member -type NoteProperty -Name "Name"    -Value $propitems.DisplayName
+    $w32item | Add-Member -type NoteProperty -Name "Vendor"  -Value $propitems.Publisher
+    $w32item | Add-Member -type NoteProperty -Name "Version" -Value $propitems.DisplayVersion
```
ferventcoder: This is likely going to need a bit of work. DisplayVersion is not required.

ferventcoder: InnoSetup doesn't use it.

justdanpo: This fix is mainly intended to be a replacement for Win32_Product which doesn't know about innosetup at all. Win32_Product works with Windows Installer (MSI).
Hooked Get-WmiObject function returns a list fully compatible with original function's output except an items order. Fields returned by Win32_Product:

- IdentifyingNumber
- Name
- Vendor
- **Version**
- Caption (allways the same as Name)

Get-Installed (called from Get_WmiObject) is able to return non-MSI applications and it does so when -GWMICompatible switch is **not** used. It had a switch with the opposite meaning in previous commits but this has been changed during a conversation (see outdated diffs).
Also InnoSetup MAY use DisplayVersion. Here is a part of output:

    IdentifyingNumber : Git_is1
    Name              : Git version 2.5.0
    Vendor            : The Git Development Community
    Version           : 2.5.0
    Caption           : Git version 2.5.0
    WindowsInstaller  : 
    UninstallString   : "C:\Program Files\Git\unins000.exe"

Why two functions?

1. Get-Installed can return MSI and non-MSI records. It may be used in packages installation/uninstallation scripts.
2. Get_WmiObject can return only MSI records. It transparently replaces Win32_Product calls with Get-Installed so there is no need to fix more than 60 existing (and may be some future) packages which use Win32_Product.

ferventcoder: Basicallly MSIs can use `Version` and/or `DisplayVersion`. Not all MSIs I queried on my system use `DisplayVersion`.

ferventcoder: The proper order for the query is probably going to be converting `Version` to a string, `DisplayVersion` if it is not available, then converting `VersionMajor`\`VersionMinor` to strings if the other information is not available.

ferventcoder: Starting to look at what is necessary so this change can come in for 0.9.10. What happens when those prop items are missing? I'm assuming it doesn't error?

DarwinJS: @ferventcoder - I have done code for this before - ran and was debugged for 14000 machines in 17 MUI language on Win XP and Win 7.

In that code I tried to reconcile win32_product with this information - you basically have to link it by product code which can be parsed out of the ARP data.  But for my purposes it looked like that road was going to be very bumpy.  For once think if there is a master MSI and child MSIs - more than one ARP item can carry the product code.  Sometimes the same was true when a setup.exe contains an MSI.  So for my more limited purposes I just parsed Product Code from ARP data and didn't bother with contacting MSI.

Here are some other challenges / issues I ran into and codeded around (with no connotations of what would be the appropriate approach for chocolatey):
- there is a third per-user location for uninstall.  Kindle reader is one thing I've found in it before.  Exist  check it and pull it into the record list if it exists.
- raw uninstall keys contain some hidden items - created a switch to include hidden items rather than have them in the list all the time
- raw uninstall keys contain parent / child relationships not normally shown in Programs and Features - created a switch to include them rather than retrieve all the time.
- raw uninstall keys can contain windows updates KB article updates that don't normall show in Programs and Features - created a switch
- occasionally got blank records when retrieving from this key - used a filter to drop them after retrieving all three locations.
- it is possible, but infrequent, for this code to choke when poorly formed uninstall keys have been laid down by a vendor.  There was some java dev kit that was laying down "size" in a bizzare format and it would choke.  Found others had reported it to the vendor.  We only had about 4 affected machines in 14,000 - so just corrected those.  (this is probably extremely rare - but when the code chokes on it - it's hard for the individual to figure out why)
- The code provides three derived properties for GUIDfromUninstallString, KBGUIDFromKeyName, KBIDFromKeyName

For KBs the DisplayName will be null.  For childitems ParentKeyName will be populated.  Many times the parent item accomplishes an uninstall of child items - this is also what the "show updates" checkbox in Programs and Features" uses.

So I agree with your suspicion that this code would need some work to not generate any errors on the millions of nodes running Chocolatey.  I agree that the code should be updated - but you probably want to get a prototype out there to flush out the above issues and any others and transition to the new function after some debug time has occurred in the real world.

ferventcoder: A prototype, and as opt-in. Exactly what I had stated before. Allow folks to opt into newer behavior that may not fully work as expected.

justdanpo: @ferventcoder 
> I took a deeper look - MSIs can use Version and/or DisplayVersion. Not all MSIs I queried on my system use DisplayVersion.
Could you give examples of such MSIs, how their records are stored in the registry and how they are returned from Win32_Product?

pauby:
@DarwinJS @justdanpo I appreciate this is a _long_ time ago, and I apologise that we are only getting to some of these much older PR's. You may not remember the context around this, but I need to ask this question nonetheless.

This is solving a problem where a package that contains `Get-WmiObject -Class Win32_Product` is very slow and is providing a faster alternative? 

If that is the case, this is useful _only for packages that contain this code_?

justdanpo:
@pauby When I made this there were about 70 Chocolatey packages which used Get-WmiObject for detecting currently installed software version. Installing/updating ANY of these packages was taking tens of minutes for me (on several different PCs).
This pr contains a hook which temporarily replaces Get-WmiObject with its own custom implementation returning software list by accessing Windows registry. That had been helping me for a long time.

I don't know if any of packages still use Get-WmiObject. As far as I remember there was an idea to use some kind of static analyzers to check packages and encourage package maintainers not to use Get-WmiObject.
I don't experience this issue now, maybe because of upgraded Windows version, maybe because of different list of installed packages, maybe because of packages I use don't use Get-WmiObject anymore.

pauby:
I spoke with @gep13 earlier this week (which prompted my question to you) and in our experience we don't see packages that use `Get-WmiObject`. So our thinking that this wasn't needed, but we wanted to clarify with you the reasons for submitting it ass it was a _long_ time ago.

As we don't think this is needed anymore, I'm going to go ahead and close it. Thanks for submitting the PR and coming back to me so quickly @justdanpo

gep13:
Just to add something to this as well...  we have a [package-validator rule](https://docs.chocolatey.org/en-us/community-repository/moderation/package-validator/rules/cpmr0059/#mainContent) to detect if Get-WmiObject is used, and suggest that it should be avoided.

PR #443, opened 2015-10-02
(GH-442) add conditional to prevent removal of unloaded modules

### LINKED ISSUE #442: Chocolatey Installer - removing modules not loaded
the chocolateyInstall script attempts to remove modules that are not necessarily loaded.
The current solution  does a dirty -erroraction -silentlycontinue.

i am proposing putting a conditional to eliminate output within stderr.
will follow up with a PR.

rismoney:
this is part of a cleaning up content that might be contained
within stderr.

Closes #442

### DIFF
--- a/nuget/chocolatey/tools/chocolateyInstall.ps1
+++ b/nuget/chocolatey/tools/chocolateyInstall.ps1
@@ -7,7 +7,9 @@ $modules = Get-ChildItem $toolsPath -Filter *.psm1
 $modules | ForEach-Object {
 														$psm1File = $_.FullName;
 														$moduleName = $([System.IO.Path]::GetFileNameWithoutExtension($psm1File))
-														remove-module $moduleName -ErrorAction SilentlyContinue;
+														if (Get-Module $moduleName) {
+                              remove-module $moduleName -ErrorAction SilentlyContinue;
+                            }
 														import-module -name  $psm1File;
 													}
 


nuget/chocolatey/tools/chocolateyInstall.ps1
```diff
@@ -7,7 +7,9 @@ $modules = Get-ChildItem $toolsPath -Filter *.psm1
 $modules | ForEach-Object {
 														$psm1File = $_.FullName;
 														$moduleName = $([System.IO.Path]::GetFileNameWithoutExtension($psm1File))
-														remove-module $moduleName -ErrorAction SilentlyContinue;
+														if (Get-Module $moduleName) {
```
ferventcoder: Am I misunderstanding in that this will actually fail if there is no module?

ferventcoder: Actually it looks like it should not error, but is it truthy?

ferventcoder: This is done twice I believe.

ferventcoder: Sorry, I mean this is done more times down in the actual module as well.

ferventcoder:
Merged into stable at https://github.com/chocolatey/choco/commit/8840209c01bc6f30718ac1ba4f57d695955566ef

PR #461, opened 2015-10-12
(GH-460) (maint) Amended Template

### LINKED ISSUE #460: Choco New creates a bad ChocolateyUninstall.ps1 script which does not work.
When using Choco New to create a new ChocolateyInstall.ps1 and ChocolaetyUninstall.ps1 script the uninstall one is bad. See below:

<code>
$key = Get-ItemProperty -Path @($machine_key6432,$machine_key, $local_key) `
                        -ErrorAction SilentlyContinue `
         | ? { $_.DisplayName -like "$softwareName" }

if ($key.Count -eq 1) {
...
</code>

That if statement will never execute as when $key only has one value it's not an Array so doesn't have a Count method. The fix is to force $key to be an array:

<code>
[array]$key = Get-ItemProperty -Path @($machine_key6432,$machine_key, $local_key) `
                        -ErrorAction SilentlyContinue `
         | ? { $_.DisplayName -like "$softwareName" }

if ($key.Count -eq 1) {
</code>

I would just submit a pull request but I have no idea where it gets the script or if it generates it on the fly.

pauby:
Using 'choco new' to produce ChocolateyUninstall.ps1 for your package would result in the uninstall code never being executed as the $key variable type was wrong. Updated the script template to force the $key variable to always be an array and therefore have a Count method that can be evaluated.

Closes #460

### DIFF
--- a/src/chocolatey/infrastructure.app/templates/ChocolateyUninstallTemplate.cs
+++ b/src/chocolatey/infrastructure.app/templates/ChocolateyUninstallTemplate.cs
@@ -58,7 +58,7 @@ $local_key     = 'HKCU:\Software\Microsoft\Windows\CurrentVersion\Uninstall\*'
 $machine_key   = 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\*'
 $machine_key6432 = 'HKLM:\SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall\*'
 
-$key = Get-ItemProperty -Path @($machine_key6432,$machine_key, $local_key) `
+[array]$key = Get-ItemProperty -Path @($machine_key6432,$machine_key, $local_key) `
                         -ErrorAction SilentlyContinue `
          | ? { $_.DisplayName -like ""$softwareName"" }
 


ferventcoder:
Super simple.

ferventcoder:
I just need you to read over our CONTRIBUTING document and then fix up the commit message.

pauby:
Amended the message - is this more suitable?

This is my second pull request so be gentle with me! I did read the CONTRIBUTING file but only got as far as Prerequisites which obviously hasn't helped!

gep13:
@pauby thanks for your contribution to chocolatey, it is really appreciated!

One of the main things that @ferventcoder is getting at is the actual commit message (which is currently `Amended Template`).  Ideally, this will include the issue number that is related to the commit.  i.e. in the same way as you have changed the name of this Pull Request to `(GH-460) Amended Template` we would actually want you to change the commit message as well.  In addition, the description of the commit message should include some details about what was actually done, and why, as described in the contributing.md file.

In Git, this is really simply using the rebase command.  Have you used this before?  If not, there are details in the contributing.md file which will help you.

Does that make sense?

ferventcoder:
Rebase can be a bit of a scary thing for new folks as it rewrites history a bit, but it's definitely what we want here.

The nice thing is that you can cheat a little since you only have the one commit - 

* `git commit --amend`
* fixup the commit message
* `git push origin pauby/feature_branch_name -f`

I see you made this from your master branch, that's going to be an issue. You may want to do this instead:

* `git checkout -b fixup_uninstall_template`
* `git push origin fixup_uninstall_template` 
* open a pull request against master with that instead.

ferventcoder:
You will still need to do `git commit --amend`. And then later you will need to back that commit off of your master branch and force push your master branch. Force pushes are generally acceptable in forks, usually only in feature branches though. Your other option is to wait for the PR to be accepted and then delete and recreate your fork and pull it locally again.

ferventcoder:
http://git-scm.com/book/en/v2/Git-Tools-Rewriting-History is a good place to learn about rebasing and fixing up commits.

ferventcoder:
You will not be able to push your new branch to this PR, so I'm going to close this pending the new push. Let me know if you have any questions! :)

PR #481, opened 2015-11-17
(GH-399) client certificate support

### LINKED ISSUE #399: Support for client certificates
I don't see any way to configure choco with a client certificate for authentication with a server.
Both apt-get and yum support this, so would be nice if choco could do the same.

et1975:
Please consider this merge for #399.

It requires chocolatey/nuget-chocolatey/pull/3/ to be pre-built and included as well.

Closes #399


src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
```diff
@@ -384,6 +384,8 @@ public sealed class SourcesCommandConfiguration
         public string Username { get; set; }
         public string Password { get; set; }
         public int Priority { get; set; }
+        public string Certificate { get; set; }
+        public string CertificatePassword { get; set; }
```
ferventcoder: Will you have both a user name / password and a certificate / password?

In cases like this maybe it makes sense that EncryptedPassword is used for both.

src/chocolatey/infrastructure.app/nuget/NugetCommon.cs
```diff
@@ -58,6 +58,7 @@ public static IPackageRepository GetRemoteRepository(ChocolateyConfiguration con
 
             // ensure credentials can be grabbed from configuration
             HttpClient.DefaultCredentialProvider = new ChocolateyNugetCredentialProvider(configuration);
+            HttpClient.DefaultCertificateProvider = new ChocolateyClientCertificateProvider(configuration);
```
ferventcoder: Doesn't this override the line right before it?

src/chocolatey/infrastructure.app/services/ChocolateyConfigSettingsService.cs
```diff
@@ -55,14 +55,14 @@ public IEnumerable<ChocolateySource> source_list(ChocolateyConfiguration configu
...
-                        string.IsNullOrWhiteSpace(source.UserName) ? string.Empty : "(Authenticated)",
+                        string.IsNullOrWhiteSpace(source.UserName) && string.IsNullOrWhiteSpace(source.Certificate) ? string.Empty : "(Authenticated)",
                         source.Priority));
                 }
                 list.Add(new ChocolateySource {
                     Id = source.Id,
                     Value = source.Value,
                     Disabled = source.Disabled,
-                    Authenticated = string.IsNullOrWhiteSpace(source.Password),
+                    Authenticated = !(string.IsNullOrWhiteSpace(source.UserName) && string.IsNullOrWhiteSpace(source.Certificate)), // I believe there was a bug here
```
ferventcoder: I think you may have found a bug. Newer code, used in messaging mostly.

src/chocolatey/infrastructure.app/services/ChocolateyConfigSettingsService.cs
```diff
@@ -55,14 +55,14 @@ public IEnumerable<ChocolateySource> source_list(ChocolateyConfiguration configu
                         source.Id,
                         source.Disabled ? " [Disabled]" : string.Empty,
                         source.Value,
-                        string.IsNullOrWhiteSpace(source.UserName) ? string.Empty : "(Authenticated)",
+                        string.IsNullOrWhiteSpace(source.UserName) && string.IsNullOrWhiteSpace(source.Certificate) ? string.Empty : "(Authenticated)",
```
ferventcoder: Could probably just go off of `password || certificate` here

ferventcoder:
@et1975 have you had a chance to read over CONTRIBUTING? Please do and update your git commit with `(GH-399)` instead of `GH-399` and include a nice git message body about what this change is and why it is necessary, etc. Thanks for your work so far!

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
et1975: I see the two authentication methods as complimentary so it's quite possible for the passwords to be different.

src/chocolatey/infrastructure.app/nuget/NugetCommon.cs
et1975: Same thing here: different layers of authentication, which is actually handy for using existing authorization mechanisms of various servers.

src/chocolatey/infrastructure.app/services/ChocolateyConfigSettingsService.cs
et1975: I thing username is what authenticates, not the password... Or do you mean IsNullOrEmpty instead of WhiteSpace? Just sticking to what was here before to avoid any regressions.

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
ferventcoder: Are you saying sometimes you would authenticate to the ***same*** source with user name / password and sometimes you would authenticate with the certificate / password? Because I'm having a really hard time seeing that.

ferventcoder: I'd also see the above as a bug if it didn't authenticate the *same* way every time.

src/chocolatey/infrastructure.app/services/ChocolateyConfigSettingsService.cs
ferventcoder: Yeah no worries. I'm not sure a username without a password can be put in, so it's likely just that it works by association.

src/chocolatey/infrastructure.app/nuget/NugetCommon.cs
ferventcoder: Unless I'm missing something, this is setting a value EQUAL to the second, and not adding to a list. What i think is needed here is either a list OR a custom meta type that incorporates both of them (this latter bit I think is the right answer).

ferventcoder:
The commit message is much better. It didn't really do line breaks at 72 characters though. We are a bit of a stickler for those.

ferventcoder:
I think you also need to compile and replace nuget-chocolatey so that builds here will pass.

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
et1975: No, like with proxy authentication, you can have different credentials for for different nodes along the way (btw, the certificate password is not and should not be sent along - it's just to open local certificate file). My use case - reverse proxy configured to do TLS and authenticate with certificates - server & client certificates exchange establishes HTTPS, NuGet server behind it - authorizes. Given that HTTPS termination is commonly done with the proxy, the HTTP server behind it won't even get a say in the negotiations.
Hope that makes sense.

et1975:
Are you reviewing disassembled binaries submitted this way? Cause otherwise there's nothing stopping me from packing it full of trojans, is there?

ferventcoder:
@et1975 I won't accept the binary changes. I just want them here to validate that in theory builds will work.

ferventcoder:
Speaking of which, there is a signer there that will use the default strong name key - you'll need to use it or things won't build. Which you probably already have seen.

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
ferventcoder: Shouldn't the certificate be validated with the remote server or am I missing something? 

So what you are stating is that it is likely in a single call for a source it will use both the certificate AND the username/password?

et1975: Yes, the public key from the certificate will be validated, by whatever is doing HTTPS termination, which may or may not be the NuGet server. The certificate validation does not however use password, but the chain of trust and revocation lists. 
Yes, just like there is apikey alongside the username/password, you could have the need for either/both, depending on how the server is implemented.

et1975:
I think the outstanding failures are due to some environmental factors, no?

ferventcoder:
Looks like it:

~~~
Errors and Failures:
1) Test Error : chocolatey.tests.integration.infrastructure.filesystem.DotNetFileSystemSpecs+when_finding_paths_to_executables_with_dotNetFileSystem.GetExecutablePath_should_find_existing_executable
   Should.Core.Exceptions.EqualException : Assert.Equal() Failure
Position: First difference is at position 3
Expected: C:\Windows\system32\cmd.exe
Actual:   C:\windows\system32\cmd.exe
   at chocolatey.tests.integration.infrastructure.filesystem.DotNetFileSystemSpecs.when_finding_paths_to_executables_with_dotNetFileSystem.GetExecutablePath_should_find_existing_executable() in c:\projects\choco\src\chocolatey.tests.integration\infrastructure\filesystem\DotNetFileSystemSpecs.cs:line 67
 
2) Test Error : chocolatey.tests.integration.infrastructure.filesystem.DotNetFileSystemSpecs+when_finding_paths_to_executables_with_dotNetFileSystem.GetExecutablePath_should_find_existing_executable_with_extension
   Should.Core.Exceptions.EqualException : Assert.Equal() Failure
Position: First difference is at position 3
Expected: C:\Windows\system32\cmd.exe
Actual:   C:\windows\system32\cmd.exe
   at chocolatey.tests.integration.infrastructure.filesystem.DotNetFileSystemSpecs.when_finding_paths_to_executables_with_dotNetFileSystem.GetExecutablePath_should_find_existing_executable_with_extension() in c:\projects\choco\src\chocolatey.tests.integration\infrastructure\filesystem\DotNetFileSystemSpecs.cs:line 77

~~~

et1975:
I think I've fixed those tests, but it doesn't look like appveyor pickup on history rewrite, and TC seems to have been broken to begin with.

ferventcoder:
Remove your second commit please. I fixed those test issues.

ferventcoder:
Ping @et1975 - can you remove your second commit here?

et1975:
@ferventcoder I believe I did, the pull request shows only one commit.
Where do you see a problem?

ferventcoder:
Ah right on. The issues lie in rebase - any chance you can rebase your commeits against current? If not I can do it.

src/chocolatey.tests/infrastructure/filesystem/DotNetFileSystemSpecs.cs
```diff
@@ -164,7 +164,7 @@ public void GetExecutablePath_should_find_existing_executable()
             [Fact]
             public void GetExecutablePath_should_find_existing_executable_with_extension()
             {
-                FileSystem.get_executable_path("cmd.exe").ShouldEqual(
+                FileSystem.get_executable_path("cmd.exe").ToLower().ShouldEqual(
```
ferventcoder: weird.

src/chocolatey/infrastructure.app/nuget/NugetCommon.cs
ferventcoder: I totally missed that these are different properties. Nevermind, nothing to see here.

src/chocolatey.tests/infrastructure/filesystem/DotNetFileSystemSpecs.cs
et1975: I've seen \Windows and \windows returned, I guess it depends on how OS was installed...

ferventcoder: It's more weird in that you are shifting it to lower and asking if it is equal to a string with upppercase.

et1975: Ah, appologies, copy-paste strikes again, although I thought I reverted my
unit-test touches...
Hm, there are two copies of those tests and I guess
chocolatey.tests.infrastructure.filesystem aren't being executed?

On Tue, Jan 5, 2016 at 3:07 PM, Rob Reynolds <notifications@github.com>
wrote:

> In src/chocolatey.tests/infrastructure/filesystem/DotNetFileSystemSpecs.cs
> <https://github.com/chocolatey/choco/pull/481#discussion_r48889951>:
>
> > @@ -164,7 +164,7 @@ public void GetExecutablePath_should_find_existing_executable()
> >              [Fact]
> >              public void GetExecutablePath_should_find_existing_executable_with_extension()
> >              {
> > -                FileSystem.get_executable_path("cmd.exe").ShouldEqual(
> > +                FileSystem.get_executable_path("cmd.exe").ToLower().ShouldEqual(
>
> It's more weird in that you are shifting it to lower and asking if it is
> equal to a string with upppercase.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/chocolatey/choco/pull/481/files#r48889951>.
>



-- 
*Eugene Tolmachev*

Principal Architect

*Prolucid Technologies Inc.*

*190 Robert Speck Parkway, Suite 101 | Mississauga, ON, L4Z 3K3*

*T*: 416.840.6634 | *F*: 416.342.1759 | prolucid.ca
<http://www.prolucid.ca/>


STRICTLY PERSONAL AND CONFIDENTIAL. This email may contain confidential and
proprietary material for the sole use of the intended recipient. Any review
or distribution by others is strictly prohibited. If you are not the
intended recipient please contact the sender and delete all copies.

ferventcoder: Both get executed.

ferventcoder: That's why it's a bit weird to me. :)

et1975: ... pushed the update

ferventcoder:
Merged into stable at https://github.com/chocolatey/choco/commit/026df767592f253aae0ec2a5fcf0b29c794af693

PR #531, opened 2016-01-09
{WIP}(GH-268) Before Modify

### LINKED ISSUE #268: Run a script before uninstall/upgrade to allow for things like services to shutdown
Do not fail if it fails though.

Current proposed name: `chocolateyBeforeRemove.ps1`

Migrated the evolution of the ideas from https://github.com/chocolatey/chocolatey/issues/291

RichardJFoster:
@ferventcoder indicated in the chat room that long-running pull requests were considered desirable, so here is my first attempt at creating a PR for case #268. At this time, this request is more for review than anything else. As Bugs Bunny might observe "Hmm, I knew I should've taken that left turn at Albuquerque.". Specifically while the build may succeed, running the full set of tests DOES NOT (although things do appear to be OK up to changeset f5525312d3ad).

In addition, I am having significant second thoughts about my design. Initially, it made sense to me to separate the selection of the scripts from their execution, but this seems to be different enough from the existing structure that it may cause confusion for future contributors. If I need to start over, just say so.

Closes #268

### DIFF
--- a/src/chocolatey.tests.integration/chocolatey.tests.integration.csproj
+++ b/src/chocolatey.tests.integration/chocolatey.tests.integration.csproj
@@ -225,6 +225,9 @@
     <None Include="context\installpackage\1.0.0\tools\Casemismatch.exe.ignore">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
+    <None Include="context\installpackage\1.0.0\tools\chocolateyBeforeModify.ps1">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
     <None Include="context\installpackage\1.0.0\tools\chocolateyinstall.ps1">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
@@ -327,6 +330,9 @@
     <None Include="context\testing.packages.config">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
+    <None Include="context\upgradepackage\1.0.0\tools\chocolateyBeforeModify.ps1">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
     <None Include="context\upgradepackage\1.0.0\tools\chocolateyinstall.ps1">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
@@ -343,6 +349,9 @@
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
     <None Include="context\upgradepackage\1.0.0\upgradepackage.nuspec">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
+    <None Include="context\upgradepackage\1.1.0\tools\chocolateyBeforeModify.ps1">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
     <None Include="context\upgradepackage\1.1.0\tools\chocolateyinstall.ps1">
--- a/src/chocolatey.tests.integration/scenarios/UninstallScenarios.cs
+++ b/src/chocolatey.tests.integration/scenarios/UninstallScenarios.cs
@@ -219,6 +219,12 @@ namespace chocolatey.tests.integration.scenarios
             {
                 packageResult.Name.ShouldEqual(Configuration.PackageNames);
             }
+
+            [Fact]
+            public void should_have_executed_chocolateyUninstall_script()
+            {
+                MockLogger.MessagesFor(LogLevel.Info).or_empty_list_if_null().Any(p => p.EndsWith("installpackage 1.0.0 Uninstalled")).ShouldBeTrue();
+            }
         }
 
         [Concern(typeof (ChocolateyUninstallCommand))]
--- a/src/chocolatey.tests.integration/scenarios/UninstallScenarios.cs
+++ b/src/chocolatey.tests.integration/scenarios/UninstallScenarios.cs
@@ -95,19 +95,13 @@ namespace chocolatey.tests.integration.scenarios
             [Fact]
             public void should_contain_a_message_that_it_would_have_run_a_powershell_script()
             {
-                bool expectedMessage = false;
-                foreach (var message in MockLogger.MessagesFor(LogLevel.Info).or_empty_list_if_null())
-                {
-                    if (message.Contains("chocolateyuninstall.ps1")) expectedMessage = true;
-                }
-
-                expectedMessage.ShouldBeTrue();
+                MockLogger.contains_message("chocolateyuninstall.ps1").ShouldBeTrue();
             }
 
             [Fact]
             public void should_contain_a_message_that_it_would_have_run_powershell_modification_script()
             {
-                MockLogger.MessagesFor(LogLevel.Info).or_empty_list_if_null().Any(p => p.Contains("chocolateyBeforeModify.ps1")).ShouldBeTrue();
+                MockLogger.contains_message("chocolateyBeforeModify.ps1").ShouldBeTrue();
             }
         }
 
@@ -229,13 +223,13 @@ namespace chocolatey.tests.integration.scenarios
             [Fact]
             public void should_have_executed_chocolateyBeforeModify_script()
             {
-                MockLogger.MessagesFor(LogLevel.Info).or_empty_list_if_null().Any(p => p.Contains("installpackage 1.0.0 Before Modification")).ShouldBeTrue();
+                MockLogger.contains_message("installpackage 1.0.0 Before Modification", LogLevel.Info).ShouldBeTrue();
             }
 
             [Fact]
             public void should_have_executed_chocolateyUninstall_script()
             {
-                MockLogger.MessagesFor(LogLevel.Info).or_empty_list_if_null().Any(p => p.EndsWith("installpackage 1.0.0 Uninstalled")).ShouldBeTrue();
+                MockLogger.contains_message("installpackage 1.0.0 Uninstalled", LogLevel.Info).ShouldBeTrue();
             }
         }
 
--- a/src/chocolatey.tests.integration/scenarios/UninstallScenarios.cs
+++ b/src/chocolatey.tests.integration/scenarios/UninstallScenarios.cs
@@ -103,6 +103,12 @@ namespace chocolatey.tests.integration.scenarios
 
                 expectedMessage.ShouldBeTrue();
             }
+
+            [Fact]
+            public void should_contain_a_message_that_it_would_have_run_powershell_modification_script()
+            {
+                MockLogger.MessagesFor(LogLevel.Info).or_empty_list_if_null().Any(p => p.Contains("chocolateyBeforeModify.ps1")).ShouldBeTrue();
+            }
         }
 
         [Concern(typeof (ChocolateyUninstallCommand))]
@@ -218,6 +224,12 @@ namespace chocolatey.tests.integration.scenarios
             public void config_should_match_package_result_name()
             {
                 packageResult.Name.ShouldEqual(Configuration.PackageNames);
+            }
+
+            [Fact]
+            public void should_have_executed_chocolateyBeforeModify_script()
+            {
+                MockLogger.MessagesFor(LogLevel.Info).or_empty_list_if_null().Any(p => p.Contains("installpackage 1.0.0 Before Modification")).ShouldBeTrue();
             }
 
             [Fact]
--- a/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
+++ b/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
@@ -551,7 +551,9 @@ Would have determined packages that are out of date based on what is
 
             get_environment_before(config, allowLogging: true);
 
-            var packageUpgrades = perform_source_runner_function(config, r => r.upgrade_run(config, action));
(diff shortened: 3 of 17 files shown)


src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
```diff
@@ -550,6 +552,11 @@ public void upgrade_noop(ChocolateyConfiguration config)
             return packageUpgrades;
         }
 
+        private void before_package_upgrade(PackageResult packageResult, ChocolateyConfiguration config)
+        {
+            _powershellService.run_package_scripts_for_command(config, packageResult, CommandNameType.upgrade);
+        }
+
```
ferventcoder: This action is usually run before uninstall as well, I'm imagining you haven't hit that point yet.

src/chocolatey/infrastructure.app/services/PackageScriptService.cs
```diff
@@ -0,0 +1,79 @@
...
+            this.Log().Debug("Locating scripts for command {0} in location {1}", command, packageDirectory);
+
+            switch (command)
+            {
+                case CommandNameType.install:
+                    append_script_if_present(packageDirectory, "chocolateyInstall.ps1", scripts);
+                    break;
+
+                case CommandNameType.upgrade:
+                    append_script_if_present(packageDirectory, "chocolateyBeforeModify.ps1", scripts);
```
ferventcoder: I'm not quite following this - when does chocolateyInstall.ps1 get run?

src/chocolatey/infrastructure.app/services/PowershellService.cs
```diff
@@ -302,9 +292,9 @@ private class PowerShellExecutionResults
             public bool StandardErrorWritten { get; set; }
         }
 
-        private PowerShellExecutionResults run_external_powershell(ChocolateyConfiguration configuration, string chocoPowerShellScript)
+        private PowershellService.PowerShellExecutionResults run_external_powershell(ChocolateyConfiguration configuration, string chocoPowerShellScript)
```
ferventcoder: Why this?

src/chocolatey/infrastructure.app/services/PowershellService.cs
```diff
@@ -503,5 +494,176 @@ private PowerShellExecutionResults run_host(ChocolateyConfiguration config, stri
...
+                    bool shouldRun = !configuration.PromptForConfirmation;
+
+                    if (!shouldRun)
+                    {
+                        this.Log().Info(ChocolateyLoggers.Important, () => "The package {0} wants to run '{1}'.".format_with(package.Id, _fileSystem.get_file_name(chocoPowerShellScript.Path)));
+                        this.Log().Info(ChocolateyLoggers.Important, () => "Note: If you don't run this script, the installation will fail.");
+                        this.Log().Info(ChocolateyLoggers.Important, () => @"Note: To confirm automatically next time, use '-y' or consider setting 
+ 'allowGlobalConfirmation'. Run 'choco feature -h' for more details.");
+
+                        var selection = InteractivePrompt.prompt_for_confirmation(@"Do you want to run the script?", new[] { "yes", "no", "print" }, defaultChoice: null, requireAnswer: true);
```
ferventcoder: I might be seeing some duplication here? Didn't I already see this above? :D

src/chocolatey.tests.integration/chocolatey.tests.integration.csproj
```diff
@@ -219,6 +219,9 @@
     <None Include="context\installpackage\1.0.0\tools\Casemismatch.exe.ignore">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
+    <None Include="context\installpackage\1.0.0\tools\chocolateyBeforeModify.ps1">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
```
ferventcoder: Do we think install package needs it?

RichardJFoster: I believe so. Scripts are only run if they are present. I imagine we should verify that no attempt is made to run the modification script on initial installation.

src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
RichardJFoster: Ah... but in the uninstall scenario both scripts are run from the same package version. In the upgrade scenario one script comes from the old package, and one from the new.

As currently implemented in master/stable, an "upgrade" appears to be converted into what PowershellService sees as an "uninstall" followed by an install of the new version.

Ah... I think I now see what you are worried about though. my current implementation (which I suspect is non functional - I think I goofed somewhere during a rebase) may be running beforeModify twice. I'll check that.

src/chocolatey/infrastructure.app/services/PackageScriptService.cs
RichardJFoster: My concern (and feel free to shoot this down in flames) was/is that the determination of which script(s) to run didn't really seem to fit in PowershellService. By extracting the selection of which script(s) should run on a per-command basis (including setting if script failures terminate execution) I thought it would be easier to create tests for it.

Any applicable script(s) will still be run by PowershellService itself through the run_package_scripts mechanism. As such, it is possible/likely that the current install, noop_install, uninstall, noop_uninstall could/would go away, being replaced by calls to run_package_scripts_for_command.

src/chocolatey/infrastructure.app/services/PowershellService.cs
RichardJFoster: I'm guessing a rebase failure. :-( ReSharper was muttering about it, but I thought that was how it was on master so I had left it there. Will fix.

src/chocolatey/infrastructure.app/services/PowershellService.cs
RichardJFoster: Yes. Yes you did. I was _trying_ not to break the existing code while bringing the new stuff online. Clearly (based on the current test failures) I did not succeed. :-( Basically this falls into a "my fault for not doing a PR earlier" and will need to be cleaned up.

src/chocolatey/infrastructure.app/services/PackageScriptService.cs
RichardJFoster: I should perhaps also clarify... run_package_scripts is intended to run one (or more, in the case of uninstall) scripts against a single package & version (i.e. in a single folder).

src/chocolatey.tests.integration/chocolatey.tests.integration.csproj
ferventcoder: Yes, that makes perfect sense!

ferventcoder:
> In addition, I am having significant second thoughts about my design. Initially, it made sense to me to separate the selection of the scripts from their execution, but this seems to be different enough from the existing structure that it may cause confusion for future contributors. If I need to start over, just say so.

I'm wondering if there might be an easier path here.

src/chocolatey/infrastructure.app/services/PackageScriptService.cs
ferventcoder: Is it that upgrade is run, then install is run later during the new version of the package being installed?

src/chocolatey.tests.integration/scenarios/UninstallScenarios.cs
```diff
@@ -219,6 +219,18 @@ public void config_should_match_package_result_name()
             {
                 packageResult.Name.ShouldEqual(Configuration.PackageNames);
             }
+
+            [Fact]
+            public void should_have_executed_chocolateyBeforeModify_script()
+            {
+                MockLogger.MessagesFor(LogLevel.Info).or_empty_list_if_null().Any(p => p.Contains("installpackage 1.0.0 Before Modification")).ShouldBeTrue();
```
RichardJFoster: Oops. I guess this should be modified to use contains_message too. Will fix.

src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
RichardJFoster: Integration tests now verify the following in UpgradeScenarios:
1) The modifications script from the initial package is run.
2) The modifications script from the new package is not run.
3) The installation script from the new package is run.

They also verify that in UninstallScenarios:
1) The modifications script from the installed package is run.
2) The uninstall script from the installed package is run.

I believe that the current functionality does match what is desired.

RichardJFoster:
OK. At this point I believe the functionality is present and operational (and verified with tests).

The changes to PowershellService.cs really aren't showing up well in the merge view. I don't think GitHub should be seeing as many modifications as it is. A couple of them appear to be whitespace only (thanks Visual Studio :-/) and I will be trying to make them go away.

The main change to PowershellService.cs was just to allow a "set" of scripts (where each script has an associated flag to indicate if errors in that script can be ignored or should terminate processing) to be run against a single package.

run_package_scripts was exposed to facilitate testing of the failure handling, The only places outside PowershellService.cs which call that method are in the PowershellServiceSpecs, but only the "noop" tests are being run at this time. This is because wrap_script_with_module requires the presence of (at least) chocolateyInstaller.psm1 and chocolateyScriptRunner.ps1 in the folder referenced by ApplicationParameters.InstallLocation.

If you have any thoughts/suggestions for what my next step(s) should be, please let me know.

src/chocolatey/infrastructure.app/services/PackageScriptService.cs
RichardJFoster: Sorry for the slow response. I guess somewhere in a rebase this change got flagged as obsolete and the message disappeared too. In response to your question, the current operation (in NugetService.cs upgrade_run) does indeed seem to convert an upgrade into an install call on the newly updated package (line 516).

RichardJFoster:
@ferventcoder Since you indicated that an "easier" path would be preferred, I went back and tried again, more closely following the already existing pattern. I feel more comfortable that others will understand what is going on. Please let me know if this is better.

ferventcoder:
I haven't forgotten about this. I plan to get back to this soon.

RichardJFoster:
Not a problem. I keep meaning to rebase it again so you don't need to . Perhaps this evening... if I'm still awake enough after getting the kids to bed etc.!

RichardJFoster:
Rebased, as per previous comment. Tests all passed locally. If, for some reason, they don't on one of the CI servers I'll do what I can to sort them out.

ferventcoder:
Merged into stable at 35ec3efb4460f75d3517652bcfee03f5ceb7f70a

src/chocolatey/infrastructure.app/services/PowershellService.cs
```diff
@@ -59,23 +60,31 @@ public PowershellService(IFileSystem fileSystem, CustomString customImports)
...
+                    file = "chocolateyBeforeModify.ps1";
+                    break;
             }
 
             var packageDirectory = packageResult.InstallLocation;
-            var installScript = _fileSystem.get_files(packageDirectory, file, SearchOption.AllDirectories).Where(p => !p.to_lower().contains("\\templates\\"));
-            if (installScript.Count() != 0)
-            {
-                var chocoInstall = installScript.FirstOrDefault();
+            var installScript = _fileSystem.get_files(packageDirectory, file, SearchOption.AllDirectories).FirstOrDefault();
```
ferventcoder: Not sure if you rebased this out or not, but next time you run into rebase issues, let me know - this removed a very important piece of functionality that needed to be here. I fixed it prior to merge.

src/chocolatey/infrastructure.app/services/PowershellService.cs
```diff
@@ -100,15 +109,27 @@ public bool uninstall(ChocolateyConfiguration configuration, PackageResult packa
...
         }
 
         public string wrap_script_with_module(string script, ChocolateyConfiguration config)
         {
-            var installerModule = _fileSystem.combine_paths(get_helpers_folder(), "chocolateyInstaller.psm1");
-            var scriptRunner = _fileSystem.combine_paths(get_helpers_folder(), "chocolateyScriptRunner.ps1");
+            var installerModules = _fileSystem.get_files(ApplicationParameters.InstallLocation, "chocolateyInstaller.psm1", SearchOption.AllDirectories);
+            var installerModule = installerModules.FirstOrDefault();
+            var scriptRunners = _fileSystem.get_files(ApplicationParameters.InstallLocation, "chocolateyScriptRunner.ps1", SearchOption.AllDirectories);
+            var scriptRunner = scriptRunners.FirstOrDefault();
```
ferventcoder: The same here. I had a security fix that was removed on rebase. I fixed it prior to merge.

ferventcoder: No worries otherwise. Rebasing is hard.

src/chocolatey/infrastructure.app/services/PowershellService.cs
ferventcoder: Rebasing is hard

RichardJFoster: Hmmm... that's worrying. The removed code doesn't look at all familiar to me (I don't remember seeing anything related to the templates folder previously, even in any "merge conflict" scenario). Thanks for doing the fixup.

ferventcoder: Could have been it was something that was not yet merged.  No worries :+1:

PR #545, opened 2016-01-15
(GH-551) adding unset to config commands

### LINKED ISSUE #551: Unset Configuration Values
Configuration such as proxy can be activated in command as ```choco config set proxy http://sample```, but there is no way to turn that off.
Due to the restriction of ```config set``` command, you can't use```choco config set proxy ""``` to remove proxy address.
I think it's better to stick with that to avoid unwanted effect and add a ```unset``` command to remove proxy settings
e.g.
```choco config unset proxy```

### PR DESCRIPTION
**Pull request by zhao-xu**
Add ```unset``` to config commands so proxy can be switched on/off in scripts.

For example:
```
choco config set proxy http://localhost:8118
choco install package_requires_proxy_to_install
choco config unset proxy
choco install package_do_not_need_proxy_can_speed_up
```

Closes #551

### DIFF
--- a/src/chocolatey.tests/infrastructure.app/commands/ChocolateyConfigCommandSpecs.cs
+++ b/src/chocolatey.tests/infrastructure.app/commands/ChocolateyConfigCommandSpecs.cs
@@ -131,6 +131,14 @@ namespace chocolatey.tests.infrastructure.app.commands
                 because();
                 configSettingsService.Verify(c => c.config_set(configuration), Times.Once);
             }
+
+            [Fact]
+            public void should_call_service_source_unset_when_command_is_unset()
+            {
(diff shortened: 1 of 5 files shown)


ferventcoder:
This is great as far as the code goes! I thought I had already created an issue for this - but apparently I never got to actually creating the issue. Would you read over https://github.com/chocolatey/choco/blob/master/CONTRIBUTING.md and create the issue and then update your commit message as mentioned in the document? Then we can move forward with this!

zhao-xu:
Sorry, I forgot that. Document's too long, didn't read them all :-p
I went throught the issue list last night and didn't find any related so I created a new one #551 
Commit message has been updated to metion the issue number.

ferventcoder:
> Document's too long, didn't read them all :-p

Better to document explicit expectations rather than leave it up for folks to interpret themselves. :)

PR #546, opened 2016-01-15
Exception if no sources

### LINKED ISSUE #490: Exception if no source is enabled
If I have all my sources disabled and try to install a package a null reference exception is thrown:

```
C:\>choco install nodejs
Installing the following packages:
nodejs
By installing you accept licenses for the packages.
Object reference not set to an instance of an object.
```

Better would be if the exception is handled and instead an error message is displayed informing the user that no source is available.

MikeFoden:
Previously if no sources were available during installation of a
package, installation would fail with an Exception. This fix will make
the choco client check the source list to make sure it is not null
before attempting to install a package.

As per [my discussion ](https://github.com/chocolatey/choco/issues/490#issuecomment-169809268) with @ferventcoder this fixes #490.

The exception itself occurs at ```\src\chocolatey\infrastructure.app\services\NugetService.cs:313``` when checking the source list contains an extension without confirming there is something in the source list first. I've added in a check before showing the licenses to confirm that there is sources defined.


MikeFoden:
Hmm, I seem to have done something strange here - it should only be my commit, I'm sorry @ferventcoder.

MikeFoden:
Superseded by #547.

PR #547, opened 2016-01-15
(GH-490) Exception if no source is enabled

### LINKED ISSUE #490: Exception if no source is enabled

MikeFoden:
Previously if no sources were available during installation of a
package, installation would fail with an Exception. This fix will make
the choco client check the source list to make sure it is not null
before attempting to install a package.

As per [my discussion ](https://github.com/chocolatey/choco/issues/490#issuecomment-169809268) with @ferventcoder this fixes #490.

The exception itself occurs at ```\src\chocolatey\infrastructure.app\services\NugetService.cs:313``` when checking the source list contains an extension without confirming there is something in the source list first. I've added in a check before showing the licenses to confirm that there is sources defined.

This PR supersedes #546.

### DIFF
--- a/src/chocolatey.tests.integration/scenarios/InstallScenarios.cs
+++ b/src/chocolatey.tests.integration/scenarios/InstallScenarios.cs
@@ -3333,5 +3333,29 @@ namespace chocolatey.tests.integration.scenarios
                 _xPathNavigator.SelectSingleNode("//configuration/appSettings/add[@key='insert']/@value").TypedValue.to_string().ShouldEqual("1.0.0");
             }
         }
+
+        [Concern(typeof(ChocolateyInstallCommand))]
+        public class when_installing_a_package_with_no_sources_enabled : ScenariosBase
+        {
+
+            public override void Context()
+            {
+                base.Context();
+                Configuration.Sources = null;
+            }
+
+            public override void Because()
+            {
+                MockLogger.reset();
+                Results = Service.install_run(Configuration);
+            }
+
+            [Fact]
+            public void should_have_no_sources_enabled_result()
+            {
+                MockLogger.contains_message("Installation was NOT successful. There are no sources enabled for packages.", LogLevel.Error).ShouldBeTrue();
+                Results.Count().ShouldEqual(0);
+            }
+        }
     }
 }
--- a/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
+++ b/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
@@ -304,9 +304,17 @@ namespace chocolatey.infrastructure.app.services
         {
             this.Log().Info(@"Installing the following packages:");
             this.Log().Info(ChocolateyLoggers.Important, @"{0}".format_with(config.PackageNames));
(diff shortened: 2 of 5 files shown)


src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
```diff
@@ -304,10 +304,18 @@ public void handle_package_result(PackageResult packageResult, ChocolateyConfigu
...
 
+            if (config.Sources == null)
+            {
+                this.Log().Error(ChocolateyLoggers.Important, @"Installation was NOT successful. There are no sources enabled for packages.");
+                Environment.ExitCode = 1;
+                return packageInstalls;
+            }
+
+            this.Log().Info(@"By installing you accept licenses for the packages.");
+
```
ferventcoder: I believe this is also necessary for upgrade_run.

ferventcoder:
Please also add a few tests that will show this working properly. That can be done in the install/upgrade scenarios if you want.

ferventcoder:
Overall :+1: so far!

MikeFoden:
@ferventcoder I have added it in for upgrade and for list, and set up tests for all three to show that if there is no sources available that it will show a message to the user explaining the issue.

src/chocolatey.tests.integration/scenarios/InstallScenarios.cs
```diff
@@ -3333,5 +3333,29 @@ public void should_add_the_insert_value_in_the_config_due_to_XDT_InsertIfMissing
...
+                Results = Service.install_run(Configuration);
+            }
+
+            [Fact]
+            public void should_have_no_sources_enabled_result()
+            {
+                MockLogger.contains_message("Installation was NOT successful. There are no sources enabled for packages.", LogLevel.Error).ShouldBeTrue();
+                Results.Count().ShouldEqual(0);
+            }
+        }
```
ferventcoder: :+1:

ferventcoder:
Do we have the ability to fix the exception at the source?

MikeFoden:
It's anything that references `config.Sources` causing the exception. The messages I have added in I guess would be considered friendly error messages.

There is a couple of things I can do:

* We could patch each usage of `config.Sources` to confirm there is at least one source defined first.
* Modify it so when `config.Sources` is loaded from the configuration file, if there is no valid sources have it get stored as an empty string, rather than null.

Which one works best for you @ferventcoder? I would assume the scenario that no sources are available would be relatively small as by default the chocolately feed is configured - a user would have to actively play with the configuration to get this situation to occur.

ferventcoder:
All good - I'll take a look at this and merge it soon.

ferventcoder:
Thanks for the contribution!

MikeFoden:
Thanks for helping me get involved!

ferventcoder:
I fixed up a little and squashed the commits. Thanks again!

PR #740, opened 2016-05-18
(GH-739) Fix: Update templates to use extension

dtgm:
Closes #739

Update templates

PR #743, opened 2016-05-22
(GH-739) Fix templates to use helper in extension to retrieve registr…

### LINKED ISSUE #739: Uninstall template fails when encountering "bad" registry keys
https://github.com/chocolatey/choco/blob/7ef16d6003dd3c689156af20d3ce329c85d1cd76/src/chocolatey/infrastructure.app/templates/ChocolateyUninstallTemplate.cs#L61-L63
```
[array]$key = Get-ItemProperty -Path @($machine_key6432,$machine_key, $local_key) `
                        -ErrorAction SilentlyContinue `
         | ? { $_.DisplayName -like ""$softwareName"" }
```
This code will fail if a registry key exists that was written with binary data and incorrectly typed. `[array]$key` does not resolve issue as Get-ItemProperty will fail before set to variable.

For more details, see https://github.com/dtgm/chocolatey-packages/issues/92#issuecomment-174078806

dtgm:
Closes #739

Repush of https://github.com/chocolatey/choco/pull/740 (same changes)

### DIFF
--- a/src/chocolatey/infrastructure.app/templates/ChocolateyUninstallTemplate.cs
+++ b/src/chocolatey/infrastructure.app/templates/ChocolateyUninstallTemplate.cs
@@ -54,13 +54,8 @@ if ($installerType -ne 'MSI') {
 }
 
 $uninstalled = $false
-$local_key     = 'HKCU:\Software\Microsoft\Windows\CurrentVersion\Uninstall\*'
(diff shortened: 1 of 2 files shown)


ferventcoder:
This is close - what I originally meant here is that the helper gets included in Chocolatey. Then add the helper as a dependency for targeting older choco versions.

ferventcoder:
Fixed up and merged into stable at https://github.com/chocolatey/choco/commit/91db949a7f100c5dac49207e28a48cc7f6dab2c4

Issue #557, opened 2016-01-19
Placement of the "You have {x} v{y} installed. Version {z} is available based on your source(s)' is wrong.


cpmcgrath:
When you upgrade multiple items, the transition from one to the next looks like this...
![chocoissue](https://cloud.githubusercontent.com/assets/1791243/12406241/ff06f388-be96-11e5-8f80-271d2996a244.png)

This associates the yellow line with the first block of data. It of course has nothing to do with that block and everything to do with the next block.

A new line should be added before the yellow line, and preferably the new line after it removed.

ferventcoder:
Thanks for logging this!

PR #787, opened 2016-06-11
(GH-557) Placement of the "You have {x} v{y} installed. Version {z} is available based on your source(s)' is wrong.

eeevans:
Closes 557

Added newline preceding the newer version available message.  Did not attempt to remove newline after message as it is part of the package name and version message when starting the install.

This implementation add the newline whether part of upgrading multiple items or single item but I could add that as I think that context is available..

### DIFF
--- a/src/chocolatey/infrastructure.app/services/NugetService.cs
+++ b/src/chocolatey/infrastructure.app/services/NugetService.cs
@@ -647,7 +647,7 @@ spam/junk folder.");
 
                 if (installedPackage.Version == availablePackage.Version)
                 {
-                    string logMessage = "{0} v{1} is the latest version available based on your source(s).".format_with(installedPackage.Id, installedPackage.Version);
+                    string logMessage = "{0}{1} v{2} is the latest version available based on your source(s).".format_with(Environment.NewLine, installedPackage.Id, installedPackage.Version);
 
                     if (!config.Force)
                     {


ferventcoder:
I appreciate this PR, however I fixed this just yesterday for other reasons.

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -647,7 +647,7 @@ public void remove_rollback_directory_if_exists(string packageName)
 
                 if (installedPackage.Version == availablePackage.Version)
                 {
-                    string logMessage = "{0} v{1} is the latest version available based on your source(s).".format_with(installedPackage.Id, installedPackage.Version);
+                    string logMessage = "{0}{1} v{2} is the latest version available based on your source(s).".format_with(Environment.NewLine, installedPackage.Id, installedPackage.Version);
```
ferventcoder: This placement was wrong - it would have ended up in the packageresult messages with the extra line as well.

eeevans:
That's cool, I'll try and figure out where your fix is as a learning exercise.

ferventcoder:
I actually didn't do the commits for the work until this morning so the dates will be todya.

ferventcoder:
Here's the commit for review - https://github.com/chocolatey/choco/commit/7875987ea318460e1605564a6c06b277d7b3cd3b

PR #831, opened 2016-06-24
(GH-828) Return OS bit width, not process

### LINKED ISSUE #828: Get-OSArchitectureWidth doesn't do what it says it does
Ref: #713 Rename Get-ProcessorBits as a more appropriately named Get-OSArchitectureWidth

`Get-OSArchitectureWidth` returns the current *process* bit-width regardless of OS arch. So in 32-bit powershell running on 64-bit windows, it returns `32` because `[System.IntPtr]::Size` returns `4` in 32-bit process regardless of OS architecture (32 or 64)

The helper should be replaced by testing whether process is running on 64-bit OS if the intent is to return what the helper says it will.

So if `[System.IntPtr]::Size` returns `4`, also do `test-path env:\PROCESSOR_ARCHITEW6432`

```
if ([IntPtr]::size -eq 8) {
  $bits = 64
} elseif (([IntPtr]::size -eq 4) -and (test-path env:\PROCESSOR_ARCHITEW6432)) {
  $bits = 64
} else {
  $bits = 32
}
```

dtgm:
(GH-828) Return OS bit width, not process

Get-OSArchitectureWidth returns the current process bit-width regardless of
OS arch. So in 32-bit powershell running on 64-bit windows, it returns 32
because [System.IntPtr]::Size returns 4 in 32-bit process regardless of
OS architecture (32 or 64). This fix returns actual OS bit width.

Closes #828

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Get-OSArchitectureWidth.ps1
+++ b/src/chocolatey.resources/helpers/functions/Get-OSArchitectureWidth.ps1
@@ -45,7 +45,9 @@ param(
   Write-Debug "Running 'Get-OSArchitectureWidth'"
 
   $bits = 64
-  if ([System.IntPtr]::Size -eq 4) {
+  if (([System.IntPtr]::Size -eq 4) -and (Test-Path env:\PROCESSOR_ARCHITEW6432)) {
+    $bits = 64
+  } elseif ([System.IntPtr]::Size -eq 4) {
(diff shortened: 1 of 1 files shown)


ferventcoder:
Merged into stable at https://github.com/chocolatey/choco/commit/45ccd77a0b73a0d0c92f0c3ffbeed0172d4f4f6b

PR #898, opened 2016-08-09
(GH-897) dism ... /all was never running for source windowsfeatures

### LINKED ISSUE #897: dism /all is never running -- no dependencies get installed.
Running choco on windows 10.0.x with source windowsfeatures, there's a problem with installing dependencies to a particular feature. 

Looks like the issue is in /src/chocolatey/infrastructure.app/services/WindowsFeatureService.cs Line 230, where the condition to add the all argument requires the minor version of windows to always be greater than 2, no matter the major version. 

Will add a pull request with a suggested fix to this issue.

jrkd:
(maint)

Issue here is the minor version is always required to be > 2, so now that we're at windows 10.0.x /all is not being added, even though this logic really only wanted to make sure the version was >= 6.2

This commit adds a check to see if the version is greater than 6 entirely, if not, only then does it check if its version 6 and at least minor ver. 6.2.

How this issue affects the functionality: Not calling /all as part of dism meant that dependencies for windows features are never installed as part of the installation.

Closes #897

### DIFF
--- a/src/chocolatey/infrastructure.app/services/WindowsFeatureService.cs
+++ b/src/chocolatey/infrastructure.app/services/WindowsFeatureService.cs
@@ -227,7 +227,7 @@ namespace chocolatey.infrastructure.app.services
             var args = ExternalCommandArgsBuilder.build_arguments(config, argsDictionary);
 
             // at least Windows 8/2012
-            if (config.Information.PlatformVersion.Major >= 6 && config.Information.PlatformVersion.Minor >= 2)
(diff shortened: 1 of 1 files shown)


jrkd:
Have now signed the CLA.

ferventcoder:
@jrkd I see that. CLAHub didn't catch it though. You are good. Thanks!

By the way `gism` == `dism`.

ferventcoder:
Fixed up and merged into stable at https://github.com/chocolatey/choco/commit/e428c85d0537cfb618c534eacf1aaaca174a010b

PR #923, opened 2016-08-23
clarify word choice

cori:
seems like the intent is to `wrap` at `72`characters, not to `indent`.


cori:
the changes in the diffs do not match what shows up in the file preview. This should be just a 1 word change that I made in the browser editor.

ferventcoder:
> the changes in the diffs do not match what shows up in the file preview. This should be just a 1 word change that I made in the browser editor.

I think it may have to do with LineEndings - I've watched SublimeText add more lines than were there.

Issue #686, opened 2016-04-11
When a prerelease is installed, it should upgrade to latest prerelease unless excluded - typically seen in choco upgrade all


ScottRFrost:
If you do "choco upgrade all -y" and it encounters a package that's newer than newest, perhaps it should check prerelease or at least offer to do so.

Current: 
`chocolatey v0.9.10-beta-20160402 is newer than the most recent.`

Proposed:
`chocolatey v0.9.10-beta-20160402 is newer than the most recent. Would you like to check prereleases for chocolatey? [y] [n]`

ferventcoder:
In the case of automated unattended scripts, how do you see it working?

ScottRFrost:
Well, it would be dangerous to ever run "choco upgrade all -y -pre" because then it would upgrade ALL of your stuff to prerelease versions.  Perhaps you could add a flag "-includepre" or something that upgrades packages that are already prerelease to a newer prerelease, but leaves stable packages alone (and does not upgrade them to prerelease versions).

This flag would allow unintended scripts to function without blocking waiting for y/n while still upgrading your prerlease packages to the latest bits.

ferventcoder:
@ScottRFrost that makes way too much sense :+1:

eeevans:
I'd like to work on this feature and tbh I've already got a lot of the code in place other than the prompting if run without the -includepre switch...I was just wondering if you really want to do that or just spit out info messages informing of the switch to allow it.

ferventcoder:
@eeevans let's see what you have.

ferventcoder:
So this feature will go into 0.10.4 like this - if you have prereleases installed it will upgrade them without needing to explicitly specify prerelease. This will allow everything to upgrade automatically. When it is a prerelease and the latest version is pre or stable, that is what it will upgrade to.

When you want the old behavior, just add `--exclude-prerelease`. I've tagged this with breaking change as it may have an effect on someone's workflow out there.

ferventcoder:
This will be in 0.10.4

eeevans:
Awesome, thanks!

FranklinYu:
If we upgrade with `--exclude-prerelease`, will the flag of remembered `--pre` be removed (forgotten)? For example, I upgrade from 1.0 to 1.1-rc1 with `--pre`, then upgrade to `1.1` with `--exclude-prerelease`, will it still upgrade to 1.2-rc1 when I `cup all`?

bc3tech:
[...]

PR #949, opened 2016-09-06
Gh686

eeevans:
(GH-686) - Add --includepre option to Upgrade command to allow a package which already has a prerelease version installed to be upgraded to the latest prerelease version.


eeevans:
My attempt at squashing didn't work

ferventcoder:
ouch on squashing.

ferventcoder:
The right thing seems to be to upgrade UNLESS opted out. Whether it is a prerelease or not is not important to an end user. Thoughts?

so `choco upgrade all --exclude-installed-prereleases`

eeevans:
Yes, that seems a reasonable expectation.  I was just wary of changing current functionality without an explicit warning but I can see how it is a more desired functionality.

ferventcoder:
@eeevans any updates on flipping this?

ferventcoder:
@eeevans Thanks for your contributions, unfortunately we went to the more desired functionality of upgrading packages that are prereleases as well as folks have installed the prerelease, so they want it to upgrade to the latest version (stable or prerelease). The old behavior can be brought back with `--exclude-prerelease` in 0.10.4+.

Issue #1015, opened 2016-10-13
Possible bug in Install-ChocolateyPath.ps1


DarwinJS:
This line checks the existing path without qualifying it for "user" or "machine"
https://github.com/darwinjs/choco/blob/master/src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1#L78

However, the path addition statements later in the code are structured so as to add the path only to the requested scope.

If a packager calls the function with the intent of adding to "machine", shouldn't the code add it to machine EVEN IF it already exists in user?

Paths in machine are [a] available when no one is logged in (e.g. group policy processing), [b] are available to system processes and services (that do not login as the current user) and [c] are available to all users - so it differs significantly in functionality. 

It seems if line 78 also qualified the check by user or machine it would seem to properly handle situations where the path exists in user, but the intent is to make the software available to all users or the system and I _think_ it should also accurately handle situations where the intent is to only update the user's path for them to use the software.

ferventcoder:
Agreed. Hacktoberfest it is.

DarwinJS:
Ok - I'll do this one too as well.

ferventcoder:
IIRC, this entire function could use some modernization (or best practices for PowerShell added in). Splitting all values into an array at `;` would be a good start.

DarwinJS:
OK cool - that's exactly the code I had planned for uninstall, but wasn't sure how much to stick to the existing patterns.

DarwinJS:
Instead of https://github.com/darwinjs/choco/blob/master/src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1#L102-L104,

Is it ok just to call Update-SessionEnvironment like earlier in the same code? : https://github.com/darwinjs/choco/blob/master/src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1#L76

DarwinJS:
Adding some more problem data here - the function does not work when the path being requested is a parent of a path that is already on the path (stumbled upon when using 'c:\users' as a test case).

I believe this would be fixed by parsing the path into an array and comparing our path to each array item for an exact, case insensitive match.

However, this will be more changes than the if statement.

DarwinJS:
PR https://github.com/chocolatey/choco/pull/1017

brunoyb:
[...]

PR #1017, opened 2016-10-15
(#1015) Path exists check retains path scope specified by user, adds semi-colon to end of path

DarwinJS:
(GH 1015) Path exists check retains path scope specified by user,  adds semi-colon to end of path

Also path exists check finds path whether or not trailing slash matches requested path, added alias 'path' for 'pathtoinstall' to align with powershell parameter naming, used chocolatey method Update-SessionEnvironment instead of simple path concetanation.

#1015

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1
+++ b/src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1
@@ -17,7 +17,7 @@ function Install-ChocolateyPath {
 .SYNOPSIS
 **NOTE:** Administrative Access Required when `-PathType 'Machine'.`
 
-This puts a directory to the PATH environment variable.
+This puts a directory to the PATH environment variable of the requested scope (Machine or User).
 
 .DESCRIPTION
 Looks at both PATH environment variables to ensure a path variable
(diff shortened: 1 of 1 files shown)


DarwinJS:
Looking for feedback on the following possible additional changes:

- Thinking I should also change "write-host" to best practice "write-output" - however, I'm not sure if that has implications due to the fact chocolatey hosts powershell by default?
- New scope option 'All' for adding path to both User and Machine
- Check if path is in other scope and log that fact
- Give a warning that paths added to User scope will ONLY be available to that user and not the System, services nor other users.

src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1
```diff
@@ -65,28 +65,23 @@ Set-EnvironmentVariable
...
   $originalPathToInstall = $pathToInstall
-
-  #get the PATH variable
-  Update-SessionEnvironment
-  $envPath = $env:PATH
-  if (!$envPath.ToLower().Contains($pathToInstall.ToLower()))
+  #array drops blanks (one of which is always created by final semi-colon)
+  $actualPathArray = (Get-EnvironmentVariable -Name 'Path' -Scope $pathType -PreserveVariables).split(';',[System.StringSplitOptions]::RemoveEmptyEntries)
+  #checks for match with and without trailing slash
+  if (($actualpathArray -inotcontains $pathToInstall.ToLower()) -AND ($actualpathArray -inotcontains "$(($pathToInstall + '\').ToLower())"))
```
Litee: $actualpathArray is compared with lower case version of $pathToInstall. Where $actualpathArray is converted to lower case?

DarwinJS: The "i" in "-inotcontains" causes the entire comparison operation to be "case **I**nsensitive".  Technically this make "ToLower()" redundant.

dragon788:
@DarwinJS I think from talking with @ferventcoder [about this on Gitter](https://gitter.im/chocolatey/choco?at=581bb24131c5cbef43b21447) Write-Host is best because it avoids major pitfalls of Write-Output's behavior, if you have Write-Output inside a function it automatically forces itself into the return value. There is a new Write-Information, but that is a PS5 only function iirc.

DarwinJS:
Trying to understand why this isn't getting merged?

ferventcoder:
Apologies @DarwinJS - this change wasn't under vNext, so it didn't get in for 0.10.4 - 0.10.5 was a bug fix release for a gating issue in 0.10.4. 0.10.6 is fixing an issue that came in from a change to how xml files are saved. 

It will make it in, and I did take a look and notice it may change the current behavior of looking at the entire path (breaking change), even if that is technically the right thing to do, and realized this changes a lot of code so it requires more scrutiny. Since it wasn't technically scheduled for the next release, it did not get merged. If we can agree that this is more of an annoyance than a critical fix - it may add an item to path twice, but it won't necessarily stop choco from working. HTH

DarwinJS:
@ferventcoder - no need to apologize - I am just trying to understand.
Is the same true of https://github.com/chocolatey/choco/pull/1019 ?

jedfrechette:
Has there been any progress on getting this merged?

I just ran in to the issue described in #1015, where the Machine PATH was silently not getting updated because the path I was trying to add already existed on the User PATH.  I never would have guessed what the problem was without finding that bug.

vexx32:
Thanks for the contribution here, and apologies that this was not reviewed for a long while.

This has been resolved during the rewrite of this command to C# which shipped in 2.3.0 and the issue this seeks to resolve is no longer present; see here: https://github.com/chocolatey/choco/issues/1015#issuecomment-2359180577

As such, I'll be closing this PR, but thank you for the contribution nonetheless! 💜

PR #1019, opened 2016-10-15
Issue310

DarwinJS:
I would like feedback on whether to incorporate any of these features:
- warn that the path also exists in the opposite scope (User or Machine) - but don't process the removal from the non-requested scope
- allow a special scope option 'All' which removes the path from both scopes (User and Machine) in one pass


src/chocolatey.resources/helpers/functions/Uninstall-ChocolateyPath.ps1
```diff
@@ -0,0 +1,117 @@
+# Copyright 2011 - Present RealDimensions Software, LLC & original authors/contributors from https://github.com/chocolatey/chocolatey
```
ferventcoder: Not that you would have known this, but this won't carry the second part of the copyright as this didn't come from the old code base.

src/chocolatey.resources/helpers/functions/Uninstall-ChocolateyPath.ps1
```diff
@@ -0,0 +1,117 @@
...
+param(
+  [parameter(Mandatory=$true, Position=0)][alias("Path")][string] $PathToUninstall,
+  [parameter(Mandatory=$false, Position=1)][ValidateSet('User','Machine')][System.EnvironmentVariableTarget] $pathType = [System.EnvironmentVariableTarget]::User,
+  [parameter(ValueFromRemainingArguments = $true)][Object[]] $ignoredArguments
+)
+  Write-Debug "Running 'Uninstall-ChocolateyPath' with PathToUninstall:`'$PathToUninstall`'";
+  Write-Output "Only evaluating and updating path scope `"$pathType`", path will not be assessed nor removed for other scope, so path may exist in other scope as well."
+  $originalPathToUninstall = $PathToUninstall
+  #First half on handling trailing slash properly - remove it from requested path:
+  $PathToUninstall = $PathToUninstall.trimend('\')
```
ferventcoder: What is trimend and is it available in PowerShell v2+ (and thus .NET Framework 2.0+)?

src/chocolatey.resources/helpers/functions/Uninstall-ChocolateyPath.ps1
```diff
@@ -0,0 +1,117 @@
...
+.LINK
+Get-ToolsLocation
+#>
+param(
+  [parameter(Mandatory=$true, Position=0)][alias("Path")][string] $PathToUninstall,
+  [parameter(Mandatory=$false, Position=1)][ValidateSet('User','Machine')][System.EnvironmentVariableTarget] $pathType = [System.EnvironmentVariableTarget]::User,
+  [parameter(ValueFromRemainingArguments = $true)][Object[]] $ignoredArguments
+)
+  Write-Debug "Running 'Uninstall-ChocolateyPath' with PathToUninstall:`'$PathToUninstall`'";
+  Write-Output "Only evaluating and updating path scope `"$pathType`", path will not be assessed nor removed for other scope, so path may exist in other scope as well."
```
ferventcoder: I know it is widely believed that `Write-Host` is wrong, and it is under most circumstances. When it comes to logging, we prefer `Write-Host` in the Chocolatey installer module. It doesn't cause issues when you want to return actual things like `Write-Output` does. So please convert these to `Write-Host`.

Yes, I am familiar with http://www.jsnover.com/blog/2013/12/07/write-host-considered-harmful/:

* The post is about scripts that will be used by other tools.
* `Write-Output` ends up in the pipeline (aka is part of the return value from the function). This can have (and has had) disastrous results, so we prefer `Write-Host`. I prefer real life examples to rhetoric, so https://github.com/chocolatey/choco/commit/44306efbe55b982d92c9c23ba3b44df01e6f242e
* We control the hosts this runs under.
* We simply want logging messages.
* This is not a library or framework, but an end tool, which doesn't appear to be considered in the post.

Consider `Write-Output` == `Log.Info()` and `Write-Host` == `Log.Info()` for the Chocolatey framework.
Yes, I realize this means we should have probably set up a logging utility. We could go on all day long about how we should have brought in an actual logging facility like log4net, but honestly at the end of the day it is added work to provide what `Write-Host` already provides.

ferventcoder:
> allow a special scope option 'All' which removes the path from both scopes (User and Machine) in one pass

I like it!

ferventcoder:
> warn that the path also exists in the opposite scope (User or Machine) - but don't process the removal from the non-requested scope

Also like it. And you could also provide that there is an "ALL" option as part of the warning.

DarwinJS:
OK - the combination of: [1] Always logging which scopes the path appears in (even if not the one being changed) and [2] Supporting the recursion of the function to switch to admin rights if possible and [3] allow use of -pathToUninstall 'All' resulted in a very different code structure.

src/chocolatey.resources/helpers/functions/Uninstall-ChocolateyPath.ps1
```diff
@@ -0,0 +1,174 @@
...
+.DESCRIPTION
+Removes path from target path scope.  Removes multiple occurances (if they exist)
+and all occurances with or without a trailing slash.
+
+.NOTES
+This command will assert UAC/Admin privileges on the machine if
+`-PathType 'Machine'`.
+
+This is used when the application/tool is not being linked by Chocolatey
+(not in the lib folder).
```
ferventcoder: I will also need to update this to indicate first supported version of Chocolatey this is in.

DarwinJS:
Ok - choco gods reference removed and corrected error in first person.

I am having a hard time discovering whether .trimend is available in PSH 2 via research.

I have temporarily lost access to my Windows 7 VMs because they were in Hyper-V and after Windows 10 Anniversary update I cannot have both Virtual Box AND Hyper-V INSTALLED at the same time.  Before Anniversary Update both could be installed, but not active at once.  I don't have any VMs with PSH 2 on Virtual Box yet and it would be a bit of fusing to reinstall hyper-V and then deinstall when done.

DarwinJS:
I was able to test to be sure that ".trimend" is available as a string method on PSH 2.0

jberezanski:
FWIW, [System.String.TrimEnd(char[])](https://msdn.microsoft.com/en-us/library/system.string.trimend(v=vs.71).aspx) is available at least since .NET 1.1 (that's as far back as MSDN docs go).

DarwinJS:
@jberezanski - thanks for the research on that!

dragon788:
Looking forward to this, was just taking a whack at implementing this within an individual package similar to one on the linked issues that DarwinJS had mentioned. Nothing drives me nuts more than clutter in the PATH that leads to aberrant behavior.

dragon788:
Since I needed to remove multiple entries from the PATH and I didn't want to loop over the list multiple times, I improved a little bit on your workaround on [ec2clitools](https://github.com/DarwinJS/ChocoPackages/blob/master/ec2clitools/tools/chocolateyuninstall.ps1).

@DarwinJS perhaps a similar "optimization" could be useful here? Since multiple foreach loops is rough, I opted to use the regex array comparison outlined http://masteringposh.com/powershell-script-multiple-array-comparisons-using-regular-expressions/ and ended up with the code below.

```
[regex] $PathsToRemove = "^(%NVM_HOME%|%NVM_SYMLINK%)"
$environmentPath = [Environment]::GetEnvironmentVariable("PATH","Machine")
[string[]]$newpath = ''
foreach ($path in $environmentPath.split(';'))
{
  If (($path) -and ($path -notmatch $PathsToRemove))
    {
        [string[]]$newpath += "$path"
        "$path added to `$newpath"
    } else {
        "Bad path found: $path"
    }
}
$AssembledNewPath = ($newpath -join(';')).trimend(';')

[Environment]::SetEnvironmentVariable("PATH",$AssembledNewPath,"Machine")
```

DarwinJS:
@dragon788 - 

- I believe multiple path removal would be an edge case.  
- Not sure what would be rough about it as it's likely milliseconds of time among a much longer wait - you could just call the function twice.
- the above does not seem to handle the trailing slash?

If multiple paths is deemed as a requirement, I would probably just make the function capable of receiving pipelined input then you could do this: 
``` @('%NVM_HOME%','%NVM_SYMLINK%') | Uninstall-ChocolateyPath```

Here is a brief article on why taking the pipeline approach can be very powerful: http://csi-windows.com/blog/all/74-powershell/438-paradigm-shift-the-power-of-functions-that-take-pipelined-input

I definitely wouldn't want have end consumers of the function have to remember that they are using a regex when it seems they won't need that most of the time.

@ferventcoder - what do you think about multiple path support with a pipeline capable function?

**Comment by dragon788** [comment]
That's pretty awesome, and I definitely agree that multiple values in the path are unlikely. Hopefully most packages are following the single responsibility principle and keeping things grouped together, I just tend to hit corner cases (Nvm for Windows that manages node versions via symlinks in Program Files but also needs awareness of itself and it's location).

**Comment by ferventcoder** [comment]
> multiple path support with a pipeline capable function?

Could be interesting. For first pass implementation, it doesn't cost much to loop over path information multiple times. So perhaps a new enhancement issue for this one?

File: src/chocolatey.resources/helpers/functions/Uninstall-ChocolateyPath.ps1 [hunk 5]
```diff
@@ -0,0 +1,174 @@
...
+  [parameter(Mandatory=$false)][switch] $RecursiveCall,
+  [parameter(ValueFromRemainingArguments = $true)][Object[]] $ignoredArguments
+)
+  Write-Debug "Running 'Uninstall-ChocolateyPath' with PathToUninstall:`'$PathToUninstall`'";
+  If (!$RecursiveCall -AND ($pathType -ine 'All')) {Write-Host "Only evaluating and updating path scope `"$pathType`", path will not be assessed nor removed for other scope, so path may exist in other scope as well."}
+  $originalPathToUninstall = $PathToUninstall
+  #First half on handling trailing slash properly - remove it from requested path:
+  $PathToUninstall = $PathToUninstall.trimend('\')
+  #array drops blanks (one of which is always created by final semi-colon)
+  $actualPathArrayUser = (Get-EnvironmentVariable -Name 'Path' -Scope 'user' -PreserveVariables).split(';',[System.StringSplitOptions]::RemoveEmptyEntries)
```
jnm2: We're going to have to do something a little smarter since apparently [semicolons can be escaped using quotes](https://superuser.com/a/594177). Too bad there isn't a Windows API for this.

src/chocolatey.resources/helpers/functions/Uninstall-ChocolateyPath.ps1
```diff
@@ -0,0 +1,174 @@
...
+  $actualPathArrayMachine = (Get-EnvironmentVariable -Name 'Path' -Scope 'machine' -PreserveVariables).split(';',[System.StringSplitOptions]::RemoveEmptyEntries)
```
jnm2: We're going to have to do something a little smarter since apparently [semicolons can be escaped using quotes](https://superuser.com/a/594177). Too bad there isn't a Windows API for this.

Reassembling the path will have to follow the same procedure in reverse.

jnm2: Confirmed using Windows' environment variable list editor: a `tl;dr` entry causes `;"tl;dr";` to be added to the path.

jnm2: Also, the path needs to be recognized and removed even if it is quoted in the PATH variable. I assume the input path would never be quoted.

ferventcoder: So luckily PATH doesn't have `"` and it doesn't have escaped `;` - if you make it have `"`, it doesn't like you anymore IIRC.

ferventcoder: Am I understanding that a semi-colon is allowed in a folder name???!!!

ferventcoder: I just died a little inside. 👎

jnm2:
@DarwinJS @ferventcoder I would really like to see this in Chocolatey. Can I do anything to help? I might start on a proof of concept for both Install- and Uninstall-ChocolateyPath which properly understands quoted paths when looking for existing paths or inserting a new path just to unblock my own package. Would be happier getting them in Chocolatey sooner though!

DarwinJS:
@jnm2 - feel free to take this over.  Handling quotes seems like a very rare case - maybe consider getting this merged and then do that as a new PR?

jnm2:
@DarwinJS Thank you.

[This was the direction](https://gist.github.com/jnm2/9ae8c9e596e7867fb0ff6da14234826a) I was hoping we might go towards. It tries to make the uninstall work as symmetrically as possible with the install and leave a minimal footprint.

Can we do this? What would the next step be? /cc @ferventcoder

jnm2:
Here's the branch with the Un/install-ChocolateyPath pair which respect semicolons: https://github.com/chocolatey/choco/compare/master...jnm2:env_path_uninstall
I put in comments explaining why it is done that way. What do you guys think of this approach?

jnm2:
Hey @gep13, Rob said I should tag you for the two comments above. I'd like to help out and I have a proof of concept that deals with all the edge cases I know of, but I'm not sure if I should start a new PR?

franklinyu:
Any news? Shall we start reviewing @jnm2’s branch?

jnm2:
@franklinyu I haven't heard anything from anyone about this, so I went ahead: https://github.com/chocolatey/choco/pull/1663

ferventcoder:
This may be superseded with @jnm2 's changes.

vexx32:
Apologies for the long wait on this PR. This is being added as part of a larger scope of work involving some rewrites that will be submitted soon, so for now we'll close this PR.

Thank you very much for the contribution! 💜

PR #1026, opened 2016-10-23
(#386) Updated Get-ChocolateyUnzip to handle unzipping embedded .tar archives

### LINKED ISSUE #386: Change Install-ChocolateyZipPackage to fully extract .tar.xz archives
It was [recently suggested](https://chocolatey.org/packages/ghc/7.10.2#comment-2168048807) that `Install-ChocolateyZipPackage` only produces a `.tar` file when called on a `.tar.xz`, and that [chocolatey#473](https://github.com/chocolatey/chocolatey/issues/473) from the old repository was intended to correct this. This should be implemented.

DarwinJS:
Handles all types of compound achives that may contain an embedded .tar - such as .tar.gz, .tgz, .tar.bz2, etc.

Recursively calls Get-ChocolateyUnzip if it detects "originalzipname.tar" in the unzip location.

Closes #386

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
+++ b/src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
@@ -77,6 +77,7 @@ param(
   [parameter(Mandatory=$true, Position=1)][string] $destination,
   [parameter(Mandatory=$false, Position=2)][string] $specificFolder,
   [parameter(Mandatory=$false, Position=3)][string] $packageName,
+  [parameter(Mandatory=$false, Position=4)][switch] $recursivecall,
   [parameter(ValueFromRemainingArguments = $true)][Object[]] $ignoredArguments
 )
   $zipfileFullPath=$fileFullPath
@@ -196,6 +197,31 @@ param(
     default { throw "7-Zip signalled an unknown error (code $exitCode)" }
   }
 
-  $env:ChocolateyPackageInstallLocation = $destination
-  return $destination
+  if ((test-path variable:url) -AND (@($url).count -gt 0))
+  {
+    # If our unzip resulted in a .tar file that matches our input zip file without it's extensions, unzip it
+    # does not work with archive names that contain a '.' before the extensions
+    $matchingtarname = "$destinationNoRedirection\$((split-path -leaf $url).split('.')[0]).tar"
+    Write-Debug "Checking to see if we have a flie called $matchingtarname"
+    If ((Test-path $matchingtarname) -AND (!$RecursiveCall))
+    {
+      Write-Debug "Automatically detected this archive contained a .tar matching the original archive name, unzipping `"$matchingtarname`" ..."
+      Get-ChocolateyUnzip -fileFullPath "$matchingtarname" -destination "$destination" -specificFolder "$specificFolder" -packagename "$packagename" -recursivecall
+      remove-item $matchingtarname -force
+    }
+    else
+    {
+      Write-Debug "The archive file processed did not contain a tar archive."
+    }
+  }
+  else
+  {
+    Write-Debug "This package does did not process an archive file."
+  }
+
+  If (!$RecursiveCall)
+    {
+      $env:ChocolateyPackageInstallLocation = $destination
+      return $destination
+    }
 }


DarwinJS:
I had this code fail a test recently - it was acting like $url was not available in that scenario.

DarwinJS:
OK, I think I have this working - but I'm not sure if it covers all use cases.

The code now checks for the existence of $url and whether it has anything in it.

But I'm still not sure how certain of an indicator $url to be able to find the original archive name - will it be there everytime?

I have also setup debug messaging to try to help debug if this code misses valid cases or is invoked when it should not be.

Successful test cases:
- package that does not contain an archive (should not ever call Get-chocolateyunzip.ps1
- package that does contain an archive AND an embedded .tar
- package that gets an archive from the web, BUT DOES NOT contain an embedded .tar

Untested:
- package that references an archive on a web URL (does anyone know of use case)

espoelstra:
@DarwinJS GitHub's Hub client is probably a good case for the untested scenario. https://chocolatey.org/packages/github-hub

You could also use the server-jre8 Java urls as a test. Though they require a magic cookie and I haven't had any luck getting the Get-ChocolateyWebFile to properly pass the cookie all the way through to the download. I ended up just using System.Net.WebClient directly.

src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
```diff
@@ -77,6 +77,7 @@ param(
   [parameter(Mandatory=$true, Position=1)][string] $destination,
   [parameter(Mandatory=$false, Position=2)][string] $specificFolder,
   [parameter(Mandatory=$false, Position=3)][string] $packageName,
+  [parameter(Mandatory=$false, Position=4)][switch] $recursivecall,
```
ferventcoder: No positional parameters please. We are working to not have anymore added. I can make this adjustment

src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
```diff
@@ -196,6 +197,31 @@ param(
     default { throw "7-Zip signalled an unknown error (code $exitCode)" }
   }
 
-  $env:ChocolateyPackageInstallLocation = $destination
-  return $destination
+  if ((test-path variable:url) -AND (@($url).count -gt 0))
```
ferventcoder: What is $url and why are we using this?

src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
```diff
@@ -196,6 +197,31 @@ param(
     default { throw "7-Zip signalled an unknown error (code $exitCode)" }
   }
 
-  $env:ChocolateyPackageInstallLocation = $destination
-  return $destination
+  if ((test-path variable:url) -AND (@($url).count -gt 0))
+  {
+    # If our unzip resulted in a .tar file that matches our input zip file without it's extensions, unzip it
+    # does not work with archive names that contain a '.' before the extensions
+    $matchingtarname = "$destinationNoRedirection\$((split-path -leaf $url).split('.')[0]).tar"
```
ferventcoder: Tar name doesn't need to match, just search for a .tar file in the extracted directory.

ferventcoder: In some cases it may not match.

src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
```diff
@@ -196,6 +197,31 @@ param(
...
   }
 
-  $env:ChocolateyPackageInstallLocation = $destination
-  return $destination
+  if ((test-path variable:url) -AND (@($url).count -gt 0))
+  {
+    # If our unzip resulted in a .tar file that matches our input zip file without it's extensions, unzip it
+    # does not work with archive names that contain a '.' before the extensions
+    $matchingtarname = "$destinationNoRedirection\$((split-path -leaf $url).split('.')[0]).tar"
+    Write-Debug "Checking to see if we have a flie called $matchingtarname"
```
ferventcoder: Minor nitpick - typo "flie"

src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
ferventcoder: Nitpick on the name - use camelCase "$matchingTarName"

src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
ferventcoder: Please add the documentation for this up in the docs above. Please observe the way it is done currently and just add to it.

src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
```diff
@@ -196,6 +197,31 @@ param(
...
+    If ((Test-path $matchingtarname) -AND (!$RecursiveCall))
```
ferventcoder: This doesn't make it a recursive call if it can't continue to process down. Either we should make this a recursive call or just run unzip again on the tar. No need to introduce complexity here in the name of saving a small amount of repetition.

src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
DarwinJS: @ferventcoder - I don't have the bandwidth nor remember my test scenarios to update these.  I can, however, let you know what the thinking was behind each part.

In this case I followed the principle of keeping with the pattern of existing code - so inserted a position as the other's had.

I didn't know about documentation, but it is also an "internal" parameter that should only be called by the function itself under specific situations.

src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
DarwinJS: I intentionally kept the scope very narrow to the most common use case of a matching tar name inside another enclosing archive.

If the functionality were to unzip non-matching names it would seem to need to look for and unzip all archives if there are more than one as it is not clear which is the target.

This would also seem to cause it to unzip a bunch of sub-archives in conditions where it is not necessary - for instance archives that have sub-archives for an internal install routine to process if and when certain features are selected.

If it did unzip any archives it found - it's not clear to me how that might affect additional chocolatey procedures that work on the folder.

So I don't really have a broad enough understanding of the rest of the procedures this folder will be subject to, to go broader than the common use case I mentioned above - obviously someone working on the product for a long time could make those judgements.

src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
DarwinJS: It is a recursivecall - it's just not unlimited recursion. 

I used a recursive call to take advantage of any of the things the function already does because I don't know what hard lessons have gone into creating that code.  For instance, the existing code pre-creates the target folder: https://github.com/chocolatey/choco/blob/47729b0e546a3e93f947903d72ee984a60c02e70/src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1#L105)

Recursion also means that any improvements to the code in the function don't have to be duplicated in the sub-unzip code.

Based on my limited, past exposure to the issue of nested archives - I was thinking that most cases could be handled with one recursion.  I also don't know the situations under which multiple recursion levels might cause problems - so thought it best not to leave the code open to problems.

src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
DarwinJS: I had submitted a question about this at the time - but I believe $url was the only reliable way (variable was always available) to find the original zip name.  I think packagename is different than the original zip name by this point in processing and another internal variable was not created in every situation.

As expressed in further on comments - I was keeping the scope of what was looked for and unzipped as tight as possible - so hence the need to have the original archive name so as to only look for matching named sub-archives.

src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
DarwinJS: In general - I wanted to keep my changes to as limited of a scope as possible:
- so that I could devise and manually run tests to ensure the reliability of the most common, known scenarios that the code was supposed to address.
- Because this code is very broadly used - allowing the new functionality to do arbitrary things like unzip all found sub-archives or have no limit on recursions seemed to me to open it up to more unexpected behaviors (and bug fixes) going forward.

pauby:
After internal discussion, we have agreed that the decision on what is extracted from an archive is the maintainers decision:

* The additional code to extract the contents of a tar, extracted from a zip (for example) is minimal.
* The number of files that have embedded tar archive inside of them is very small (in our experience) so the impact this change would have would be very small.

We are also concerned about the change in behaviour here. 

I'll go ahead and close this PR. @DarwinJS thank you for the PR and apologies for taking so long to get to this.

PR #1051, opened 2016-11-14
(gh 1033)search on tags

### LINKED ISSUE #1033: Search by tags only
As per https://groups.google.com/forum/#!topic/chocolatey/HYrC9dUhdZ4, the idea here is to enhance the search/list functionality to be able to search on package tags.

Currently search/list is implemented in the execute_package_search method in file src\chocolatey\infrastructure.app\nuget\NugetList.cs - it calls packageRepository.Search() and gets back a list of packages and then filters that list based on things like Id, pre-release, etc. The list of packages returned is a list of IPackage interfaces, which are here: https://github.com/grendello/nuget/blob/master/src/Core/Packages/IPackage.cs

The actual class is (I think) https://github.com/grendello/nuget/blob/master/src/Core/Packages/DataServicePackage.cs and as you can see along with Id and prerelease it exposes a "tags" property. Thus it should be fairly simple to add a filter for tags.

My end use case is that I want to search for all packages that are installed that I have published - I've published them all with a tag 'aboutgolf' so I would like to do this:

#> choco list -l --tags aboutgolf

I had also prefixed all my package names with 'ag', however, using the IdStartsWith option is not reliable (there already exists a package called simply 'ag').

Apteryx0:
As per (GH 1033)

I looked at implementing a unit test, however, couldn't find a good existing one to emulate. If you can recommend one I'm happy to go back and add a unit test...

Closes #1033

### DIFF
--- a/src/chocolatey/infrastructure.app/commands/ChocolateyListCommand.cs
+++ b/src/chocolatey/infrastructure.app/commands/ChocolateyListCommand.cs
@@ -93,6 +93,9 @@ namespace chocolatey.infrastructure.app.commands
                  .Add("by-id-only",
                      "ByIdOnly - Only return packages where the id contains the search filter. Available in 0.9.10+.",
                      option => configuration.ListCommand.ByIdOnly = option != null)
+                 .Add("by-tag-only",
+                     "ByTagOnly - Only return packages where the search filter matches one of the tags. Available in 0.10.4+.",
+                     option => configuration.ListCommand.ByTagOnly = option != null)
                  .Add("id-starts-with",
                      "IdStartsWith - Only return packages where the id starts with the search filter. Available in 0.9.10+.",
                      option => configuration.ListCommand.IdStartsWith = option != null)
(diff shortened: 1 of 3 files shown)


Apteryx0:
I don't see why appveyor failed - can anyone help?

```
              [msbuild]   
              [msbuild] c:\projects\choco\src\SolutionVersion.cs(29,12): warning CS1699: Use command line option '/keyfile' or appropriate project settings instead of 'AssemblyKeyFileAttribute' [C:\projects\choco\src\chocolatey\chocolatey.csproj]
              [msbuild] infrastructure.app\nuget\NugetList.cs(83,104): error CS0103: The name 'searchTerm' does not exist in the current context [C:\projects\choco\src\chocolatey\chocolatey.csproj]
```

ferventcoder:
Okay that is weird since it was just available above. Could be some other issue.

ferventcoder:
Fixed, rebased onto stable, and adjusted the queries to work both remote and local. Merged into stable at https://github.com/chocolatey/choco/commit/52e7394eb85dbbb850ed4f3e1d20065999557306

Thanks for your contribution! It will be released in 0.10.6.

PR #1079, opened 2016-12-06
Add missing dependency

RichiCoder1:
Chocolatey takes a dependency on AlphaFS that doens't appear to get properly packed. Added an explicit dependency.


nuget/chocolatey.lib/chocolatey.lib.nuspec
```diff
@@ -22,7 +22,10 @@ This is the core package which allows Chocolatey to be embedded in your applicat
     <tags>apt-get yum machine repository chocolatey</tags>
     <iconUrl>https://raw.githubusercontent.com/chocolatey/choco/master/docs/logo/chocolateyicon.gif</iconUrl>
     <dependencies>
-      <dependency id="log4net" version="[2.0.3]" />
+      <group targetFramework="net40">
+        <dependency id="AlphaFS" version="2.0.1" />
```
ferventcoder: I was under the impression that this was baked in (ILMerge). It should be. So if it is not, let's correct the build scripts and not require an explicit dependency.

Issue #1053, opened 2016-11-15
Show the entire text to turn on the allowGlobalConfirmation flag


jkodroff:
When chocolatey is run, it asks for confirmation before installing.  There's an option to make confirmation happen implicitly, but the prompt to the user does not display the explicit command of the text to do this.

I propose the text of the confirmation show the entire command syntax to turn on automatic confirmation, e.g. 

```
Note: To confirm automatically next time, Run `choco feature enable -n allowGlobalConfirmation`
```

ferventcoder:
Now in backlog and up for grabs.

Cubix651:
Hi, I would like to contribute. I can do this as the first step.

Is it OK to change this note to: `Note: To confirm automatically next time, use '-y' or consider 'choco feature enable -n allowGlobalConfirmation'.`?

ferventcoder:
Yes.

RichiCoder1:
I didn't realize how much I wanted this until now. Yes please.
@Cubix651 @ferventcoder I'd also recommend showing the command on it's own line if possible. Make it easier to C+P.

PR #1100, opened 2016-12-19
(GH-1053) Update note about global confirmation

Cubix651:
This commit introduces easy to copy full command to enable
allowGlobalConfirmation which is printed during installation
of package.


ferventcoder:
Merged into stable at https://github.com/chocolatey/choco/commit/4b955e809f4877d8f49d6ea8c379023d7c276cca. Thanks for your contribution! This will be released with 0.10.4.

Issue #1098, opened 2016-12-19
Get-FtpFile fails with integer overflow when downloading file more than 2gb in size


jamestelfer:
### What You Are Seeing?
When downloading a >2Gb file via ftp with `Get-FtpFile` (called by `Get-ChocolateyWebFile`), the following error is observed:

```
Running Get-FtpFile -url 'ftp://<elided>/files/iso/Microsoft/SQL/2016/en_sql_server_2016_developer_with_service_pack_1_x64_dvd_9548071.
iso' -fileName 'C:\ProgramData\chocolatey\lib\sql-server-developer\tools\local.iso'
ERROR: The remote file either doesn't exist, is unauthorized, or is forbidden for url 'ftp://<elided>/files/iso/Microsoft/SQL/2016/en_s
ql_server_2016_developer_with_service_pack_1_x64_dvd_9548071.iso'. Cannot convert value "2147483648" to type "System.Int32". Error: "Value was eit
her too large or too small for an Int32."
 at Get-FtpFile, C:\ProgramData\chocolatey\helpers\functions\Get-FtpFile.ps1: line 196
at Get-ChocolateyWebFile, C:\ProgramData\chocolatey\helpers\functions\Get-ChocolateyWebFile.ps1: line 345
at <ScriptBlock>, C:\ProgramData\chocolatey\lib\sql-server-developer\tools\chocolateyInstall.ps1: line 50
at <ScriptBlock>, C:\ProgramData\chocolatey\helpers\chocolateyScriptRunner.ps1: line 48
at <ScriptBlock>, <No file>: line 1
```
### What is Expected?
The file should download successfully.

### How Did You Get This To Happen? (Steps to Reproduce)

1. Host a large (>2Gb) file on an FTP server to which you have access
2. Download file using the `Get-FtpFile` function

### Output Log
https://gist.github.com/jamestelfer/f10314dab426b9e47c447ca3f9db76bd

jamestelfer:
Note also in the log that the local destination file remained locked after the FTP download failed, hindering choco's cleanup.

ferventcoder:
That is definitely an issue. Marked for fixing soon.

ferventcoder:
Also marking this as up for grabs if you are interested

jamestelfer:
I've forked the repo, let me have a quick squiz. Looking at the `Get-WebFile` source, it shouldn't be difficult for the actual fix (for this and #1099). Not sure about the testing strategy for this one though aside from manual checks.

[...]

PR #1104, opened 2016-12-21
(GH-1098) Change counter and length variables to long

jamestelfer:
This change avoids an integer overflow error, and brings the implementation
into line with the `Get-WebFile` function.

There is further discussion on #1098 around modifying the way the streams are handled to ensure that the output stream is closed on failure. The initial commit deals only with the core of the issue; any further changes can be added subsequently.


jamestelfer:
Note that as far as I can tell the changes as of now are complete and correct. Let me know if anything further is required.

ferventcoder:
Merged into stable at https://github.com/chocolatey/choco/commit/e1e3865f60a062a3b20483e1bf64079008400f11

ferventcoder:
@jamestelfer thanks for fixing this issue!

PR #1127, opened 2017-01-04
(GH-906) Fix source path used when installing from nuspec file.

### LINKED ISSUE #906: choco install relativepath/to/some.nuspec fails
For installs directly against a .nuspec, it appears that the .nupkg is runtime compiled into the working folder, but then when the call to 'install' is passed internally in choco, the source folder is given as the same folder as the .nuspec.

So if do this:
c:\somefolder> cinst -y subfolder\my.nuspec

The .nupkg ends up in "somefolder" but choco errors out that it can't find the package in c:\somefolder\subfolder.

choco version 0.10.0.

woodburysoft:
Previously when the install target was a nuspec file and the output
folder was not set to the folder that contained the nuspec file then
the installation failed because the nupkg file could not be found.
Update sets the sources parameter to the output folder during the
packaging of the nuspec file.

Closes #906

### DIFF
--- a/src/chocolatey.tests.integration/scenarios/PackScenarios.cs
+++ b/src/chocolatey.tests.integration/scenarios/PackScenarios.cs
@@ -70,6 +70,12 @@ namespace chocolatey.tests.integration.scenarios
 
                 File.Exists(package_path).ShouldBeTrue();
             }
+
+            [Fact]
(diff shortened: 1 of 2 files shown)


ferventcoder:
Merged into stable at https://github.com/chocolatey/choco/commit/94a448d51b9edc56ce0601a3c372f564f7d64193

PR #1128, opened 2017-01-04
(GH-523) Be explicit about the version when installing a nupkg.

### LINKED ISSUE #523: When pointing to a nupkg, choco should use only that nupkg to install and not a newer version in the same directory
If you have two versions of a package in a directory and you point directly to the older one, it should be installed. Use version as part of the install arguments.

woodburysoft:
Previously if there were two versions of a nupkg in the same
directory and you tried to install the older one, it would install
the newer one. Now the nupkg version is explicitly specified as
part of the install arguments.

Closes #523

### DIFF
--- a/src/chocolatey/infrastructure.app/services/NugetService.cs
+++ b/src/chocolatey/infrastructure.app/services/NugetService.cs
@@ -359,6 +359,7 @@ folder.");
                     else
                     {
                         var packageFile = new OptimizedZipPackage(_fileSystem.get_full_path(packageName));
+                        version = packageFile.Version;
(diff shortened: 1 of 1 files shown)


woodburysoft:
I haven't added any tests for this as I couldn't find any existing ones that touched this area but I'm happy to add some if someone can point me in the right direction.

ferventcoder:
Merged into stable at https://github.com/chocolatey/choco/commit/6bc9dd22c1bb491051f04df1f08ab5014e191e3b. Again, thank you for your contributions! Really appreciate your help.

PR #1137, opened 2017-01-14
(GH-473) Correct exit code when help is requested

### LINKED ISSUE #473: Help should exit zero if called with the help switch, otherwise non-zero on bad commands
Also related to #468 

Badly passed commands display the help menu and exit with non-zero, but also explicitly calling for help will exit non-zero.

Cubix651:
I've introduced new property UnsuccessfulParsing in ChocolateyConfiguration to differ if parsing failed or not (respectively to show help menu or not to show). Thanks to that choco returns proper exit code. Moreover I've added 2 test cases. I hope everything is OK.

Closes #473

### DIFF
--- a/src/chocolatey.console/Program.cs
+++ b/src/chocolatey.console/Program.cs
@@ -99,10 +99,10 @@ namespace chocolatey.console
                     }
                 }
 
-                if (config.HelpRequested)
+                if (config.HelpRequested || config.UnsuccessfulParsing)
                 {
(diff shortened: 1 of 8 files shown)


ferventcoder:
Merged into stable at https://github.com/chocolatey/choco/commit/68c3b16c3be890d416f87f4884eb6de211dcab4b.

Thanks for your contributions!

Issue #1020, opened 2016-10-16
Build fails on newer Mono versions


Daniel15:
Chocolately successfully builds on a pretty old version of Mono (as can be seen in the [Travis build](https://travis-ci.org/chocolatey/choco)), but it fails on newer 4.x versions of Mono. I tried fixing this myself but ran into an issue.

The first issue was that Mono no longer has its libraries in `/usr/lib/mono/4.0`, they all moved to `4.5` when Mono added support for .NET Framework 4.5 as a drop-in replacement for 4.0 (runtime version is still `4.0.30319` but the assemblies are in `/usr/lib/mono/4.5`). I fixed that with the following patch:

``` diff
diff --git a/build.sh b/build.sh
index 648b60c..719b2f4 100755
--- a/build.sh
+++ b/build.sh
@@ -25,7 +25,6 @@ displayUsage $1
 export PKG_CONFIG_PATH=/opt/local/lib/pkgconfig:/Library/Frameworks/Mono.framework/Versions/Current/lib/pkgconfig:$PKG_CONFIG_PATH

 #mono ./lib/NAnt/NAnt.exe /logger:"NAnt.Core.DefaultLogger" /nologo /quiet /f:"$(cd $(dirname "$0"); pwd)/.build/default.build" /D:build.config.settings="$(cd $(dirname "$0"); pwd)/.uppercut" /D:microsoft.framework="mono-3.5" $*
-mono --runtime=v4.0.30319 ./lib/NAnt/NAnt.exe /logger:"NAnt.Core.DefaultLogger" /nologo /quiet /f:"$(cd $(dirname "$0"); pwd)/.build/default.build" /D:build.config.settings="$(cd $(dirname "$0"); pwd)/.uppercut" /D:microsoft.framework="mono-4.0" /D:run.ilmerge="false" /D:run.nuget="false" $*
+mono --runtime=v4.0.30319 ./lib/NAnt/NAnt.exe /logger:"NAnt.Core.DefaultLogger" /nologo /quiet /f:"$(cd $(dirname "$0"); pwd)/.build/default.build" /D:build.config.settings="$(cd $(dirname "$0"); pwd)/.uppercut" /D:microsoft.framework="mono-4.5" /D:run.ilmerge="false" /D:run.nuget="false" $*

 #/quiet /nologo /debug /verbose /t:"mono-4.0"
diff --git a/lib/NAnt/NAnt.exe.config b/lib/NAnt/NAnt.exe.config
[...]

PR #1153, opened 2017-01-31
Dockerfile to Enable Mono Based Choco Builds

Linuturk:
This PR provides a Dockerfile and choco_wrapper script to enable easy Linux based choco.exe usage.

The Dockerfile copies the entire source tree into the container and then builds choco.exe using the build.sh script.

The choco_wrapper is copied into /usr/local/bin/choco and allows for easier execution of choco.exe using mono.

I suggest linking the repository to Docker Hub and releasing various versions of the Docker image using their build process.

### DIFF
--- a/docker/Dockerfile.linux
+++ b/docker/Dockerfile.linux
@@ -1,0 +1,17 @@
+FROM mono:3.12.1
+
+MAINTAINER Justin Phelps <linuturk@onitato.com>
+
+COPY . /usr/local/src/choco/
+
+WORKDIR /usr/local/src/choco
+RUN chmod +x build.sh
+RUN chmod +x zip.sh
+RUN ./build.sh
+
+WORKDIR /usr/local/bin
+RUN ln -s /usr/local/src/choco/build_output/chocolatey
+
+COPY docker/choco_wrapper /usr/local/bin/choco
+
+WORKDIR /root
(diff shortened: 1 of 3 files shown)


ferventcoder:
Added more to that second bit.

Linuturk:
Thanks for reviewing.

- Let me play with the build arguments for the docker image and see if the sub folder is possible. I'm not an expert in Windows docker images so I can't speak to the need for separate windows / linux sub folders. It might make more sense to name the files `Dockerfile.linux` and add a `Dockerfile.windows` when/if that's contributed.
- Sure, I'll merge the commits, no problem.
- I think I understand what you mean here. I'll push up the changes to this PR and you can tell me yes or no.

Linuturk:
Hopefully that satisfies your change requests.

ferventcoder:
That looks great!

Linuturk:
See this is still marked as needing changes. Anything else I can do?

ferventcoder:
No sir. All good to go. Merged in. Thanks!

Linuturk:
:+1: Now we've got the right stuff in place for an official Docker Hub build! Excited for that next step.

petemounce:
Linking bazelbuild/bazel#4559 for a +1. #1020 is a likely fix, since it would allow us to use supported mono containers which (presumably!) have up-to-data SSL certs within.

Issue #962, opened 2016-09-14
Before modify script not called when install command is used to upgrade


chrisdrobison:
### What You Are Seeing?

If I have a package already installed and then I call:

```choco install package --version <next version>```

The before modify script doesn't get run, even though that seems to me like it should be an upgrade scenario. The script does get run if I call:

```choco upgrade package --version <next version>```

### What is Expected?

IMHO, if the upgrade command can install a package that doesn't exist, then the install command should detect an upgrade scenario if you aren't installing side-by-side.

### How Did You Get This To Happen? (Steps to Reproduce)

Detailed above

ferventcoder:
Reference conversation - https://gitter.im/chocolatey/choco?at=57d876cab8d4c138408ed7d9

ferventcoder:
Typically we see folks using install for install and not for upgrades. I'm not fully aware of what else may be missed when using install with version as a method for upgrades. It's typically not recommended. 

One thing I think we may easily be able to do is see that this is installed (with an older version) and call upgrade instead when a newer version is passed against an existing pre-installed package (not side by side).

Isopolito:
Hi, is this still up for grabs? If so I'll start researching it. My schedule is pretty tight between work, family, etc. so it might take me a while if that's okay

ferventcoder:
Still up for grabs :D

PR #1132, opened 2017-01-11
(GH-533)

Spitzbub:
Implements a subset of the functionality, support for packages.config
files in the upgrade command. Also extended the tests to test the new
code. I did not see any value in implementing package.config support for
uninstall because the problem I am interested in solving is being able
to upgrade multiple packages each with a specific version in one
command. Since uninstall (generally?) doesn't require specifying a
version this can be done on the uninstall command line, i.e. just give
the package list to uninstall.

PR #1204, opened 2017-03-20
(GH-962) Upgrade outdated on install

Isopolito:
Hi, I made the changes for issue #962 (https://github.com/chocolatey/choco/issues/962) and submitted them on this PR. I'm still working out how to get the unit test(s) created for it, but in the meantime i wanted to submit this since I have a few questions:

* Is deep copying a config and only changing PackageNames to what should be upgraded sufficient (line 1337), or do other things need to be modified in the new config before sending it off to upgrade_run? 

* Is just removing a package name from config.PackageNames (line 1332) enough to make it disappear from the install process?

Thanks!


ferventcoder:
@Isopolito it may be best to wait for https://github.com/chocolatey/choco/pull/1132 to be implemented so you won't need to do this.

ferventcoder:
Also, please read CONTRIBUTING - it does state we ***never*** want to see merge commits in pull requests.

Isopolito:
Sorry about that, I'm new to this process. I read over contributing but still messed that up. I'll look over #1132 and wait for that to be implemented. Is there a preferred way to back out the merge commit?

ferventcoder:
> Is there a preferred way to back out the merge commit?

@Isopolito  Yes, `choco install gitextensions`, ensure that gets configured appropriately and use it to help you back up to the older commit, then do a force push back to the branch on that (but be careful, force pushes should only be used in feature branches). We have an example of this in contributing.

PR #1211, opened 2017-03-28
(GH-1210) Prevent nested chocolatey directories in CacheLocation

### LINKED ISSUE #1210: [API] Chocolatey file cache adds a 'chocolatey' directory on each install
Note that this only occurs when using the `chocolatey.dll` and installing multiple packages in a single process. Choco creates the `CacheLocation` by appending a `chocolatey` directory to the `TEMP` environment variable (https://github.com/chocolatey/choco/blob/master/src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs#L198). Then later it sets the `PATH` environment variable from the `CacheLocation` (https://github.com/chocolatey/choco/blob/master/src/chocolatey/infrastructure.app/configuration/EnvironmentSettings.cs#L79). If one installs multiple packages in the same process, each install will cache to a deeper chocolatey directory until eventually `path too long` errors start failing packages. This is common in an application leveraging chocolatey as a library such as Boxstarter. This addresses the issue submitted at https://github.com/mwrock/boxstarter/issues/241.

Soon to follow I will submit a PR tha twill only append a `chocolatey` directory if one does not already exist.

mwrock:
This ensures that the CacheLocation only appends chocolatey to the TEMP variable if it has not already been added. Otherwise multiple package installs in the same process result in an ever deeper nested chocolatey cache folders.

Closes #1210

### DIFF
--- a/src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
+++ b/src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
@@ -195,7 +195,13 @@ namespace chocolatey.infrastructure.app.builders
         private static void set_config_items(ChocolateyConfiguration config, ConfigFileSettings configFileSettings, IFileSystem fileSystem)
         {
             config.CacheLocation = Environment.ExpandEnvironmentVariables(set_config_item(ApplicationParameters.ConfigSettings.CacheLocation, configFileSettings, string.IsNullOrWhiteSpace(configFileSettings.CacheLocation) ? string.Empty : configFileSettings.CacheLocation, "Cache location if not TEMP folder."));
-            if (string.IsNullOrWhiteSpace(config.CacheLocation)) config.CacheLocation = fileSystem.combine_paths(fileSystem.get_temp_path(), "chocolatey"); // System.Environment.GetEnvironmentVariable("TEMP");
+            if (string.IsNullOrWhiteSpace(config.CacheLocation)) {
+                config.CacheLocation = fileSystem.get_temp_path(); // System.Environment.GetEnvironmentVariable("TEMP");
+                if(!config.CacheLocation.EndsWith("chocolatey")) {
+                    config.CacheLocation = fileSystem.combine_paths(fileSystem.get_temp_path(), "chocolatey");
+                }
+            }
+
             // if it is still empty, use temp in the Chocolatey install directory.
             if (string.IsNullOrWhiteSpace(config.CacheLocation)) config.CacheLocation = fileSystem.combine_paths(ApplicationParameters.InstallLocation, "temp");
             


mwrock:
The integration test failures don't appear to be related to this PR.

ferventcoder:
They are not related. Still chasing that down. Has issues when run together

src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
```diff
@@ -195,7 +195,13 @@ private static void set_machine_sources(ChocolateyConfiguration config, ConfigFi
         private static void set_config_items(ChocolateyConfiguration config, ConfigFileSettings configFileSettings, IFileSystem fileSystem)
         {
             config.CacheLocation = Environment.ExpandEnvironmentVariables(set_config_item(ApplicationParameters.ConfigSettings.CacheLocation, configFileSettings, string.IsNullOrWhiteSpace(configFileSettings.CacheLocation) ? string.Empty : configFileSettings.CacheLocation, "Cache location if not TEMP folder."));
-            if (string.IsNullOrWhiteSpace(config.CacheLocation)) config.CacheLocation = fileSystem.combine_paths(fileSystem.get_temp_path(), "chocolatey"); // System.Environment.GetEnvironmentVariable("TEMP");
+            if (string.IsNullOrWhiteSpace(config.CacheLocation)) {
+                config.CacheLocation = fileSystem.get_temp_path(); // System.Environment.GetEnvironmentVariable("TEMP");
+                if(!config.CacheLocation.EndsWith("chocolatey")) {
+                    config.CacheLocation = fileSystem.combine_paths(fileSystem.get_temp_path(), "chocolatey");
```
ferventcoder: It looks like formatting is a bit off, but I will fix that as I pull it in.

mwrock: oh what did I format wrong? Its not clear to me what's off but I'm also not familiar with the formatting standards.

ferventcoder: Actually, I'm a bit perplexed why this would happen. Noting that it doesn't set it again unless it is empty, why is it nesting?

mwrock: because it sets the PATH from CacheLocation at https://github.com/chocolatey/choco/blob/master/src/chocolatey/infrastructure.app/configuration/EnvironmentSettings.cs#L79. So the sequence is like this:
1. CacheLocation = PATH + chocolatey
2. Then PATH is set to CacheLocation
3. Next Install - CacheLocation = PATH + chocolatey but now PATH is the old path + chocolatey

ferventcoder: Because it sets TEMP, not cache location! I get it now.

ferventcoder:
Merged into stable at https://github.com/chocolatey/choco/commit/3046d60700ddbffa628c88ee4d8c1c3240eb76d6

ferventcoder:
Thanks for getting that over so quickly!

src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
flcdrg: If `config.CacheLocation` ends with a trailing slash, then `config.CacheLocation.EndsWith("chocolatey")` will not work

ferventcoder: @flcdrg do you want to create a followup issue?

flcdrg: Sure.. though I'm only guessing (from the fact that the newer Boxstarter release still has the same problem)

dviry:
This still happens to me on 0.10.8, so I am not sure why this is closed (unless I am missing something)...
e.g.
`C:\Users\Administrator\AppData\Local\Temp\chocolatey\chocolatey\chocolatey\chocolatey`

flcdrg:
Are you using Chocolatey by itself or with Boxstarter?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

dviry:
Yes indeed within Boxstarter (latest v2.10.3).

Thanks

ferventcoder:
@dviry then you are not using Chocolatey v0.10.8 if you are using Boxstarter. I have a feeling that is the same thing you are running into over in the other issue you commented on.

Boxstarter uses a baked in Chocolatey.lib. It looks like you are using choco.exe when you use Boxstarter, but you are actually not.

flcdrg:
Boxstarter had to revert back a previous version of Chocolatey due to another problem.

See https://github.com/mwrock/boxstarter/issues/241#issuecomment-335368855 for a workaround

Issue #1231, opened 2017-04-03
[API] Chocolatey file cache still adds a 'chocolatey' directory on each install


flcdrg:
Still seeing the problem described in #1211 in Boxstarter 2.9.5, which presumably was built with Chocolatey 0.10.4+.

Guessing if it could be because CacheLocation has as trailing slash?

PR #1232, opened 2017-04-03
(GH-1231) Use DirectoryInfo to evaluate CacheLocation

mwrock:
The former fix (GH-1210) fails when the CacheLocation ends with a slash. This uses a DirectoryInfo instance to examine the last directory in the path in a more robust manner

### DIFF
--- a/src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
+++ b/src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
@@ -199,8 +199,8 @@ namespace chocolatey.infrastructure.app.builders
                 config.CacheLocation = fileSystem.get_temp_path(); // System.Environment.GetEnvironmentVariable("TEMP");
                 // TEMP gets set in EnvironmentSettings, so it may already have 
                 // chocolatey in the path when it installs the next package from
-                // the API. 
-                if(!config.CacheLocation.EndsWith("chocolatey")) {
+                // the API.
+                if(!fileSystem.get_directory_info_for(config.CacheLocation).Name.is_equal_to("chocolatey")) {
                     config.CacheLocation = fileSystem.combine_paths(fileSystem.get_temp_path(), "chocolatey");
                 }
             }
--- a/.build.custom/nugetPrepare.post.step
+++ b/.build.custom/nugetPrepare.post.step
@@ -42,14 +42,6 @@
   </target>
   
   <target name="copy_legal">
-    <echo level="Warning" message="Copying LICENSE from '${dirs.current}${path.separator}LICENSE' to nuget packages"/>
-    <copy file="${dirs.current}${path.separator}LICENSE" tofile="${dirs.drop.nuget}${path.separator}chocolatey${path.separator}tools${path.separator}chocolateyInstall${path.separator}LICENSE.txt" failonerror="true" />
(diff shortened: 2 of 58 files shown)


mwrock:
I just tested the appveyor drop with boxstarter and this looks to be working.

mwrock:
fixes #1231

src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
```diff
@@ -199,8 +200,8 @@ private static void set_config_items(ChocolateyConfiguration config, ConfigFileS
                 config.CacheLocation = fileSystem.get_temp_path(); // System.Environment.GetEnvironmentVariable("TEMP");
                 // TEMP gets set in EnvironmentSettings, so it may already have 
                 // chocolatey in the path when it installs the next package from
-                // the API. 
-                if(!config.CacheLocation.EndsWith("chocolatey")) {
+                // the API.
+                if(new DirectoryInfo(config.CacheLocation).Name.ToLower() != "chocolatey") {
```
ferventcoder: Use `fileSystem.get_directory_info()`  and use either `.to_lower()` or `Name.is_equal_to()`. The latter is preferred.

mwrock:
hmm this is dying with `'string' does not contain a definition for 'is_equal_to'` but that seems to be a valid extension method used throughout this file. My c# is very rusty. Will look more later tonight.

ferventcoder:
> hmm this is dying with 'string' does not contain a definition for 'is_equal_to' but that seems to be a valid extension method used throughout this file. My c# is very rusty. Will look more later tonight.

My guess?

`new DirectoryInfo(config.CacheLocation).Name.ToLower() != "chocolatey"` was changed to `fileSystem.get_directory_info(config.CacheLocation).Name.is_equal_to() != "chocolatey"` instead of `!fileSystem.get_directory_info(config.CacheLocation).Name.is_equal_to("chocolatey")`. The first change would definitely produce the error, it's needs a value passed to the method.

mwrock:
well `!fileSystem.get_directory_info(config.CacheLocation).Name.is_equal_to("chocolatey")` is exacty what's in this pr.

mwrock:
actually its `get_directory_info_for` not `get_directory_info` but I think thats right.

ferventcoder:
So it is. So it is.

mwrock:
ok. I just ran `MSBuild.exe C:\dev\choco\src\chocolatey\chocolatey.csproj /t:Build` and it compiles just fine. So I'm not entirely sure why that fails in appveyor. The appveyor log is huge and theer might be something in there I am missing.

BTW: the build.bat breaks on nuget.exe in the choco repo. If I swap in the latest nuget.exe the nuget installs work but then the build fails in ilmerge.exe. I think the nuget problems are due to API incompatibilities with older nuget clients (just an educated guess).

Anyways I found that vanilla msbuild ended up being the easiest way for me to locally sanity check the build.

mwrock:
Looking again at the appveyor log I actually think this is a test setup issue and not related directly to the config builder itself. Perhaps somthing needs to be imported or added to a container.

ferventcoder:
Don't get too hung up on the appveyor issues. Also, as a committer, you may think about using stable and not master.

mwrock:
closing in favor of #1233 targeting stable

PR #1254, opened 2017-04-15
(GH-1237) Implemented choco list --id-only.

### LINKED ISSUE #1237: Add command to list packages only by name
This is related to a question on [stackoverflow](http://stackoverflow.com/questions/43167087/how-to-get-a-list-of-packages-from-one-machine-and-install-in-another-with-choco), there should be a simple way to list the packages without package versions.

Analogous to a `dir /b`, there should be a `clist -b` or something that simply lists package names, one per line without versions. This makes things much simpler to automate than dealing with xml files or powershell.

Even if there is a simple way to generate xml files, it's still useful to allow a simple list of packages we can `xargs` into a new command line.

torhovland:
I named the option --id-only for consistency with the existing --by-id-only, used for searching, and for naming consistency with --local-only.

Closes #1237

### DIFF
--- a/src/chocolatey.resources/helpers/ChocolateyTabExpansion.ps1
+++ b/src/chocolatey.resources/helpers/ChocolateyTabExpansion.ps1
@@ -41,7 +41,7 @@ $proNewOptions = " --file='' --build-package --file64='' --from-programs-and-fea
 $proUninstallOptions = " --from-programs-and-features"
 
 $commandOptions = @{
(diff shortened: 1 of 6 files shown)


ferventcoder:
Rebased and merged into stable at https://github.com/chocolatey/choco/commit/082e062e6aa35f4622f6d8df9641ef057891238a. Thanks for your contributions! This will be in 0.10.6

PR #1275, opened 2017-04-29
FIX: Install-ChocolateyPath won't add a path that is a substring of a…

glachancecmaisonneuve:
**FIX:** Install-ChocolateyPath won't add a path that is a substring of an existing path
**FIX:** Install-ChocolateyPath adds duplicate paths if the added path ends in a folder separator and the one already in the path does not.
**New Feature:** Check if path exists on disk before adding it.  Default behavior is to ignore the install if the path doesnt exist.  A Force parameter was added for weird cases where that would be desirable.
**New Feature:** Install-ChocolateyPath now understands cmd-type variables.

A complete pester test file is available.  I didn't add it to the commit as I haven't seen any for powershell methods in the project.

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1
+++ b/src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1
@@ -1,6 +1,6 @@
-# Copyright Â© 2017 Chocolatey Software, Inc.
-# Copyright Â© 2015 - 2017 RealDimensions Software, LLC
-# Copyright Â© 2011 - 2015 RealDimensions Software, LLC & original authors/contributors from https://github.com/chocolatey/chocolatey
+# Copyright © 2017 Chocolatey Software, Inc.
+# Copyright © 2015 - 2017 RealDimensions Software, LLC
+# Copyright © 2011 - 2015 RealDimensions Software, LLC & original authors/contributors from https://github.com/chocolatey/chocolatey
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -54,6 +54,9 @@ Install-ChocolateyPath -PathToInstall "$($env:SystemDrive)\tools\gittfs"
 .EXAMPLE
 Install-ChocolateyPath "$($env:SystemDrive)\Program Files\MySQL\MySQL Server 5.5\bin" -PathType 'Machine'
 
+.EXAMPLE
+Install-ChocolateyPath "%ANDROID_HOME%\Tools" -PathType 'Machine'
+
 .LINK
 Install-ChocolateyEnvironmentVariable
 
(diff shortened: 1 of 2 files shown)


glachancecmaisonneuve:
Test file is available here: https://github.com/chocolatey/choco/commit/aba1d04fdc3a90fa2fef786861986f19c85b9e41#diff-94c826cb9ab70d9c220a3d018358890f

ferventcoder:
I see the Pester file.

src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1
```diff
@@ -1,6 +1,6 @@
-# Copyright © 2017 Chocolatey Software, Inc.
-# Copyright © 2015 - 2017 RealDimensions Software, LLC
-# Copyright © 2011 - 2015 RealDimensions Software, LLC & original authors/contributors from https://github.com/chocolatey/chocolatey
+# Copyright � 2017 Chocolatey Software, Inc.
+# Copyright � 2015 - 2017 RealDimensions Software, LLC
+# Copyright � 2011 - 2015 RealDimensions Software, LLC & original authors/contributors from https://github.com/chocolatey/chocolatey
```
ferventcoder: It looks like you possibly changed line endings on this file. That will break it for authenticode signing.

src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1
```diff
@@ -69,42 +72,38 @@ Get-ToolsLocation
 param(
   [parameter(Mandatory=$true, Position=0)][string] $pathToInstall,
   [parameter(Mandatory=$false, Position=1)][System.EnvironmentVariableTarget] $pathType = [System.EnvironmentVariableTarget]::User,
+  [parameter(Mandatory=$false, Position=2)][switch] $Force = $false,
   [parameter(ValueFromRemainingArguments = $true)][Object[]] $ignoredArguments
 )
-  Write-Debug "Running 'Install-ChocolateyPath' with pathToInstall:`'$pathToInstall`'";
-  $originalPathToInstall = $pathToInstall
+  Set-StrictMode -Version 2
```
ferventcoder: This won't work. On PowerShell 4/5 systems, you have to install the v2 engine so that you can do this. This breaks those systems.

src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.tests.ps1
```diff
@@ -0,0 +1,126 @@
+﻿. $PSScriptRoot\Install-ChocolateyPath.ps1
```
ferventcoder: This file is in a non-good location. If we are going to introduce Pester tests again (or I should say when we do, as they are in the original chocolatey/chocolatey), they need to be in a test project directory.

src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1
```diff
@@ -69,42 +72,38 @@ Get-ToolsLocation
...
-      Set-EnvironmentVariable -Name 'Path' -Value $actualPath -Scope $pathType
-    }
-
-    #add it to the local path as well so users will be off and running
-    $envPSPath = $env:PATH
-    $env:Path = $envPSPath + $statementTerminator + $pathToInstall
+    $newPath = $pathToInstall + [System.IO.Path]::PathSeparator + (Get-EnvironmentVariable -Name 'Path' -Scope $pathType)
+    
+    Set-EnvironmentVariable -Name 'PATH' -Value $newPath -Scope $pathType
+    Update-SessionEnvironment
```
ferventcoder: This is already a PR up for this particular file and this appears to be repeating some of what is being handled there. You may wish to review some of their changes as well in this area.

glachancecmaisonneuve:
Where do you suggest I create/put the pester file?
The PR in waiting (327) should be closed.  ctaggart said he is unable to continue the PR (https://github.com/chocolatey/choco/pull/327#issuecomment-287627139)
I'll try and fix my line endings to it doesn't mess up on commit.  auto-crlf does not work in this case.

src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1
glachancecmaisonneuve: The PR in waiting (327) should be closed.  ctaggart said he is unable to continue the PR (https://github.com/chocolatey/choco/pull/327#issuecomment-287627139)

src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1
glachancecmaisonneuve: Will remove.  Thanks

src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1
glachancecmaisonneuve: Will fix.  Thanks

src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.tests.ps1
glachancecmaisonneuve: Where do you suggest I create/put the pester file?

src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1
jberezanski: @ferventcoder In what circumstances exactly did you encounter problems with that? I run scripts containing
```
#Requires -Version 2
Set-StrictMode -Version 2
```
on Windows 10 1511 and 1607 systems and on Server 2012 R2 with both .NET 2.0 and PowerShell 2.0 removed and no issues occur. The docs for [Set-StrictMode](https://msdn.microsoft.com/en-us/powershell/reference/5.1/microsoft.powershell.core/set-strictmode) and [#Requires](https://msdn.microsoft.com/en-us/powershell/reference/5.1/microsoft.powershell.core/about/about_requires) imply that the number is interpreted as a minimum, not an exact version.

ferventcoder: @jberezanski remove the PowerShell v2 engine feature and `#Requires -Version 2` fails. I've seen it before. Would love for the minimum to be true, but have not seen it implemented in that way.

src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1
ferventcoder: I don't believe that was the one I was talking about.

src/chocolatey.resources/helpers/functions/Install-ChocolateyPath.ps1
jberezanski: (The line of code in question was `Set-StrictMode`, not `#Requires`, but the behavior should be the same for both constructs.)
On what systems exactly have you seen that? I'm asking because, as I wrote before, this works as expected on all machines I have access to.
Besides, two widespread Chocolatey extensions (`chocolatey-windowsupdate.extension` and `chocolatey-visualstudio.extension`) use it and there has not been a single issue reported.

Or perhaps you were thinking about executing `powershell.exe -version 2`? This is asking specifically for the 2.0 engine, so naturally it fails if that engine is not available.

PR #1308, opened 2017-05-23
(GH-1279) Fix resending cookies in Get-WebFile

### LINKED ISSUE #1279: Cookie isn't being resent in Get-WebFile with a redirect
This package https://chocolatey.org/packages/server-jre8 is currently having to use .NET WebClient to download as for some reason if the script tries to set the Cookie using the Options collection, the cookie isn't maintained in the HTTP requests.

I wonder if this is because the cookie is added using the request url as the cookie domain, but in this case the server responds with a 302 that redirects to an akamai location. 

Presumably because the second request then doesn't include the cookie, the download fails.

Tracking with Fiddler, I can see that the WebClient call does send the cookie on the request following the redirect response.

Cubix651:
HttpWebRequest does not resend cookies after redirection. This commit resolves that problem by manual loop.

Closes #1279.

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Get-WebFile.ps1
+++ b/src/chocolatey.resources/helpers/functions/Get-WebFile.ps1
@@ -107,95 +107,97 @@ param(
     #continue on
   }
 
-  $req = [System.Net.HttpWebRequest]::Create($url);
-  $defaultCreds = [System.Net.CredentialCache]::DefaultCredentials
-  if ($defaultCreds -ne $null) {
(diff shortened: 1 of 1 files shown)


ferventcoder:
Hi @Cubix651 - there is a lot of change here and it is a bit hard to reason about the changes. Sorry I didn't comment before. We'll need to find a better way to see what has changed here.

AdmiringWorm:
@ferventcoder https://github.com/chocolatey/choco/pull/1308/files?w=1
Easier to see the changes

PR #1314, opened 2017-05-27
(GH-1313) Allow properties to be passed to 'pack'

### LINKED ISSUE #1313: pack - pass arbitrary properties like nuget pack
### What You Are Seeing?

Unlike `nuget pack`, I cannot pass properties to `choco pack` which blocks implementation of Microsoft/vswhere#30.

### What is Expected?

I can pass `--properties "foo=1;bar=baz;"` to `choco pack` (using the same syntax as `nuget pack` for uniformity). I want to be able to build packages in either configuration for 1) testing, and 2) to make sure no one who might have rearranged content broke package build. Only our release configuration would be published, but we want to build both flavors in case devs only build one (typically debug).

heaths:
Passes properties through to the nuget pack command to replace variables
in the nuspec file.

Note that I have no yet cleared the CLA. I need to run this by our legal department but any preliminary feedback would be appreciated so the PR is ready to merge. Given it's a holiday weekend and I'm out on vacation next week, I may not be able to sign the CLA for a week or two. (I'm in the office one day next week so maybe I can get clearance if our legal reps are in.)

Closes #1313

### DIFF
--- a/src/chocolatey.tests.integration/scenarios/PackScenarios.cs
+++ b/src/chocolatey.tests.integration/scenarios/PackScenarios.cs
@@ -115,6 +115,54 @@ namespace chocolatey.tests.integration.scenarios
             }
         }
 
+        [Concern(typeof(ChocolateyPackCommand))]
+        public class when_packing_with_properties : ScenariosBase
+        {
+            private readonly string package_path = Path.Combine(Scenario.get_top_level(), "test-package.0.1.0.nupkg");
+
+            public override void Context()
+            {
+                base.Context();
+
+                Scenario.reset(Configuration);
+
+                Configuration.Version = "0.1.0";
+                Configuration.PackCommand.Properties.Add("commitId", "1234abcd");
+                Scenario.add_files(new[] { new Tuple<string, string>("myPackage.nuspec", NuspecContentWithVariables) });
+
+                if (File.Exists(package_path))
+                {
+                    File.Delete(package_path);
+                }
+            }
+
+            public override void Because()
+            {
+                MockLogger.reset();
+                Service.pack_run(Configuration);
+            }
+
+            [Fact]
+            public void generated_package_should_be_in_current_directory()
+            {
+                var infos = MockLogger.MessagesFor(LogLevel.Info);
+                infos.Count.ShouldEqual(2);
+                infos[0].ShouldEqual("Attempting to build package from 'myPackage.nuspec'.");
+                infos[1].ShouldEqual(string.Concat("Successfully created package '", package_path, "'"));
+
+                File.Exists(package_path).ShouldBeTrue();
+            }
+
+            [Fact]
+            public void property_settings_should_be_logged_as_debug_messages()
+            {
+                var messages = MockLogger.MessagesFor(LogLevel.Debug);
+                messages.Count.ShouldEqual(2);
+                messages.Contains("Setting property 'commitId': 1234abcd");
+                messages.Contains("Setting property 'version': 0.1.0");
+            }
+        }
+
         private const string NuspecContent = @"<?xml version=""1.0"" encoding=""utf-8""?>
 <package xmlns=""http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd"">
   <metadata>
@@ -128,6 +176,24 @@ namespace chocolatey.tests.integration.scenarios
     <tags>test admin</tags>
     <copyright></copyright>
     <licenseUrl>http://apache.org/2</licenseUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <releaseNotes></releaseNotes>
+  </metadata>
+</package>";
+
+        private const string NuspecContentWithVariables = @"<?xml version=""1.0"" encoding=""utf-8""?>
+<package xmlns=""http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd"">
+  <metadata>
+    <id>test-package</id>
+    <title>Test Package</title>
+    <version>$version$</version>
+    <authors>package author</authors>
+    <owners>package owner</owners>
+    <summary>A brief summary</summary>
+    <description>A big description</description>
+    <tags>test admin</tags>
+    <copyright></copyright>
+    <licenseUrl>https://github.com/chocolatey/choco/tree/$commitId$/LICENSE</licenseUrl>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
     <releaseNotes></releaseNotes>
   </metadata>
--- a/src/chocolatey/infrastructure.app/commands/ChocolateyPackCommand.cs
+++ b/src/chocolatey/infrastructure.app/commands/ChocolateyPackCommand.cs
@@ -16,7 +16,9 @@
 
 namespace chocolatey.infrastructure.app.commands
 {
+    using System;
     using System.Collections.Generic;
+    using System.Linq;
     using attributes;
     using commandline;
     using configuration;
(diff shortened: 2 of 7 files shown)


src/chocolatey/infrastructure.app/commands/ChocolateyPackCommand.cs
```diff
@@ -104,5 +108,25 @@ public virtual bool may_require_admin_access()
...
+                if (pair.Length != 2)
+                {
+                    throw new ArgumentException(
+                        string.Format("Ill-formed name=value pair in token: {0}", token),
+                        "properties");
+                }
+
+                properties.Add(pair[0], pair[1].remove_surrounding_quotes());
+            }
+        }
```
ferventcoder: @heaths there is already a pattern used in ChocolateyNewCommand - let's keep this consistent with that if possible

ferventcoder: In other words, there is a function for handling parsing anything that is unparsed

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
```diff
@@ -497,6 +506,17 @@ public sealed class ApiKeyCommandConfiguration
     }
 
     [Serializable]
+    public sealed class PackCommandConfiguration
+    {
+        public PackCommandConfiguration()
+        {
+            Properties = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
```
ferventcoder: 👍

src/chocolatey.tests.integration/scenarios/PackScenarios.cs
```diff
@@ -115,6 +115,54 @@ public void sources_should_be_set_to_specified_output_directory()
...
+        {
+            private readonly string package_path = Path.Combine(Scenario.get_top_level(), "test-package.0.1.0.nupkg");
+
+            public override void Context()
+            {
+                base.Context();
+
+                Scenario.reset(Configuration);
+
+                Configuration.PackCommand.Properties.Add("version", "0.1.0");
```
ferventcoder: Strange that you added version - it is already a first class citizen - should `--version 1.2.3` override `version=1.2.3` if both are specified?

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -268,6 +268,17 @@ public void pack_run(ChocolateyConfiguration config)
                 properties["version"] = config.Version;
             }
 
+            // Add any other properties passed to the pack command overriding any present.
+            foreach (var property in config.PackCommand.Properties)
+            {
+                this.Log().Debug(() => "{0} property '{1}': {2}".format_with(
```
ferventcoder: Debug, Verbose, or Trace here? I think Debug is fine.

Any concerns about sensitive data being exposed here? I will need to check if we log these in new.

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
heaths: I'm not sure what to make of this comment. Please clarify.

src/chocolatey/infrastructure.app/commands/ChocolateyPackCommand.cs
heaths: I saw that, but not sure how this fits that pattern. These aren't unparsed. I have to split the string passed to `--properties`.

src/chocolatey/infrastructure.app/services/NugetService.cs
heaths: The properties passed to the command line are already dumped in Debug. Everything in the command configuration is. I figured 1) this helps me test the feature, and 2) might help someone figure out why some properties was overwritten. Using `version` above was more a means to an end to test this behavior (real-world scenario might be someone having built up the value passed to `--properties` and one value overwrites another - just like with MSBuild properties). But if you want `--version` to always win, I can certainly change that and modify the test to pass the same variable name twice.

src/chocolatey.tests.integration/scenarios/PackScenarios.cs
heaths: That might be best. Seems nuget does it that way as well.

src/chocolatey/infrastructure.app/commands/ChocolateyPackCommand.cs
ferventcoder: I see. For consistency with Chocolatey, I think these would just be passed in as `key=value key2=value2` and wouldn't need a `--properties` bit. It matches consistency with choco, even if it is not consistent with nuget. I need to think some on this as there are advantages to either direction.

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
ferventcoder: Thumbs up? Just means it looks good.

src/chocolatey/infrastructure.app/services/NugetService.cs
ferventcoder: With Chocolatey, Trace is the lowest level - verbose is a separate thing that lets you log to debug or any other level, but only on verbose logger. So you could log something at info level to verbose, and it wouldn't show up without adding the verbose switch.

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
heaths: I get that. Not used to see it in a PR (we generally use comments for either asking about code or asking for a change). I have our internal build defs set up to even require resolving all comments one way or another, so it just seemed a bit odd.

src/chocolatey/infrastructure.app/services/NugetService.cs
heaths: I see. I spent some time grokking the code base but focused more on areas I needed to update. So would you prefer `Trace` or `Debug`?

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
ferventcoder: Ah fair - we like to tell folks the things we like as well as what we see that we that has questions, comments, or needs changes. Some things come down to differences in preferences and don't require any changes, just comments.

src/chocolatey/infrastructure.app/commands/ChocolateyPackCommand.cs
heaths: Okay. I looked for prior art (I'm a fan of project codebase consistency - just ask my teammates when I review most of our PRs) but didn't think to look at `install`. I didn't see that in usage output, though. Obviously (I think) this isn't like `--params` which I'm assuming at face value just passes the string through to the main install program as-is. But looking at `ChocolateyInstallCommand.handle_additional_argument_parsing`, would all name/value pairs be passed as `--foo=1 --bar=baz`. I guess I could see that as an alternative but - at least for the current install command help - that doesn't seem as obvious. I could update the documentation for both `install` and `pack` if you'd rather go that way. I can queue up the other requested changes for now.

src/chocolatey/infrastructure.app/services/NugetService.cs
ferventcoder: The log file contains everything debug/verbose by default - it does not log trace anywhere unless specifically requested. Trace is new, and I have some specific fixes that make the above true in a branch not yet pushed back up, so I might have you rebase your changes once that is in. We need some more docs on the levels.

As for debug logging in other places, typically a parameter that has "password" in the name will cause the logging of the passed in parameters to stop logging that data.

Let's keep it as is for now

src/chocolatey/infrastructure.app/commands/ChocolateyPackCommand.cs
ferventcoder: New, look at ChocolateyNewCommand, not install. :)

ferventcoder: Now that I'm back at a computer, I can point you directly to where this is - https://github.com/chocolatey/choco/blob/237de6ae53b09e38daa0c21fc0da83bf216d3cb4/src/chocolatey/infrastructure.app/commands/ChocolateyNewCommand.cs#L86-L103

ferventcoder: Also note how this gets changed in the help menu section as well - https://github.com/chocolatey/choco/blob/237de6ae53b09e38daa0c21fc0da83bf216d3cb4/src/chocolatey/infrastructure.app/commands/ChocolateyNewCommand.cs#L129:

` [<property=value> <propertyN=valueN>]`

ferventcoder: We'd also likely add at least one example in the help menu section.

heaths: Okay. Maybe I'll break that out into a common function. Could be an extension method itself in perhaps `StringExtensions`?

heaths:
@ferventcoder the CLA is signed and all tests fixed (seems _test.bat_ doesn't run integration tests so I missed a break, but was able to reproduce the same command AppVeyor uses).

heaths:
I'd love to be able to add Chocolatey-specific properties to my _vswhere_ package but that will require this change so I can pass in a couple of variables in my build. Anything else you require for changes?

ferventcoder:
Howdy - thanks for the reminder - this looks good, and will go into 0.10.8 (barring any high priority fixes).

ferventcoder:
One note @heaths - the commits seem to be all about the same thing, so I can squash the commits down or you can. It will still have you as the author if I do it. Up to you.

ferventcoder:
Pulling this in now.

ferventcoder:
Merged into stable at https://github.com/chocolatey/choco/commit/82224f191ce84955b7fd270af786f2f04f3dfbf3. This will be included in 0.10.8, thank you very much for your contribution!

Issue #1364, opened 2017-07-18
Choco new should output a .nuspec in UTF-8 without a byte order mark


jnm2:
I'm totally on board with this:

> Do not save your *.nuspec files with a Byte Order Mark (BOM). A BOM is neither required nor recommended for UTF-8, because it can lead to several issues.

https://chocolatey.org/docs/create-packages#character-encoding

But choco new is not. It's currently creating UTF-8-BOM nuspecs.

<sub>Moved from https://github.com/chocolatey/chocolatey.org/issues/527</sub>

ferventcoder:
This one is open for up for grabs.

Ben-Shirley:
I would like to have a go if I may?

Ben-Shirley:
anyone have time to show me where im going wrong?

bcurran3:
I found a bunch of my nuspec files with BOM. I later made https://chocolatey.org/packages/choco-nuspec-checker to fix them automatically... and then got a little ADHD with it. :)

PR #1409, opened 2017-09-29
(GH-1364) - update template service to generate UTF8 files without BOM.

Benny1007:
Updated TemplateService.cs to use UTF8Encoding with no BOM when creating the template files in relation to chocolatey#1364

Not sure why other files were updated, this is my first ever pull request.


TheCakeIsNaOH:
@Benny1007 Are you still interested in completing this?

Benny1007:
i can't even remember this, is it still required?

AdmiringWorm:
@Benny1007 IMO, it is still a valid contribution as files are even being generated with BOM.
It looks like the PR would need to be cleaned up, though, and at least the PowerShell files you have changed to not generate with BOM needs to have BOM (otherwise PowerShell will read them as ASCII).

However, PR #2064 perhaps could supersede this one as it is also up to date.

ferventcoder:
I think we could go for #2064 to supersede this - there were some files in here that did not appear to need to be adjusted that needed to be pulled out. It also doesn't appear that the CLA was signed (CLAHub the older one is showing as not signed, but it could have been at the time this was submitted). Thoughts?

TheCakeIsNaOH:
@gep13 This should be able to be closed as #2064 has been merged

gep13:
@TheCakeIsNaOH makes sense to me.  Thanks for pointing this out!

Issue #842, opened 2016-06-30
Parsing "-version" results in a warning message (instead of "--version")


pascalberger:
### What You Are Seeing?

I'm installing a Chocolatey package from a private feed, versioned accordingly to SemVer:

```
choco install mypackage -Version 1.16.0-unstable0017 -Source http://myFeed
```

While installing works fine, 'choco.exe' outputs this message:

```
Parsing -Version resulted in exception:
 Cannot bundle unregistered option '-e'.
```
### What is Expected?

'choco.exe' should not output a message since the version is a valid version according to SemVer.
### How Did You Get This To Happen? (Steps to Reproduce)
- Create a package with the above version number
- Install it

ferventcoder:
use `--version` instead of `-version`. It's complaining about the "e" because it thinks it is a bundled option.

ferventcoder:
"e" in the word v***e***rsion. Like `-dvyf` means `--debug --verbose --confirm --force`

pascalberger:
OK. Thanks @ferventcoder. Feel free to close this issue if this is by design and the expected behavior.

ferventcoder:
It's something we should keep open. It does present an issue, even if it is just a warning that it is providing.

KostyaSha:
> use --version instead of -version.

http://chocolatey.github.io/usage.html `-version` ?

KostyaSha:
so chef and puppet following docs either it was changed in non-backward compatible manner ?

gep13:
@KostyaSha that is a VERY old site, and I don't think it has been updated in a LONG time.  Best to use this:

https://chocolatey.org/docs/commands-install

Out of interest, how did you end up on that site?

KostyaSha:
@gep13 google :D

KostyaSha:
And why it old? It's your github.io page with documentation, it under your power.

gep13:
>@KostyaSha said...
And why it old? It's your github.io page with documentation, it under your power.

Oh it is, it is just that we moved away from using it.  I suspect we should probably get rid of it to avoid any further confusions :smile: @ferventcoder ?

ferventcoder:
@KostyaSha It's just a warning, not an exception. So not an incompatible change.

ferventcoder:
@KostyaSha I just updated the title to provide clarity and why it is low priority.

KostyaSha:
So, does chocolatey accept specified version or ends with warning print?

ferventcoder:
[...]

PR #1412, opened 2017-10-01
(GH-842) Detect wrong use of -version

initram:
Do not try to parse bundled options if the user most likely tried to use the --version option.

#842

### DIFF
--- a/src/chocolatey/infrastructure/commandline/Options.cs
+++ b/src/chocolatey/infrastructure/commandline/Options.cs
@@ -871,6 +871,8 @@ namespace chocolatey.infrastructure.commandline
             IDictionary<Option,string> normalOptions = new Dictionary<Option, string>();
 			if (f != "-")
 				return false;
+			if (n == "version")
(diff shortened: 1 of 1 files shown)


ferventcoder:
How about detecting "version" and handling it or ignoring it instead of throwing an error?

initram:
I was basing my work on the suggestion of the original case. Also it leads the user to the correct usage instead of allowing (an undocumented feature) or ignoring the users request. But I will gladly change it to whatever you prefer, I do not have much experience with choco. So I am unsure of how it has evolved (if using -version is an old style usage or something like that).

The error is caught by the way and presented to the user as a helpful message. If there is a better way of doing this, then please let me know.

ferventcoder:
> -version is an old style usage or something like that

That's exactly it.

initram:
So what are you suggesting that I change? 
Should I just close this pull request?

ferventcoder:
Suggesting maybe detect `-version` and delay looking it over.

PR #1436, opened 2017-10-29
(GH-1060) Add BeforeInstall parameter to Install-ChocolateyPackage.ps1

### LINKED ISSUE #1060: Add -BeforeInstall ScriptBlock for Install-ChocolateyPackage
I would like to execute some logic between the download and the install.. eg. around here: https://github.com/chocolatey/choco/blob/stable/src/chocolatey.resources/helpers/functions/Install-ChocolateyPackage.ps1#L313

If there was a new optional argument for Install-ChocolateyPackage that was of type [ScriptBlock] that was executed just before the call to Install-ChocolateyInstallPackage, that would do the job.

My particular scenario is needing to configure some firewall rules (to avoid Windows Firewall displaying a dialog about an application requiring access). I'm currently doing this before the call to Install-ChocolateyPackage, but ideally it would be done as late as possible, but before the actual install takes place, such that if there was a problem with the download then the firewall would not have been set.

flcdrg:
- Add a ScriptBlock parameter to Install-ChocolateyPackage.ps1 that if present, is called after download but before install.

Closes #1060

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Install-ChocolateyPackage.ps1
+++ b/src/chocolatey.resources/helpers/functions/Install-ChocolateyPackage.ps1
@@ -220,6 +220,13 @@ functionality (see links).
 .PARAMETER IgnoredArguments
 Allows splatting with arguments that do not apply. Do not use directly.
 
+.PARAMETER BeforeInstall Script
+Specifies the commands to run after download has completed but before install steps have begun.
+Available in 0.10.16+.
+
+Use this for starting an auxilary process such as AutoHotkey, so that any timeouts are not 
+affected by the time to download.
+
 .EXAMPLE
 >
 $packageName= 'bob'
@@ -359,6 +366,7 @@ param(
   [parameter(Mandatory=$false)]
   [alias("useOnlyPackageSilentArgs")][switch] $useOnlyPackageSilentArguments = $false,
   [parameter(Mandatory=$false)][switch]$useOriginalLocation,
+  [parameter(Mandatory=$false)][scriptblock] $beforeInstall,
   [parameter(ValueFromRemainingArguments = $true)][Object[]] $ignoredArguments
 )
   [string]$silentArgs = $silentArgs -join ' '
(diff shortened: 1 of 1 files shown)


Mistuke:
Would it also be possible to have a `BeforeUninstall`? I have a package that I haven't been able to upload to Chocolatey before of the way it performs uninstalls. It seems to first move a package and then run the uninstall script. But the move would obviously fail if a file is locked, such as a service running.

A `BeforeUninstall` would allow me to stop and unregister the service properly before chocolatey does it's thing.

ferventcoder:
@Mistuke that already exists, and for exactly that purpose - it's called the `chocolateyBeforeModify.ps1`. It needs to be in the already installed package to be called. It is run before upgrade and before uninstall. Run `choco new test` from the latest version of Chocolatey (v0.10.8 as of this message) and take a look at the output.

By the way, Chocolatey is actually quite good at handling locked files in the package directory, we even have tests on this. You may see some warnings and error messages but it does successfully perform upgrades with locked package files.

Mistuke:
@ferventcoder Thanks! I didn't know this was added, I should update my chocolatey. Last time I tried it with locked files it seems to just hang on the file move. But this was a while ago. I'll try it again.

gep13 (approved):
LGTM!

gep13:
@flcdrg thank you very much for taking the time to make this addition.  This looks good to me, and once the CI builds are complete, I will get this merged in.

src/chocolatey.resources/helpers/functions/Install-ChocolateyPackage.ps1
```diff
@@ -220,6 +220,13 @@ functionality (see links).
 .PARAMETER IgnoredArguments
 Allows splatting with arguments that do not apply. Do not use directly.
 
+.PARAMETER BeforeInstall Script
+Specifies the commands to run after download has completed but before install steps have begun.
+Available in 0.10.16+.
+
+Use this for starting an auxilary process such as AutoHotkey, so that any timeouts are not 
+affected by the time to download.
+
```
ferventcoder, line 229: Just a note for consistency - we put parameter docs in the order they appear. This should later be moved ABOVE IgnoredArguments.

PR #1497, opened 2018-02-19
(GH-1496) Enable ability to create a trimmed down version of Chocolatey.Lib

### LINKED ISSUE #1496: Need ability to have a LibOnly version of Chocolatey.Lib
Currently, the Chocolatey.Lib that is generated as part of the overall Chocolatey build, generates an assembly that contains everything that is needed to run choco.exe.  In some cases, all the contents of this assembly are not required, for example when only running one or two of the choco commands.  In these circumstances, it makes sense to make a trimmed down version of Chocolatey.Lib that only contains the bare minimum of requirements to function.

Obviously, this will mean that this trimmed down version of Chocolatey.Lib should only be used in a few special cases, as it will not have all the functionality of the main Chocolatey.Lib.

gep13:
This includes:

* New Compilation Directive, `LibOnly` to control when certain pieces of code are compiled
* Conditional inclusion of referenced projects
* Modification to build process to only ilmerge certain projects, based on configuration

Fixes #1496

### DIFF
--- a/src/chocolatey/GetChocolatey.cs
+++ b/src/chocolatey/GetChocolatey.cs
@@ -1,13 +1,13 @@
 ﻿// Copyright © 2017 Chocolatey Software, Inc
 // Copyright © 2011 - 2017 RealDimensions Software, LLC
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
-// 
+//
 // You may obtain a copy of the License at
-// 
+//
 // 	http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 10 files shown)


src/chocolatey/GetChocolatey.cs
```diff
@@ -500,7 +502,9 @@ private void extract_resources()
 
             try
             {
+#if !LibOnly
                 AssemblyFileExtractor.extract_all_resources_to_relative_directory(_container.GetInstance<IFileSystem>(), Assembly.GetAssembly(typeof(ChocolateyResourcesAssembly)), ApplicationParameters.InstallLocation, folders, ApplicationParameters.ChocolateyFileResources);
+#endif
```
ferventcoder: Would it be better to surround the entire try/catch with `#if !LibOnly`? I'm not sure how it likes an empty try with a filled out catch.

gep13:
@ferventcoder fixed up as requested.

ferventcoder:
Looks like this will need rebased.

gep13:
@ferventcoder rebased 😄

gep13:
Okay, looks like I messed up something with the merge...  Give me a couple minutes...

gep13:
@ferventcoder ah, there we go, all better 😸

PR #1507, opened 2018-02-22
(GH-1401) Added initial call to Chocolatey after installation

### LINKED ISSUE #1401: installation/setup: run choco once to initialize the config file
### What You Are Seeing?

First time after installing Chocolatey as admin, I start a non-admin command prompt, run `choco pack` and see this error:
image

### What is Expected?
It should work fine (whatever Chocolatey wants to write to ProgramData needs to be written at install time). Alternatively, consider saving the config file to LocalAppData (this doesn't need admin permissions to write, and is generally a better thing to do. Program Data goes to one place, configs, logs etc. go to another).

### Output Log
Unfortunately I didn't grab the output log other than the screenshot above. I tried a workaround (which is to run pack under admin, and that created the config file and I got unblocked).

gep13:
Fixes #1401

### DIFF
--- a/nuget/chocolatey/tools/chocolateysetup.psm1
+++ b/nuget/chocolatey/tools/chocolateysetup.psm1
@@ -126,6 +126,7 @@ Creating Chocolatey folders if they do not already exist.
 
   Add-ChocolateyProfile
   Install-DotNet4IfMissing
+  Invoke-Chocolatey-Initial
   if ($env:ChocolateyExitCode -eq $null -or $env:ChocolateyExitCode -eq '') {
     $env:ChocolateyExitCode = 0
   }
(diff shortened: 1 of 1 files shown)


nuget/chocolatey/tools/chocolateysetup.psm1
```diff
@@ -700,4 +701,17 @@ param(
   }
 }
 
+function Run-Chocolatey {
+	Write-Debug "Running Chocolatey..."
```
ferventcoder: Do we call write-debug elsewhere in here or does it have another name?

ferventcoder: Found one at https://github.com/chocolatey/choco/pull/1507/files#diff-98d6abdebfea2b377e7a6d36d36c4cfaR591

nuget/chocolatey/tools/chocolateysetup.psm1
```diff
@@ -700,4 +701,17 @@ param(
...
 }
 
+function Run-Chocolatey {
+	Write-Debug "Running Chocolatey..."
+
+	try {
+		$chocoInstallationFolder = Get-ChocolateyInstallFolder
+		$chocoExe = Join-Path -Path $chocoInstallationFolder -ChildPath "choco.exe"
+    & $chocoExe | Out-Null
+    Write-Debug "Chocolatey execution completed successfully."
```
ferventcoder: tabs versus spaces. Also ensure all line endings are CRLF

nuget/chocolatey/tools/chocolateysetup.psm1
gep13: Yeah, I borrowed `Write-Debug` from elsewhere 😄

nuget/chocolatey/tools/chocolateysetup.psm1
```diff
@@ -700,4 +701,17 @@ param(
   }
 }
 
+function Invoke-Chocolatey {
+  Write-Debug "Running Chocolatey..."
```
ferventcoder: `Initializing Chocolatey files, etc by running Chocolatey...`

Issue #63, opened 2015-02-04
When using choco push, provide ability to use alias


gep13:
Use Case:

As a user, I want to be able to use the alias name for a source when using `choco push`

i.e. `choco push -source chocolatey` is the same as `choco push -source http://chocolatey.org/`

ferventcoder:
Related to https://github.com/chocolatey/choco/issues/356

dragon788:
Definitely would love to see this one, especially since using the -s"http://somelongurl" is potentially error prone, since adding a trailing slash (or not) seems to cause it to not match if it isn't exactly the same as seen in the sources list even if it's the same url. Mostly biting me since the push url doesn't seem to want the /chocolatey or /nuget trailing text.

ferventcoder:
It's already done.

ferventcoder:
Oh wait, this is for push. NM

BaconTentacles:
To be honest, I am a little surprised this wasn't base functionality all along.  Given that choco behaves similarly to the nuget CLI, I would have expected "choco push -source MySourceName" to Just Work.  

I see this issue is also rather ancient.  Are there any plans to ever implement this?

gep13:
@vexx32 can you confirm whether this issue has been implemented or not?  Based on a PR that you have merged, there is a commit that is tagged against this issue.  If it has been addressed, then we should get this added to the milestone.

vexx32:
@gep13 yes, this was pulled in alongside other changes to choco push.

gep13:
While it was originally believed that this issue was solved, it turns out that it wasn't.  Let me try to explain...

In order to push to a source, you need to have an API Key defined for that source.  Within the chocolatey.config file, you will then have an entry that looks like this:

```
  <apiKeys>
    <apiKeys source="https://push.chocolatey.org" key="not_a_real_key />
  </apiKeys>
```

Similarly, when accessing a source for install/upgrade/download of packages, you may need to require authentication, and as such, you might have an entry in the chocolatey.config file, it may look something like this:

```
<source id="chocolatey" value="https://community.chocolatey.org/api/v2/" disabled="false" bypassProxy="false" selfService="false" adminOnly="false" priority="0" />
```

[...]

PR #1511, opened 2018-02-23
(#63) Support passing sources by name to push command.

DamianMaslanka5:
Related to #63

### DIFF
--- a/src/chocolatey.tests/infrastructure.app/commands/ChocolateyPushCommandSpecs.cs
+++ b/src/chocolatey.tests/infrastructure.app/commands/ChocolateyPushCommandSpecs.cs
@@ -1,13 +1,13 @@
 ﻿// Copyright © 2017 - 2021 Chocolatey Software, Inc
 // Copyright © 2011 - 2017 RealDimensions Software, LLC
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
-// 
+//
 // You may obtain a copy of the License at
-// 
+//
 // 	http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/src/chocolatey.tests/infrastructure.app/commands/ChocolateyPushCommandSpecs.cs
+++ b/src/chocolatey.tests/infrastructure.app/commands/ChocolateyPushCommandSpecs.cs
@@ -26,6 +26,7 @@ namespace chocolatey.tests.infrastructure.app.commands
     using chocolatey.infrastructure.app.services;
     using chocolatey.infrastructure.commandline;
     using Moq;
+    using NUnit.Framework;
     using Should;
 
     public class ChocolateyPushCommandSpecs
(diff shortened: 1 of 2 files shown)


DamianMaslanka5:
@gep13 Updated.

DamianMaslanka5:
@gep13 Can you take a look a this PR again?
Are you going to merge the changes or should I close the PR?

gep13:
@DamianMaslanka5 sorry for not getting back to you sooner about this!  I have started working on going through the backlog of PR's into this repository.  Are you in a position to sign the CLA for the project?  You can find a link in the comment above.  That will allow us to move forward with this change.

DamianMaslanka5:
@gep13 I signed the CLA

gep13:
@DamianMaslanka5 I am coming back to look at this again, and the more I think about this, the more that I think that the current approach used in this PR is not the right one.  Let me try to explain...

Out of the box, Chocolatey ships with the following sources configured:

```
  <sources>
    <source id="chocolatey" value="https://community.chocolatey.org/api/v2/" disabled="false" bypassProxy="false" selfService="false" adminOnly="false" priority="0" />
  </sources>
```

So being able to do something like:

```
choco install windirstat --source="'chocolatey'"
choco install windirstat --source="'https://community.chocolatey.org/api/v2/'"
```

makes a lot of sense, since both of these pieces of information exist in the source element of the chocolatey.config file.

However, when it comes to pushing packages, the sources element in the chocolatey.config file isn't actually used, it is only the entries in the apikey section that are inspected.  By default, there aren't any entries in there, but when populated, it would look something like the following:

```
<apiKeys source="https://push.chocolatey.org/" key="encrypted_stuff" />
```

Now, while there is a source attribute here, there isn't an id attribute as there was before.

In order for this PR to work, there would need to be an additional entry in the source element, so that the API Key could be found, i.e.

```
  <sources>
    <source id="chocolatey" value="https://community.chocolatey.org/api/v2/" disabled="false" bypassProxy="false" selfService="false" adminOnly="false" priority="0" />
    <source id="chocolatey_push" value="https://push.chocolatey.org/" disabled="false" bypassProxy="false" selfService="false" adminOnly="false" priority="0" />
  </sources>
```

But this would lead to duplicated sources being shown when they aren't actually required.

I think the correct approach here would be to add a new attribute onto the apikey elements.

What do you think?

@TheCakeIsNaOH what are your thoughts here, as this likely impacts on your follow up PR for defining a default named push source for Chocolatey commands.

If I am missing something here, then please feel free to let me know.

gep13:
Or, thinking about this some more, after a chat with @pauby...

Do we want/need an additional attribute on the source element, which is the push URL for the source, i.e. the source element for Chocolatey would become:

```
<sources>
    <source id="chocolatey" value="https://community.chocolatey.org/api/v2/" disabled="false" bypassProxy="false" selfService="false" adminOnly="false" priority="0" pushUrl="https://push.chocolatey.org/" />
  </sources>
```

And then this is used to locate the API Key associated with the source when someone does:

```
choco push temp.nupkg --source="'chocolatey'"
```

In the absence of anything defined in the pushUrl, it could fallback to attempting to use the URL in the `value` attribute, which may or may not work.

Thoughts?

TheCakeIsNaOH:
> @TheCakeIsNaOH what are your thoughts here, as this likely impacts on your follow up PR for defining a default named push source for Chocolatey commands.

This only should affect #2430 insofar as it requires I change tests.

The configuration value would still go through the same parsing and whatnot that a source passed in via `--source` would, AFAIK, so it should not matter how it is implemented.

> Do we want/need an additional attribute on the source element, which is the push URL for the source, i.e. the source element for Chocolatey would become:

I'm not really a fan of that. If we are adding elements, IMHO it should go to the apikeys.

> I think the correct approach here would be to add a new attribute onto the apikey elements.

Yes, something like:
`<apiKeys id="chocolatey" source="https://push.chocolatey.org/" key="encrypted_stuff" />`

TheCakeIsNaOH:
Oh, and I could remove the dependency on this from #2430 if desired, then this PR could be updated to add those tests to ensure that a name works with `DefaultPushSource`. That would allow #2430 to be merged.

vexx32:
Apologies @DamianMaslanka5 I hadn't realised #2430 was built off this PR already.

That PR is now merged, so relevant to the discussion above @gep13 I think that's a worthwhile enhancement, but it doesn't necessarily require this PR (or @TheCakeIsNaOH's) to include it IMO. I think that's worth its own separate follow up enhancement issue. I will open a follow up issue to address that.

With that, because the bulk of this PR has been incorporated by merging #2430, I'm going to go ahead and close this one off for now. Thank you for your work here @DamianMaslanka5! 💖

PR #1512, opened 2018-02-26
(GH-1365) Support architecture folder in tools folder

### LINKED ISSUE #1365: Ability to install from /tools/x64 and /tools/x86 without PowerShell
In the NuGet world it's a common convention to put platform-specific things in platform-specific folders, both for native resources and managed resources.

I think it would be cool if Chocolatey adopted the convention of ignoring `/tools/x64` unless you're doing a 64-bit install, in which case it treats the contents of `/tools/x64` as though they were in `/tools`. Likewise for `/tools/x86` and 32-bit installs.

This would remove the need for PowerShell and I could simply distribute:
```
/tools/x64/my-exe-to-shim.exe
/tools/x86/my-exe-to-shim.exe
```

<sub>Moved from https://github.com/chocolatey/chocolatey.org/issues/526</sub>

gep13:
This allows a Chocolatey Package structure similar to the following:

image

Which simply contain:

image

On installation, a `.ignore` file will be created for the alternative architecture (relative to what is being installed):

image

And a shim will be created pointing at the correct architecture:

image

Fixes #1365

### DIFF
--- a/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
+++ b/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
@@ -381,6 +381,9 @@ Did you know Pro / Business automatically syncs with Programs and
                 //review: is this a Windows only kind of thing?
                 pkgInfo.FilesSnapshot = _filesService.capture_package_files(packageResult, config);
 
+                var is32Bit = !config.Information.Is64BitProcess || config.ForceX86;
+                create_ignore_files_for_executables(packageResult.InstallLocation, !is32Bit);
+
                 if (packageResult.Success) _shimgenService.install(config, packageResult);
             }
             else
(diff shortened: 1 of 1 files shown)


src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
```diff
@@ -451,6 +454,19 @@ public void handle_package_result(PackageResult packageResult, ChocolateyConfigu
...
+        {
+            // If we are using a 64 bit architecure, we want to ignore exe's targetting x86
+            // This is done by adding a .ignore file into the package folder for each exe to ignore
+            var searchLocation = is64Bit ? "tools\\x86" : "tools\\x64";
+            var exeFiles = _fileSystem.get_files(_fileSystem.combine_paths(installLocation, searchLocation), pattern: "*.exe", option: SearchOption.AllDirectories);
+
+            foreach (var exeFile in exeFiles)
+            {
+                _fileSystem.create_file(exeFile + ".ignore");
+            }
```
ferventcoder: If there are only 32bit files, what is the expectation here? I think it should probably go ahead and use those files.

gep13: @ferventcoder good catch.  I have updated this PR with a slight modification.

ferventcoder: Somehow your whitespace changes also snuck in.

src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
```diff
@@ -446,11 +449,31 @@ public void handle_package_result(PackageResult packageResult, ChocolateyConfigu
...
+        private void create_ignore_files_for_executables(string installLocation, bool is64Bit)
+        {
+            // If we are using a 64 bit architecure, we want to ignore exe's targetting x86
+            // This is done by adding a .ignore file into the package folder for each exe to ignore
+            var exeFiles32Bit = _fileSystem.get_files(_fileSystem.combine_paths(installLocation, "tools\\x86"), pattern: "*.exe", option: SearchOption.AllDirectories).ToArray();
+            var exeFiles64Bit = _fileSystem.get_files(_fileSystem.combine_paths(installLocation, "tools\\x64"), pattern: "*.exe", option: SearchOption.AllDirectories).ToArray();
+
+            // If 64bit, and there are only 32bit files, we should shim the 32bit versions,
+            // therefore, don't ignore anything
+            if (!exeFiles64Bit.Any() && exeFiles32Bit.Any() && is64Bit)
```
ferventcoder: Micro-optimization: You should check `is64bit` first to short-circuit, and then you only need to check if there are no 64bit files prior to returning.

ferventcoder: Let me restate the second part of that. You only need to check if there are no 64bit files prior to returning when 64bit.

src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
```diff
@@ -446,11 +449,31 @@ public void handle_package_result(PackageResult packageResult, ChocolateyConfigu
...
+            {
+                return;
+            }
+
+            foreach (var exeFile in is64Bit ? exeFiles32Bit : exeFiles64Bit)
```
ferventcoder: Add `.or_empty_list_if_null()` to the end of those lists (and be sure that is what it is called, I'm using pseudocode).

gep13:
@ferventcoder back to you again 😄

PR #1583, opened 2018-05-29
(GH-1580) Only output log messages when RegularOutput is enabled

### LINKED ISSUE #1580: Logging - warnings for ".registry.bad." files are emitted with "-r" switch
## warnings for corrupted registry files are displayed with the `-r` switch

on each choco.exe call following message is getting emmited (twice)
```
A corrupt .registry file exists at C:\ProgramData\chocolatey\.chocolatey\accurev-client.5.7.1.170613\.registry.bad.
 Open this file in a text editor, and remove/escape any characters that
 are regarded as illegal within XML strings not surrounded by CData.
 These are typically the characters &, `<`, and `>`. Again, this
 is an XML document, so you will see many < and > characters, so just
 focus exclusively in the string values not surrounded by CData. Once
 these have been corrected, rename the .registry.bad file to .registry.
 Once saved, try running the same Chocolatey command that was just
 executed, so verify problem is fixed.
 NOTE: It will not be possible to rename the file in Windows Explorer.
 Instead, you can use the following PowerShell command:
 Move-Item .\.registry.bad .\.registry
```
the choco.exe parameter `-r` should not print warnings:
```
 -r, --limitoutput, --limit-output
     LimitOutput - Limit the output to essential information
```

gep13:
Fixes #1580

### DIFF
--- a/src/chocolatey/infrastructure.app/services/ChocolateyPackageInformationService.cs
+++ b/src/chocolatey/infrastructure.app/services/ChocolateyPackageInformationService.cs
@@ -19,8 +19,10 @@ namespace chocolatey.infrastructure.app.services
     using System;
     using System.IO;
     using System.Text;
+    using configuration;
     using NuGet;
     using domain;
+    using infrastructure.configuration;
     using tolerance;
     using IFileSystem = filesystem.IFileSystem;
 
@@ -29,6 +31,7 @@ namespace chocolatey.infrastructure.app.services
         private readonly IFileSystem _fileSystem;
         private readonly IRegistryService _registryService;
         private readonly IFilesService _filesService;
+        private readonly ChocolateyConfiguration _config;
         private const string REGISTRY_SNAPSHOT_FILE = ".registry";
         private const string REGISTRY_SNAPSHOT_BAD_FILE = ".registry.bad";
         private const string FILES_SNAPSHOT_FILE = ".files";
@@ -44,6 +47,15 @@ namespace chocolatey.infrastructure.app.services
             _fileSystem = fileSystem;
             _registryService = registryService;
             _filesService = filesService;
+            _config = Config.get_configuration_settings();
+        }
+
+        public ChocolateyPackageInformationService(IFileSystem fileSystem, IRegistryService registryService, IFilesService filesService, ChocolateyConfiguration config)
+        {
+            _fileSystem = fileSystem;
+            _registryService = registryService;
+            _filesService = filesService;
+            _config = config;
         }
 
         public ChocolateyPackageInformation get_package_information(IPackage package)
@@ -51,7 +63,7 @@ namespace chocolatey.infrastructure.app.services
             var packageInformation = new ChocolateyPackageInformation(package);
             if (package == null)
             {
-                this.Log().Debug("No package information as package is null.");
+                if (_config.RegularOutput) { this.Log().Debug("No package information as package is null."); }
                 return packageInformation;
             }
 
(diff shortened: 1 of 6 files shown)


src/chocolatey/infrastructure.app/services/ChocolateyPackageInformationService.cs
```diff
@@ -46,12 +47,12 @@ public ChocolateyPackageInformationService(IFileSystem fileSystem, IRegistryServ
             _filesService = filesService;
         }
 
-        public ChocolateyPackageInformation get_package_information(IPackage package)
+        public ChocolateyPackageInformation get_package_information(IPackage package, ChocolateyConfiguration config)
```
ferventcoder: This needs an overload

src/chocolatey/infrastructure.app/services/IChocolateyPackageInformationService.cs
```diff
@@ -16,13 +16,14 @@
...
     using NuGet;
 
     public interface IChocolateyPackageInformationService
     {
-        ChocolateyPackageInformation get_package_information(IPackage package);
-        void save_package_information(ChocolateyPackageInformation packageInformation);
-        void remove_package_information(IPackage package);
+        ChocolateyPackageInformation get_package_information(IPackage package, ChocolateyConfiguration config);
+        void save_package_information(ChocolateyPackageInformation packageInformation, ChocolateyConfiguration config);
+        void remove_package_information(IPackage package, ChocolateyConfiguration config);
```
ferventcoder: All of these need an overload

gep13:
@ferventcoder I have updated based on your comments, and discussion last night.

Let me know if there is additional work required here.  I simply did `new ChocolateyConfiguration` in the overload, wasn't sure if anything needed to be explicitly set here or not.

src/chocolatey/infrastructure.app/services/ChocolateyPackageInformationService.cs
```diff
@@ -47,11 +48,16 @@ public ChocolateyPackageInformationService(IFileSystem fileSystem, IRegistryServ
         }
 
         public ChocolateyPackageInformation get_package_information(IPackage package)
+        {
+            return get_package_information(package, new ChocolateyConfiguration());
```
ferventcoder: This is probably fine, but I would be a bit more complete by adding the config into the service itself.

ferventcoder: That way you don't need to pass config around to the functions.

src/chocolatey/infrastructure.app/services/IChocolateyPackageInformationService.cs
ferventcoder: won't need overloads if we don't pass config.

gep13:
@ferventcoder well, that simplified the change quite considerably!

I have left the original whitespace corrections for the files that were originally changed in this PR.  Let me know if you want to back those out as well.

src/chocolatey/infrastructure.app/services/ChocolateyPackageInformationService.cs
```diff
@@ -44,14 +47,23 @@ public ChocolateyPackageInformationService(IFileSystem fileSystem, IRegistryServ
             _fileSystem = fileSystem;
             _registryService = registryService;
             _filesService = filesService;
+            _config = Config.get_configuration_settings();
```
ferventcoder: 👍

PR #1587, opened 2018-06-01
(GH-1584) Fixed null reference exception in NugetCommon when source is null

### LINKED ISSUE #1584: Object reference exception when there are no sources
Steps to reproduce on 0.10.11:
* configure default chocolatey source if not present
* install git package (package is not relevant, applies to any package)
* remove all package sources
* call ```choco uninstall git -y```

Actual result:
Following error message is returned: ```Object reference not set to an instance of an object.```

Expected result:
No message is shown and package is successfully uninstalled, since you dont need package sources for uninstall.

Note:
* The same issue applies also for other commands (but i didnt note which one was the second).
* Exception is thrown from NugetCommon.GetRemoteRepository line 80, where split method fails null sources property.

jirkapok:
* Added new unit test for NugetCommon
* Fixed possible null reference exception when parsing sources from chocolateyconfig

Closes #1584


jirkapok:
Codacy issues are false positive, i copied structure from other tests.

ferventcoder:
This has been rebased and merged into stable at https://github.com/chocolatey/choco/commit/7deb88e05bc22773a07096e88c9408ff1c8cd91f. Thanks!

Issue #1477, opened 2018-01-24
new - allow absolute or relative paths for templates


Pilskalns:
Currently, Chocolatey expects package templates to be placed in `C:\ProgramData\chocolatey\templates\<templateName>`

It would be very practical to allow also use of relative or absolute paths. Why? It's easier to git-version package template together in the same location with code.

Also, to run code editor on file inside `ProgramData`  needs editor to be launched in elevated mode.

P.S. Currently possible hack/work-around is to use `choco new "$package" -t "..\..\..\Path\From\Root\<templateName>"` but it doesn't allow jump to another drive.

Pilskalns:
Same day later I figured better workaround...

``` powershell
# UPDATE TEMPLATE TO CHOCO WORK DIRECTORY
New-Item C:\ProgramData\chocolatey\templates -type directory -force | out-null
if(Test-Path C:\ProgramData\chocolatey\templates\$template){
	Remove-Item C:\ProgramData\chocolatey\templates\$template -recurse
}
Copy-Item setup\chocotemplate C:\ProgramData\chocolatey\templates\$template -recurse
```
Gets executed from build script inside git directory

ferventcoder:
Howdy! I meant to comment when I saw this come across my email but this sounds like a great idea! I've set it up for "Up For Grabs"

david-proctor:
Hey, long time user, first time contributor here. Can I pick up this issue?

ferventcoder:
There is a related issue at #1637, I think we'll want to close the other one as a duplicate as this already provides a PR.

pauby:
The PR for this issue has had no response in some time and has been closed, so I'm going to go ahead and close this issue too.

PR #1620, opened 2018-07-30
(#1477) Allow relative paths for templates

david-proctor:
Previously, a `choco new` command expected templates to be located in
c:\ProgramData\chocolatey\templates\. This change allows running
`choco new` with a template path relative to the current working
directory.

This pull request relates to issue GH-1477. See discussion at https://github.com/chocolatey/choco/issues/1477.

### DIFF
--- a/src/chocolatey/infrastructure.app/services/TemplateService.cs
+++ b/src/chocolatey/infrastructure.app/services/TemplateService.cs
@@ -138,8 +138,7 @@ namespace chocolatey.infrastructure.app.services
             else
             {
                 configuration.NewCommand.TemplateName = string.IsNullOrWhiteSpace(configuration.NewCommand.TemplateName) ? "default" : configuration.NewCommand.TemplateName;
-
-                var templatePath = _fileSystem.combine_paths(ApplicationParameters.TemplatesLocation, configuration.NewCommand.TemplateName);
+                var templatePath = parse_template_path(configuration);
                 if (!_fileSystem.directory_exists(templatePath)) throw new ApplicationException("Unable to find path to requested template '{0}'. Path should be '{1}'".format_with(configuration.NewCommand.TemplateName, templatePath));
 
                 this.Log().Info(configuration.QuietOutput ? logger : ChocolateyLoggers.Important, "Generating package from custom template at '{0}'.".format_with(templatePath));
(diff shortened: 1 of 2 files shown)


gep13:
@david-proctor I am starting to go through the backlog of PR's into this repository, apologies for not getting back to you sooner about this.  Are you in a position to sign the CLA for the project?  You can find it linked in the previous comment.  Thanks

david-proctor:
CLA signed.

src/chocolatey/infrastructure.app/services/TemplateService.cs
```diff
@@ -188,5 +187,21 @@ public void generate_file_from_template(ChocolateyConfiguration configuration, T
             _fileSystem.create_directory_if_not_exists(_fileSystem.get_directory_name(fileLocation));
             _fileSystem.write_file(fileLocation, template, encoding);
         }
+
+        private string parse_template_path(ChocolateyConfiguration configuration)
+        {
+            if (!configuration.NewCommand.TemplateName.Equals("default"))
+            {
+                var rawRelativePath = _fileSystem.combine_paths(_fileSystem.get_current_directory(), configuration.NewCommand.TemplateName);
```
TheCakeIsNaOH, line 195: I don't think this is required, you should be able to call `_filesystem.get_full_path()` directly with the `TemplateName`. That would allow absolute paths to work as well as relative paths.

src/chocolatey/infrastructure.app/services/TemplateService.cs
```diff
@@ -188,5 +187,21 @@ public void generate_file_from_template(ChocolateyConfiguration configuration, T
...
         }
+
+        private string parse_template_path(ChocolateyConfiguration configuration)
+        {
+            if (!configuration.NewCommand.TemplateName.Equals("default"))
+            {
+                var rawRelativePath = _fileSystem.combine_paths(_fileSystem.get_current_directory(), configuration.NewCommand.TemplateName);
+                var fullPath = _fileSystem.get_full_path(rawRelativePath);
+
+                if (_fileSystem.directory_exists(fullPath))
```
TheCakeIsNaOH, line 198: So the order of checks here is:
1. Check if templatename is a path that exists
1. Check if templatename exists in the template directory
1. Throw an error

I'd personally switch the order of one and two, so the templatename is checked in the template directory first, and only if it does not exist is the templatename checked if it is a path that exists. 

Say if you have a template called msi that you use, but you also have a directory called msi. Then, as this is currently setup, if you try to use the msi template, it will actually be using the msi directory as the template, instead of the installed msi template.

PR #1624, opened 2018-08-08
(GH-1623) Change $host.Version to $PSVersionTable.PSVersion

MaxBurgert:
With the latest release of Powershell, this line throws an SetValueInvocationException: "The requested security protocol is not supported."
Change the SecurityProtocol to SystemDefault ensures backwards compatibility and is necessary since Ssl3 is deprecated.

Fixes GH-1623

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Get-ChocolateyWebFile.ps1
+++ b/src/chocolatey.resources/helpers/functions/Get-ChocolateyWebFile.ps1
@@ -305,7 +305,7 @@ param(
     try {
       $headers = Get-WebHeaders -Url $url -ErrorAction "Stop"
     } catch {
-      if ($host.Version -lt (New-Object 'Version' 3,0)) {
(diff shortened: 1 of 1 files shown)


chuanjiao10:
PS c:\> [System.Enum]::getvalues([Net.SecurityProtocolType])
SystemDefault
Ssl3
Tls
Tls11
Tls12

It is recommended to use tls12

MaxBurgert:
You're right, this was not the issue. 
The issue was with Visual Studio Code, which returned for `$host.Version` not the actual PowerShell version, whereby a comparison with `$PSVersionTable.PSVersion` will result in the expected behaviour. 
I've updated the title of the PR.
With cfdadc6, the comparison is done in the appropriate way as The Scripting Guys mentioned [here](https://blogs.technet.microsoft.com/heyscriptingguy/2014/10/13/powertip-check-version-of-powershell/).

Issue #1234, opened 2017-04-04
Inconsistent docs for the parameter "-WorkingDirectory" in Install-ChocolateyShortcut


OlafHaag:
### What You Are Seeing?
Inconsistent use for the parameter "-WorkingDirectory" in _Install-ChocolateyShortcut_ helper function documentation. In Example 2 it's "-WorkDirectory", in the chocolateyInstall.ps1 that comes with choco new <packagename> it's "-workDirectory".

### What is Expected?
That it's always called "-WorkingDirectory" in order to not confuse me and make errors.
### How Did You Get This To Happen? (Steps to Reproduce)
choco new, and looking at the docs: https://chocolatey.org/docs/helpers-install-chocolatey-shortcut, looking at the source code

ferventcoder:
Documentation bug. Up for grabs.

gep13:
This was merged into stable at https://github.com/chocolatey/choco/commit/26ace8054b2ba50cd3a4dabe086edfa62b79af26

gep13:
Merged into stable to be released with 0.10.12.

PR #1660, opened 2018-10-12
(doc) Fix example in Install-ChocolateyShortcut.ps1

OndrejVicar:
Fix one example which had wrong param - `WorkDirectory` -> `WorkingDirectory`.


ferventcoder:
#1234 was this issue for this. @OndrejVicar Thanks so much for the contribution! I think we missed this one due to not having been linked to an issue. We have started the process on pulling in #1730 instead.

PR #1719, opened 2019-01-29
(GH-1604) Fix DateTime formatting issue in Env:ChocolateyLastPathUpdate

### LINKED ISSUE #1604: ChocolateyLastPathUpdate stores date as locale-specific
### What You Are Seeing?

On my Windows 10 box, the `ChocolateyLastPathUpdate` environment variable's value contains Swedish, non-Ascii characters (such as `å` or `ö`) for certain timestamps because my date and currencies settings in Windows are set to Swedish. This has caused several third-party applications which read the environment to crash or malfunction.

Indeed, this problem lies more with the third-party applications than with choco. However, I'd argue it's bad form to store the date like this if this isn't necessary.

Example value when date and currency format is set to English: `Tue Apr 24 11:17:15 2018`.
Unfortunately, I don't have an existing example string in Swedish anymore, as I had to change this. But as an example, Saturday is "Lördag" in Swedish and abbreviated into `Lör`.

SneWs:
(GH-1604) Fix DateTime formatting issue in Env:ChocolateyLastPathUpdate

Due to the DateTime format used this would cause all kinds of issues when a tool or environment uses the environment variables. For example, a python script that reads all environment variables on a system with a Swedish DateTime culture set and tries to serialize the key/values as JSON would/could error out when a ÅÄ or Ö is written by choco into ChocolateyLastPathUpdate, it's also a hard bug to track down since in the Swedish case it would only trigger on a Saturday, Sunday or Monday where mentioned chars would appear. 

This change changes the formatting used to force it to a long in the form of using ToFileTime() on the returned Date object instead of the formatting flags used before the change. 

Closes #1604

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Set-EnvironmentVariable.ps1
+++ b/src/chocolatey.resources/helpers/functions/Set-EnvironmentVariable.ps1
@@ -118,7 +118,7 @@ public static extern IntPtr SendMessageTimeout(
 
     # Set a user environment variable making the system refresh
     $setx = "$($env:SystemRoot)\System32\setx.exe"
-    & "$setx" ChocolateyLastPathUpdate `"$(Get-Date -UFormat %c)`" | Out-Null
+    & "$setx" ChocolateyLastPathUpdate `"$((Get-Date).ToFileTime())`" | Out-Null
   } catch {
     Write-Warning "Failure attempting to let Explorer know about updated environment settings.`n  $($_.Exception.Message)"
   }


SneWs:
Ah, my bad. I did read through the contributing docs. I misunderstood it. I thought the PR should contain the detailed message, not the commit message. 

@ferventcoder If you could update the commit message for me that would be super. Or I can try to figure out how to do that when I get into the office tomorrow morning.

SneWs:
Amended the commit and updated the commit message to be the same as the PR message.

SneWs:
So. Seems to be very low activity here. Am I missing something else after the amend or will this be merged?

FurkanKambay:
@SneWs The doc also says the commit message body should wrap at 72 characters; that might be it. Also, you might have to wait a few weeks until someone can review again.

Gotta say, my OCD side loves all these rules but they really do seem like kinda overboard.

I think this should work:

```
(GH-1604) Fix DateTime formatting issue in Env:ChocolateyLastPathUpdate

Due to the DateTime format used this would cause all kinds of issues
when a tool or environment uses the environment variables. For example,
a python script that reads all environment variables on a system with a
Swedish DateTime culture set and tries to serialize the key/values as
JSON would/could error out when a ÅÄ or Ö is written by choco into
ChocolateyLastPathUpdate, it's also a hard bug to track down since in
the Swedish case it would only trigger on a Saturday, Sunday or Monday
where mentioned chars would appear.

This change changes the formatting used to force it to a long in the
form of using ToFileTime() on the returned Date object instead of the
formatting flags used before the change.
```

ferventcoder:
> edit: Just realized the doc also says the PR needs to be from a feature branch and yours is from master. Maybe that rule isn't that strict but thought I should mention it.

@FurkanKambay The master branch is not super strict, but once we accept he will need to force push to his master - not ideal. And he can't fetch and update his master. Another reason why a feature branch is best.

ferventcoder:
> Gotta say, my OCD side loves all these rules but they really do seem like kinda overboard.

@FurkanKambay If someone can't be bothered to follow along with strict guidance, what does that state about the quality of the work? How meticulous has someone been in preparing a contribution for something that is used by hundreds of thousands of people and many hundreds of organizations? It requires care and attention. It's a bit like the [brown M&Ms](https://www.fastcompany.com/1550881/business-advice-van-halen).

There is also a very specific reason - we tend to search back across commits and the reasoning behind those commits, long after everyone has forgotten why a specific line was put into the code. So we need more precise care in this codebase to enable that ability. Not every codebase has stringent needs, and not every codebase in our organization has these specific needs. HTH

ferventcoder:
@SneWs we don't typically forget about these things. The commit body does need a bit of an adjustment as @FurkanKambay noticed. I can make this for you when I merge it in unless you are ahead of me on this.

FurkanKambay:
@ferventcoder Of course; I didn't mean to complain. My comment was really just about the 72-char wrap rule (which I agree is pretty easy to follow; I just thought it could always be solved with tools like IDEs)
Otherwise, I absolutely support the other rules and I love that you guys have them. Thanks!

ferventcoder:
This has been merged into stable at https://github.com/chocolatey/choco/commit/7e1143a248b8c4e99a9edc85db579e1566f8c5b3. Thank you for your contribution!

PR #1741, opened 2019-03-01
(GH-1788) Use LiteralPath in Get-UninstallRegistryKey

### LINKED ISSUE #1788: Get-UninstallRegistryKey doesn't work with keys that have bracket characters(`[]`)
### What You Are Seeing?

Get-UninstallRegistryKey showing registry keys contains some characters(eg `[` or `]`) as bad keys because where `Get-ItemProperty -Path` wouldn't accept that in the path.

### What is Expected?

It should work even if the registry keys contain `[]` characters

Call `Get-UninstallRegistryKey $softwareName` when you have a $softwareName that represents your software which contains `[]`.

alFReD-NSH:
This fixes cases where registry keys contains some characters(eg `[` or `]`) where `Get-ItemProperty -Path` wouldn't accept.

I'm not sure the badKeys check is still required or not though, because I think this fixes the problem that it was trying to fix there.

Fixes #1788

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Get-UninstallRegistryKey.ps1
+++ b/src/chocolatey.resources/helpers/functions/Get-UninstallRegistryKey.ps1
@@ -134,13 +134,13 @@ param(
 
     $keyPaths = $keys | Select-Object -ExpandProperty PSPath
     try {
-      [array]$foundKey = Get-ItemProperty -Path $keyPaths -ErrorAction Stop | ? { $_.DisplayName -like $softwareName }
+      [array]$foundKey = Get-ItemProperty -LiteralPath $keyPaths -ErrorAction Stop | ? { $_.DisplayName -like $softwareName }
       $success = $true
     } catch {
       Write-Debug "Found bad key."
       foreach ($key in $keys){
         try {
-          Get-ItemProperty $key.PsPath > $null
+          Get-ItemProperty -LiteralPath $key.PsPath > $null
         } catch {
           $badKey = $key.PsPath
         }


src/chocolatey.resources/helpers/functions/Get-UninstallRegistryKey.ps1
```diff
@@ -134,10 +134,10 @@ param(
 
     $keyPaths = $keys | Select-Object -ExpandProperty PSPath
     try {
-      [array]$foundKey = Get-ItemProperty -Path $keyPaths -ErrorAction Stop | ? { $_.DisplayName -like $softwareName }
+      [array]$foundKey = Get-ItemProperty -LiteralPath $keyPaths -ErrorAction Stop | ? { $_.DisplayName -like $softwareName }
       $success = $true
     } catch {
-      Write-Debug "Found bad key."
+      Write-Debug -LiteralPath "Found bad key."
```
ferventcoder: I'm not sure this is necessary.

src/chocolatey.resources/helpers/functions/Get-UninstallRegistryKey.ps1
```diff
@@ -134,10 +134,10 @@ param(
 
     $keyPaths = $keys | Select-Object -ExpandProperty PSPath
     try {
-      [array]$foundKey = Get-ItemProperty -Path $keyPaths -ErrorAction Stop | ? { $_.DisplayName -like $softwareName }
+      [array]$foundKey = Get-ItemProperty -LiteralPath $keyPaths -ErrorAction Stop | ? { $_.DisplayName -like $softwareName }
```
ferventcoder: This is fine for me.

alFReD-NSH:
I have updated the PR with your comment fixed and created a GH issue

src/chocolatey.resources/helpers/functions/Get-UninstallRegistryKey.ps1
alFReD-NSH: Fixed now.

ferventcoder:
Thank you - last thing is to update the git commit itself. We can handle this as we pull it in if you prefer

ferventcoder:
The way the commit should look is defined at https://github.com/chocolatey/choco/blob/master/CONTRIBUTING.md#prepare-commits. The big thing is `(GH-1788) Use Literal Path w/Get-UninstallRegistryKey` being the summary and then the body. We don't put control comments into commits (like `Closes #1788` or `Fixes #1788`). HTH

ferventcoder:
In thinking about this, I'm hopeful we don't break anything here for folks. Since we are pulling these items out, we should feel comfortable these are literal paths.

ferventcoder:
Confirmed this works for PowerShell v2 as well  - http://adamringenberg.com/powershell2/get-itemproperty/

ferventcoder:
This has been merged to stable at https://github.com/chocolatey/choco/commit/c0e826798428112e01630cce3687ae0518092e68 and will be part of the Chocolatey v0.10.14 release. Thanks you for helping to make Chocolatey better!

Issue #1689, opened 2018-12-05
upgrade - remove automation scripts prior to upgrade even if changed


ferventcoder:
tl;dr - Package Reducer causes this behavior because of removing files from the nupkg to reduce the size. Let's fix it so that it no longer does this.

### What You Are Seeing?

An upgrade appears to be using an old chocolateyInstall.ps1 after a newer version removed the need for the script. This means the file would not have been overwritten, but should have been removed since it was changed.

### What is Expected?

Files that have not been changed from the older nupkg installation should be removed.

### How Did You Get This To Happen? (Steps to Reproduce)

Turn on Package Reducer (no more files in the older nupkg to compare to).

Install `transifex-client` version 0.12.4. Note it has a chocolateyInstall.ps1. Then upgrade to 0.13.2 (note it does not have a chocolateyInstall.ps1).

See it fail because it did not remove the file. 

### Output Log

<details>
<summary>Full Log Output</summary>

<p>

~~~sh
2018-12-05 15:59:23,257 884 [DEBUG] - XmlConfiguration is now operational
2018-12-05 15:59:23,271 884 [DEBUG] - Evaluating license file found at 'C:\ProgramData\chocolatey\license\chocolatey.license.xml'
2018-12-05 15:59:23,404 884 [DEBUG] - Loading Licensed Extensions and Commands
2018-12-05 15:59:23,417 884 [INFO ] - ============================================================
2018-12-05 15:59:23,555 884 [INFO ] - Chocolatey v0.10.11 Business
2018-12-05 15:59:23,563 884 [DEBUG] - Chocolatey is running on Windows v 6.1.7601.65536
2018-12-05 15:59:23,564 884 [DEBUG] - Attempting to delete file "C:/ProgramData/Chocolatey/choco.exe.old".
2018-12-05 15:59:23,565 884 [DEBUG] - Attempting to delete file "C:\ProgramData\Chocolatey\choco.exe.old".
2018-12-05 15:59:23,574 884 [DEBUG] - Command line: "C:\ProgramData\Chocolatey\choco.exe" upgrade transifex-client
2018-12-05 15:59:23,574 884 [DEBUG] - Received arguments: upgrade transifex-client
2018-12-05 15:59:23,615 884 [DEBUG] - RemovePendingPackagesTask is now ready and waiting for PreRunMessage.
2018-12-05 15:59:23,617 884 [DEBUG] - CountdownTask is now ready and waiting for PostRunMessage.
[...]

PR #1751, opened 2019-03-12
(GH-1689) Remove packaging script before upgrade

gep13:
It is possible, due to the way that Package Reducer works, that there
are extra files left in the package directory when doing an upgrade
of a package.  This can cause a package upgrade to fail.  Prior to
performing the upgrade, delete all packaging scripts (chocolateyInstall,
chocolateyUninstall, and chocolateyBeforeModify) from the package
folder, these will then be replaced during the upgrade, if they are
still required.

### DIFF
--- a/src/chocolatey/infrastructure.app/services/NugetService.cs
+++ b/src/chocolatey/infrastructure.app/services/NugetService.cs
@@ -899,7 +899,7 @@ Please see https://chocolatey.org/docs/troubleshooting for more
                 var pkgInfo = _packageInfoService.get_package_information(installedPackage);
                 bool isPinned = pkgInfo.IsPinned;
 
-                // if the package is pinned and we are skipping pinned, 
+                // if the package is pinned and we are skipping pinned,
                 // move on quickly
                 if (isPinned && config.OutdatedCommand.IgnorePinned)
                 {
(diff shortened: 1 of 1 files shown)


src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -1167,6 +1169,26 @@ public virtual void backup_existing_version(ChocolateyConfiguration config, IPac
             }
         }
 
+        public virtual void remove_packaging_files_prior_to_upgrade(string directoryPath)
+        {
+            // Due to the way that Package Reducer works, there is a potential that a Chocolatey Packaging
+            // script could be incorrectly replaced during an upgrade operation.  To guard against this,
```
ferventcoder: incorrectly left in place, not replaced. We want it replaced. 👍

ferventcoder:
As expected, there is an integration test that verifies what is removed and what should not be, so it will need adjusted to account for this. I can manage grabbing that if I finish it tonight. Otherwise you may want to review the failed AppVeyor log and take a look at the test to update it for this behavior.

gep13:
@ferventcoder I went down the route of changing the tests, but then I realised that the function that I added wasn't constrained to only happen during an upgrade, which I believe it should be.  I have added this in, and the tests that were failing are now all passing again.

Let me know if you want additional changes.

ferventcoder:
This was rebased against stable and merged in at https://github.com/chocolatey/choco/commit/cb2799b09f0a9c91e652e264e2dfc496c9724323. Thanks!

PR #1752, opened 2019-03-13
 (GH-1603) Ensure full path to log is used

### LINKED ISSUE #1603: Logging - Choco --log-file parameter does not create log files relative to current directory
This was checked using FOSS and all licensed versions of v0.10.11 with the same results.

### What You Are Seeing?

When using the `--log-file` parameter I see the following:

* If you specify a relative filename (ie. `--log-file=test.log`) the file is created under the Chocolatey ProgramData folder rather than in the current folder;
* If you specify a relative path (ie. `--log-file=temp\test.log`) an empty folder `temp` is created under the current folder and the full path (`temp\test.log`) is created under the Chocolatey ProgramData folder;
* If you specify a full path (ie. `--log-file=c:\temp\test.log`) the log file is created correctly at `c:\temp\test.log`;

### What is Expected?

I'd expect relative paths to be create under the current folder and not under Chocolatey ProgramData.

`choco new test <LOG FILE PARAMS FROM ABOVE>`

gep13:
Use helper method in IFileSystem implementation to ensure that the full
path to the requested log file is used, rather than making it relative
to where Chocolatey is currently located.

Fixes #1603

### DIFF
--- a/src/chocolatey.console/Program.cs
+++ b/src/chocolatey.console/Program.cs
@@ -83,7 +83,7 @@ namespace chocolatey.console
 
                 if (!string.IsNullOrWhiteSpace(config.AdditionalLogFileLocation))
                 {
-                  Log4NetAppenderConfiguration.configure_additional_log_file(config.AdditionalLogFileLocation);
+                  Log4NetAppenderConfiguration.configure_additional_log_file(fileSystem.get_full_path(config.AdditionalLogFileLocation));
                 }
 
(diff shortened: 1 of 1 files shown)


ferventcoder:
Rebased onto stable and merged in at https://github.com/chocolatey/choco/commit/3cd9c0d7765bb35748b3398c3111971abe7343ba. Thank you sir!

PR #1773, opened 2019-03-21
(GH-1772) Ensure chocolateyForceX86 is only set when requested

### LINKED ISSUE #1772: Chocolatey remembers --forceX86 value
### What You Are Seeing?

Boxstarter calling Chocolatey where the first package specifies --x86 then causes all subsequent packages to be installed as though --x86 was specified

### What is Expected?
Only install x86 packages when explicitly requested

This relates to https://github.com/chocolatey/boxstarter/issues/240

Looking at [prepare_powershell_environment](https://github.com/chocolatey/choco/blob/master/src/chocolatey/infrastructure.app/services/PowershellService.cs#L364), there's a bunch of environment variables that are nulled out at the start, but not `chocolateyForceX86`.

I'm thinking this should be nulled out as well

flcdrg:
When the Chocolatey library is called multiple times (eg. by Boxstarter), then don't persist the `chocolateyForceX86` environment variable between invocations

- Fixes #1772


flcdrg:
Yeah.. i did see the other one but wasn't sure about whether that was in the same boat. All good!

Issue #1806, opened 2019-04-20
Support upgrading Cygwin packages


tylerszabo:
Cygwin setup supports `--upgrade-also` (see https://cygwin.com/faq/faq.html#faq.setup.cli) which allows Cygwin to perform an unattended upgrade of all installed packages.

```
choco upgrade -s=cygwin all
```

Related to #1355

ferventcoder:
Thanks for the issue and the contribution! We'll take a look at this in more depth hopefully in the next few weeks.

bcurran3:
If this moves forward, consider also Python's PIP.
(Will create as issue in the future if/when relevant.)

tylerszabo:
I'd like to reconsider this feature; the PR is still viable.

PR #1807, opened 2019-04-20
(GH-1806) Support upgrading Cygwin packages

tylerszabo:


tylerszabo:
To test roll back a Cygwin package using the following (you may need to alter it slightly if you installed with non-default params)

```
"C:\tools\cygwin\cygwinsetup.exe" --no-desktop --no-startmenu --root C:\tools\cygwin --local-package-dir C:\tools\cygwin\packages --site http://mirrors.kernel.org/sourceware/cygwin/
```

There's a fair bit of duplication; I'm open to doing a refactor but I err'd on the side of minimal changes to existing functionality.

This is my first contribution to Chocolatey itself so I may need a few revisions to get everything right,. but I'm looking forward to it! :)

Issue #1800, opened 2019-04-17
Support ARM64 by selecting x86 32bit architecture


joaocgreis:
When running on ARM64 Windows (`PROCESSOR_ARCHITECTURE` environment variable is `ARM64` or `Get-WMIObject win32_Processor | Select-Object Architecture` returns 12), Chocolatey should default to install x86 packages instead of x64.

This is because Windows can only emulate 32 bit x86. Running 64 bit executables is not supported. I don't know if there are plans to support ARM64 natively in the future, but for now this change would make Chocolatey run well.

cc @jkunkee @MichelLopez

### What You Are Seeing?

On ARM64, packages that have a x64 version fail either during installation or when used.

If `--forcex86` is used, packages install and work well.

### What is Expected?

The x86 version should be installed without having to specify `--forcex86` when running on ARM64.

### How Did You Get This To Happen? (Steps to Reproduce)

Install any package with a x64 version on a ARM64 Windows machine.

### Output Log

<details>
<summary>Full Log Output</summary>

<p>

~~~sh
2019-04-17 08:00:02,348 7424 [DEBUG] - XmlConfiguration is now operational
2019-04-17 08:00:02,724 7424 [INFO ] - ============================================================
2019-04-17 08:00:04,283 7424 [INFO ] - Chocolatey v0.10.13
2019-04-17 08:00:04,331 7424 [DEBUG] - Chocolatey is running on Windows v 10.0.17134.0
2019-04-17 08:00:04,347 7424 [DEBUG] - Attempting to delete file "C:/ProgramData/chocolatey/choco.exe.old".
2019-04-17 08:00:04,347 7424 [DEBUG] - Attempting to delete file "C:\ProgramData\chocolatey\choco.exe.old".
2019-04-17 08:00:04,395 7424 [DEBUG] - Command line: "C:\ProgramData\chocolatey\choco.exe" install -y python2 --debug --verbose
2019-04-17 08:00:04,411 7424 [DEBUG] - Received arguments: install -y python2 --debug --verbose
2019-04-17 08:00:04,797 7424 [DEBUG] - RemovePendingPackagesTask is now ready and waiting for PreRunMessage.
2019-04-17 08:00:04,880 7424 [DEBUG] - Sending message 'PreRunMessage' out if there are subscribers...
2019-04-17 08:00:04,912 7424 [DEBUG] - [Pending] Removing all pending packages that should not be considered installed...
[...]

PR #1810, opened 2019-04-29
(GH-1800) Prefer 32-bit in ARM64 emulated PowerShell

joaocgreis:
https://github.com/chocolatey/choco/issues/1800 was fixed by @ferventcoder in https://github.com/chocolatey/choco/commit/b30427d620f1ea27a4a3dbc205cb528247dd2d0a. However, the fix does not work because PowerShell runs as an emulated x86 binary in ARM64. Thus, it is also necessary to check the `PROCESSOR_ARCHITEW6432` to determine if Chocolatey should prefer 32-bit packages.

Ref: https://github.com/chocolatey/choco/issues/1800#issuecomment-486038173


ferventcoder:
This has been rebased against stable and merged in at https://github.com/chocolatey/choco/commit/0a16b41a25706b1e52322035500dead8aeea429c. This will be released in 0.10.14. Thanks so much for the assist!

PR #1825, opened 2019-05-13
(#357) Add Export Command

### LINKED ISSUE #357: Export a packages.config file
Leaving this here as a suggested enhancement because I couldn't find it anywhere. If there's support for it, I'll look into whether I could contribute the change myself.
### Use case / story

Sometimes, my whole system will become bricked (such as it did yesterday -- ouch). However, this is a lot less painful because [I export my package list](http://seankilleen.com/2014/01/quick-tip-chocolatey-powershell-dropbox-for-package-management-and-backup/) and restore using that. 

Have I mentioned that Chocolatey is amazing? :)

However, there is still some pain here -- I have to manually concatenate the package names, remove the explicit version numbers, etc. 
### Proposed change
- Add a command, such as `choco export` or `cexp` 
- Allow switches for format -- I suggest the below:
  - list: a list of the package names and their versions
  - conf: a packages.conf file
- Required argument for the path of export file
### Example usage

In a scheduled task, I can see running something along the lines of the following:

`choco export -conf C:\MyBackupLocation\packages_yyyy-mm-dd.conf`

Then, the next time my system gets bricked, it should be as easy as:

`choco install -y C:\MyBackupLocation\packages_yyyy-mm-dd.conf`

Thoughts?

gep13:
This will allow the creation of a packages.config file of all the
currently installed packages on the machine.  Usage of this command
will be:

choco export -o c:/temp/packages.config --include-version-numbers

This command is particularly useful when re-building a machine.  i.e.
First export all packages currently installed on machine, and then
replay this packages.config via choco install packages.config on new
machine.

Fixes #357

### DIFF
--- a/src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
+++ b/src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
@@ -1,0 +1,197 @@
+﻿// Copyright © 2017 - 2018 Chocolatey Software, Inc
+// Copyright © 2011 - 2017 RealDimensions Software, LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+//
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace chocolatey.infrastructure.app.commands
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.IO;
+    using System.Text;
+    using System.Xml;
+    using attributes;
+    using commandline;
+    using configuration;
+    using filesystem;
+    using infrastructure.commands;
+    using logging;
+    using services;
+    using tolerance;
+
+    [CommandFor("export", "exports list of currently installed packages")]
+    public class ChocolateyExportCommand : ICommand
+    {
+        private readonly INugetService _nugetService;
+        private readonly IFileSystem _fileSystem;
+
+        public ChocolateyExportCommand(INugetService nugetService, IFileSystem fileSystem)
+        {
+            _nugetService = nugetService;
+            _fileSystem = fileSystem;
+        }
+
+        public void configure_argument_parser(OptionSet optionSet, ChocolateyConfiguration configuration)
+        {
+            optionSet
+                .Add("o=|output-file-path=",
+                     "Output File Path - the path to where the list of currently installed packages should be saved. Defaults to packages.config.",
+                     option => configuration.ExportCommand.OutputFilePath = option.remove_surrounding_quotes())
+                .Add("include-version-numbers|include-version",
+                     "Include Version Numbers - controls whether or not version numbers for each package appear in generated file.  Defaults to false.",
+                     option => configuration.ExportCommand.IncludeVersionNumbers = option != null)
+                ;
+        }
+
+        public void handle_additional_argument_parsing(IList<string> unparsedArguments, ChocolateyConfiguration configuration)
+        {
+            configuration.Input = string.Join(" ", unparsedArguments);
+
+            if (string.IsNullOrWhiteSpace(configuration.ExportCommand.OutputFilePath) && unparsedArguments.Count >=1)
+            {
+                configuration.ExportCommand.OutputFilePath = unparsedArguments[0];
+            }
+
+            // If no value has been provided for the OutputFilePath, default to packages.config
+            if (string.IsNullOrWhiteSpace(configuration.ExportCommand.OutputFilePath))
+            {
+                configuration.ExportCommand.OutputFilePath = "packages.config";
+            }
+        }
+
+        public void handle_validation(ChocolateyConfiguration configuration)
+        {
+            // Currently, no additional validation is required.
+        }
+
+        public void help_message(ChocolateyConfiguration configuration)
+        {
+            this.Log().Info(ChocolateyLoggers.Important, "Export Command");
+            this.Log().Info(@"
+Export all currently installed packages to a file.
+
+This is especially helpful when re-building a machine that was created
+using Chocolatey.  Export all packages to a file, and then re-install
+those packages onto new machine using `choco install packages.config`.
+");
+            "chocolatey".Log().Info(ChocolateyLoggers.Important, "Usage");
+            "chocolatey".Log().Info(@"
+    choco export [<options/switches>]
+");
+
+            "chocolatey".Log().Info(ChocolateyLoggers.Important, "Examples");
+            "chocolatey".Log().Info(@"
+    choco export
+    choco export --include-version-numbers
+    choco export ""'c:\temp\packages.config'""
+    choco export ""'c:\temp\packages.config'"" --include-version-numbers
+    choco export -o=""'c:\temp\packages.config'""
+    choco export -o=""'c:\temp\packages.config'"" --include-version-numbers
+    choco export --output-file-path=""'c:\temp\packages.config'""
+    choco export --output-file-path=""'c:\temp\packages.config'"" --include-version-numbers
+
+NOTE: See scripting in the command reference (`choco -?`) for how to
+ write proper scripts and integrations.
+
+");
+
+            "chocolatey".Log().Info(ChocolateyLoggers.Important, "Exit Codes");
+            "chocolatey".Log().Info(@"
+Exit codes that normally result from running this command.
+
+Normal:
+ - 0: operation was successful, no issues detected
+ - -1 or 1: an error has occurred
+
+If you find other exit codes that we have not yet documented, please
+ file a ticket so we can document it at
+ https://github.com/chocolatey/choco/issues/new/choose.
+
+");
+
+            "chocolatey".Log().Info(ChocolateyLoggers.Important, "Options and Switches");
+        }
+
+        public bool may_require_admin_access()
+        {
+            return false;
+        }
+
+        public void noop(ChocolateyConfiguration configuration)
+        {
+            this.Log().Info("Export would have been with options: {0} Output File Path={1}{0} Include Version Numbers:{2}".format_with(Environment.NewLine, configuration.ExportCommand.OutputFilePath, configuration.ExportCommand.IncludeVersionNumbers));
+        }
+
+        public void run(ChocolateyConfiguration configuration)
+        {
+            var packageResults = _nugetService.get_all_installed_packages(configuration);
+            var settings = new XmlWriterSettings { Indent = true, Encoding = new UTF8Encoding(false) };
+
+            FaultTolerance.try_catch_with_logging_exception(
+                () =>
+                {
+                    using (var stringWriter = new StringWriter())
+                    {
+                        using (var xw = XmlWriter.Create(stringWriter, settings))
+                        {
+                            xw.WriteProcessingInstruction("xml", "version=\"1.0\" encoding=\"utf-8\"");
+                            xw.WriteStartElement("packages");
+
+                            foreach (var packageResult in packageResults)
+                            {
+                                xw.WriteStartElement("package");
+                                xw.WriteAttributeString("id", packageResult.Package.Id);
+
+                                if (configuration.ExportCommand.IncludeVersionNumbers)
+                                {
+                                    xw.WriteAttributeString("version", packageResult.Package.Version.ToString());
+                                }
+
+                                xw.WriteEndElement();
+                            }
+
+                            xw.WriteEndElement();
+                            xw.Flush();
+                        }
+
+                        var fileExists = _fileSystem.file_exists(configuration.ExportCommand.OutputFilePath);
+
+                        // If the file doesn't already exist, just write the new one out directly
+                        if (!fileExists)
+                        {
+                            _fileSystem.write_file(
+                                configuration.ExportCommand.OutputFilePath,
+                                stringWriter.GetStringBuilder().ToString(),
+                                new UTF8Encoding(false));
+
+                            return;
+                        }
+
+
+                        // Otherwise, create an update file, and resiliently move it into place.
+                        var tempUpdateFile = configuration.ExportCommand.OutputFilePath + "." + Process.GetCurrentProcess().Id + ".update";
+                        _fileSystem.write_file(tempUpdateFile,
+                            stringWriter.GetStringBuilder().ToString(),
+                            new UTF8Encoding(false));
+
+                        _fileSystem.replace_file(tempUpdateFile, configuration.ExportCommand.OutputFilePath, configuration.ExportCommand.OutputFilePath + ".backup");
+                    }
+                },
+                errorMessage: "Error exporting currently installed packages",
+                throwError: true
+            );
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


RichiCoder1:
Does this/could this include pinned packages info? And/or installation/package args?

gep13:
There is no reason that additional features couldn’t be added to this new command. I would see that as a separate enhancement issue personally though, as there would be design decisions that would need to be made around the persisted arguments. @ferventcoder and I have spoken briefly about changes that need to happen in this area, and things would need to be flushed out.

Regarding the pinned package side of things, are you asking that if a pinned is pinned, this makes it into the packages.config file and when replayed on a new machine, that package also ends up being pinned?

gep13:
@RichiCoder1

RichiCoder1:
> There is no reason that additional features couldn’t be added to this new command. I would see that as a separate enhancement issue personally though, as there would be design decisions that would need to be made around the persisted arguments. @ferventcoder and I have spoken briefly about changes that need to happen in this area, and things would need to be flushed out.

Fair, that makes sense.

> Regarding the pinned package side of things, are you asking that if a pinned is pinned, this makes it into the packages.config file and when replayed on a new machine, that package also ends up being pinned?

Exactly. So if I pinned VS Code on one machine, it would end up pinned on another machine.

gep13:
> @RichiCoder1 said...
Exactly. So if I pinned VS Code on one machine, it would end up pinned on another machine

In the current implementation, pin is a standalone command, I.e. you can’t pin a package during the process of installation. As such, this may need to be implemented as a second pass of the packages.config file to look for pinned packages, or a modification to the install command to also pin when installing.

Again, more discussion would need to be had on this topic.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+        public ChocolateyExportCommand(INugetService nugetService, IFileSystem fileSystem)
+        {
+            _nugetService = nugetService;
+            _fileSystem = fileSystem;
+        }
+
+        public void configure_argument_parser(OptionSet optionSet, ChocolateyConfiguration configuration)
+        {
+            optionSet
+                .Add("o=|output-file-path=",
```
ferventcoder: Do we use `o` anywhere else with regards to output?

Additionally we use `output` in a few places, let's keep this consistent.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+                     "Output File Path - the path to where the list of currently installed packages should be saved. Defaults to packages.config.",
+                     option => configuration.ExportCommand.OutputFilePath = option.remove_surrounding_quotes())
+                .Add("include-version-numbers",
```
ferventcoder: Could we reuse `--version` here?

gep13: No :smile:

ferventcoder: Right on, makes sense. I wasn't thinking reusing `version` really explained what this was doing.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+                     "Include Version Numbers - controls whether or not version numbers for each package appear in generated file.  Defaults to false.",
+                     option => configuration.ExportCommand.IncludeVersionNumbers = option != null)
+                ;
+        }
+
+        public void handle_additional_argument_parsing(IList<string> unparsedArguments, ChocolateyConfiguration configuration)
+        {
+            if (unparsedArguments.Count > 0)
```
ferventcoder: There is a feature switch to ignore unparsed arguments, we should ensure that is on before throwing an error.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+
+        public void help_message(ChocolateyConfiguration configuration)
+        {
+            this.Log().Info(ChocolateyLoggers.Important, "Export Command");
+            this.Log().Info(@"
+Export all currently installed packages to a file.
+
+This is especially helpful when re-building a machine that was created
+using Chocolatey.  Export all packages to a file, and then re-install
+those packages onto new machine using `choco install packages.config`.
```
mwallner: typo? - https://github.com/gep13/choco/pull/1

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+
+            "chocolatey".Log().Info(ChocolateyLoggers.Important, "Exit Codes");
+            "chocolatey".Log().Info(@"
+Exit codes that normally result from running this command.
+
+Normal:
+ - 0: operation was successful, no issues detected
+ - -1 or 1: an error has occurred
+
+If you find other exit codes that we have not yet documented, please
```
mwallner: typo? - https://github.com/gep13/choco/pull/1

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
AdmiringWorm: Using `--include-version` would maybe make sense?
Less verbose, but the meaning isn't lost.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+ file a ticket so we can document it at
+ https://github.com/chocolatey/choco/issues/new/choose.
```
ferventcoder: Very nice! I see you've already brought in the new exit codes!

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+            "chocolatey".Log().Info(ChocolateyLoggers.Important, "Usage");
+            "chocolatey".Log().Info(@"
+    choco export [<options/switches>]
+");
+
+            "chocolatey".Log().Info(ChocolateyLoggers.Important, "Examples");
+            "chocolatey".Log().Info(@"
+    choco export
+    choco export --include-version-numbers
+    choco export --output-file-path=""c:\temp\packages.config""
```
ferventcoder: Here we should use best practices, so `--output-file-path=""'c:\temp\packages.config'""` (adding the apostrophes).

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+
+        public void noop(ChocolateyConfiguration configuration)
+        {
+            this.Log().Info("Export would have been with options: {0} Output File Path={1}{0} Include Version Numbers:{2}".format_with(Environment.NewLine, configuration.ExportCommand.OutputFilePath, configuration.ExportCommand.IncludeVersionNumbers));
+        }
+
+        public void run(ChocolateyConfiguration configuration)
+        {
+            var packageResults = _nugetService.get_all_installed_packages(configuration);
+            var settings = new XmlWriterSettings { Indent = true, Encoding = new UTF8Encoding(false) };
```
ferventcoder: Would like to be able to include other items to build out a bigger packages.config (for future use). One thing I would like to see here at the very least is to refactor to IXmlService and XmlService.

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
```diff
@@ -544,4 +547,17 @@ public sealed class ProxyConfiguration
         public string BypassList { get; set; }
         public bool BypassOnLocal { get; set; }
     }
+
+    [Serializable]
```
ferventcoder: Nice.

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
```diff
@@ -544,4 +547,17 @@ public sealed class ProxyConfiguration
         public string BypassList { get; set; }
         public bool BypassOnLocal { get; set; }
     }
+
+    [Serializable]
+    public sealed class ExportCommandConfiguration
+    {
+        public ExportCommandConfiguration()
+        {
+            OutputFilePath = "packages.config";
```
ferventcoder: Do we do this other places?

gep13: Yip, in at least one other place: https://github.com/chocolatey/choco/blob/master/src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs#L410

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
gep13: @ferventcoder no, this is the first place that `o` is introduced as a short option, so I think we are good to use it, if you agree.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
gep13: Having thought about this, throwing an error here doesn't make any sense.  Original code borrowed from pin command, which was handling parsing of subcommand, which is required here, so let's remove this code.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
gep13: @mwallner agreed.  Minor tweak required on your PR, and then I can get it pulled in.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
gep13: @mwallner actually, no.  This part does read ok, can you update your PR to not include this change.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
gep13: Agreed, will implement this.

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
gep13: Since this is done elsewhere, I think we are good to do this here as well.

mwallner:
is there a timeline for this feature to become available? :-)

gep13:
@mwallner I have some tidy up work that I need to do on this PR before it would be ready to be merged in.  I haven't forgotten about it, just need to get round to it.  So, to answer your question, no, there isn't a timeline 😄

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
gep13: @ferventcoder do you remember what the thought process was here again?  Looking at the code again just now, and I am not immediately clear on why this would help here.  Thanks

gep13:
@ferventcoder I believe this is ready for another review.  There is one outstanding question I believe with regard to the use of XmlService, which I am not clear on.  If this is required, can you let me know exactly how you see this working?  Thanks

eabase:
@ferventcoder 
Maybe it's time to merge this?

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
vexx32: Does this need to be an option switch? To me reading it, it feels kind of like i'd expect `choco export ./packages.config` to work without requiring an `-o`. The filepath here feels like it should be the primary argument?

vexx32 (approved):
Overall this looks good to me, bar one question I had around the design here: https://github.com/chocolatey/choco/pull/1825#discussion_r616830796

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
gep13: That is a good point. In order to be in-keeping with the rest of the Chocolatey commands, I have added the ability to either provide the explicit option, or not.  So the following all work:

```
    choco export
    choco export "'c:\temp\packages.config'"
    choco export -o="'c:\temp\packages.config'"
    choco export --output-file-path="'c:\temp\packages.config'"
```

gep13:
@vexx32 I have updated this PR based on your feedback.  If you could take another look, I would appreciate it.  I have also added some Unit Tests since the last review.  Thanks

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -0,0 +1,197 @@
+﻿// Copyright © 2017 - 2018 Chocolatey Software, Inc
```
vexx32, line 1: I guess not specifically related to this PR, but should we look at updating the license / copyright dates for this repo at some point? 🤔

Should this have a newer date since it's a newer file? Not sure what the process is for these tbqh.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -0,0 +1,197 @@
...
+            if (string.IsNullOrWhiteSpace(configuration.ExportCommand.OutputFilePath) && unparsedArguments.Count >=1)
+            {
+                configuration.ExportCommand.OutputFilePath = unparsedArguments[0];
+            }
+
+            // If no value has been provided for the OutputFilePath, default to packages.config
+            if (string.IsNullOrWhiteSpace(configuration.ExportCommand.OutputFilePath))
+            {
+                configuration.ExportCommand.OutputFilePath = "packages.config";
+            }
```
vexx32, line 71: If I call `choco export` with no other parameters, will this code be reached? Does `handle_additional_argument_parsing()` get called if there are no additional arguments to parse?

vexx32:
Apart from those two things, everything here looks good to me!

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
gep13: I actually had the same thought about this as I was adding it in....

I have created an issue to track getting this done ahead of the next release: https://github.com/chocolatey/choco/issues/2261 @ferventcoder has a method for bulk updating these, so will get with him about how that is done.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
gep13: Yes, this method is always executed.

vexx32 (approved):
Looks good to me, nice work! 💖

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
gep13: I have went through some notes that I had on this, as well as watching back the videos of the Twitch streams that we did on this, and the issue here was (I believe) that if we go down this route, there was no direct way of including/excluding specific attributes within the resulting XML file.  Not without implementing a custom XML Serializer/DeSerializer for the class.  I would suggest that this might be something that we could look at if/when we bring in the ability to include additional attributes into the resulting XML file, but for now, I think the code that we have here is a good compromise.

PR #1826, opened 2019-05-13
(GH-1816) Add NuGet.Config beside sln file

### LINKED ISSUE #1816: Choco project NuGet.config does not specify any package source so build fails unless dev has local configuration
### What You Are Seeing?
            restore_packages:
            
                 [echo] Finding and restoring all packages (c:\src\choco\lib\NuGet\NuGet.exe restore 'c:\src\choco\src\chocolatey.sln')
...
                 [exec] Unable to find version '1.1.86' of package 'coveralls.io'.
                 [exec] Unable to find version '2.6.4' of package 'NUnit.Runners'.
                 [exec] c:\src\choco\.build\compile.step(66,6):
                 [exec] External Program Failed: c:\src\choco\lib\NuGet\NuGet.exe (return code was 1)

This is also what caused the initial issue for https://github.com/chocolatey/choco/issues/700

### What is Expected?
The project should define the source(s) it depends on.

This is as simple as adding the three lines:

```
  <packageSources>
    <add key="NuGet official package source" value="https://api.nuget.org/v3/index.json" />
  </packageSources>
```

To existing nuget.config

Then you well get successful package restore that doesn't depend on the dev machine config of NuGet:

            restore_packages:
            
                 [echo] Finding and restoring all packages (c:\src\choco\lib\NuGet\NuGet.exe restore 'c:\src\choco\src\chocolatey.sln')
                 [exec] MSBuild auto-detection: using msbuild version '15.9.21.664' from 'C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\bin'.
                 [exec] Restoring NuGet package NUnit.Runners.2.6.4.
                 [exec] Restoring NuGet package coveralls.io.1.1.86.
                 [exec]   GET https://api.nuget.org/v3-flatcontainer/nunit.runners/2.6.4/nunit.runners.2.6.4.nupkg
                 [exec]   OK https://api.nuget.org/v3-flatcontainer/nunit.runners/2.6.4/nunit.runners.2.6.4.nupkg 31ms
                 [exec] Installing NUnit.Runners 2.6.4.
                 [exec] Adding package 'NUnit.Runners.2.6.4' to folder 'C:\CxCache'
                 [exec] Added package 'NUnit.Runners.2.6.4' to folder 'C:\CxCache'

Run build.bat after fresh repo cloning, don't have a machine specific definition of the nuget source.

gep13:
Depending on the end user setup, the machine level NuGet Configuration
file can include an entry for the v3 NuGet feed.  This causes problems
when restoring packages for the Chocolatey build, since two packages
(NUnit.Runners and coveralls) do not restore from the v3 feed, only the
v2 feed.  By specifying this feed in the solution level NuGet.Config
file, should mean that they restore correctly on any machine.

NOTE: The existing NuGet.Config file in the .nuget folder actually
doesn't seem to be being used at all.  Based on the documentation here:

https://docs.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior

A NuGet.Config file is looked for at the location of the sln file,
upwards to the root of the drive.  As a result, the existing file is
being ignored.

Fixes #1816

### DIFF
--- a/src/NuGet.Config
+++ b/src/NuGet.Config
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="NuGet Official v2 Package Source" value="https://www.nuget.org/api/v2/" />
+  </packageSources>
+</configuration>


src/NuGet.Config
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8?>
+<configuration>
+  <packageSources>
+    <add key="NuGet Official v2 Package Source" value="https://www.nuget.org/api/v2/" />
```
AdmiringWorm: You probably should also add `<clear />` before adding the new key.
Otherwise, this would only add the key to the existing sources on the developers machen.

gep13: @AdmiringWorm I actually debated about adding the `<clear/>` or not.  In the test machine that I set up for testing this out, it wasn't needed, so I opted to not include it.  I was reliably able to make the build fail until I made the change in this PR.

AdmiringWorm: I just remember that it caused issues for me in the past (several years ago I believe, just around the time nuget started using the V3 api).

If it isn't needed anymore, then that is just great.

devlead: Clear, could increase perf. as it would skip non working feeds.

ferventcoder: I think we'll want to ensure maximum compatibility in this case. Great suggestions.

PR #1833, opened 2019-05-26
(GH-53) Confirmation - Allow Yes To All

### LINKED ISSUE #53: install/upgrade/uninstall - allow confirming for remaining installs (aka Yes to All)
Choco seems to want a confirmation for each dependency.  It would be helpful to have an option on each prompt that says "Confirm all subsequent dependencies for package x".  Would be helpful to have this option on every dependency confirm prompt.

Similar to #52

ferventcoder:
Sometimes one may run upgrade all and forget to pass for all. It would
be helpful to be able to confirm all related at once.

Fixes #53

### DIFF
--- a/src/chocolatey/infrastructure.app/services/PowershellService.cs
+++ b/src/chocolatey/infrastructure.app/services/PowershellService.cs
@@ -213,7 +213,7 @@ namespace chocolatey.infrastructure.app.services
                     this.Log().Info(ChocolateyLoggers.Important, () => @"choco feature enable -n allowGlobalConfirmation");
 
                     var selection = InteractivePrompt.prompt_for_confirmation(@"Do you want to run the script?",
-                        new[] { "yes", "no", "print" },
+                        new[] { "yes", "all - yes to all", "no", "print" },
                         defaultChoice: null,
                         requireAnswer: true,
                         allowShortAnswer: true,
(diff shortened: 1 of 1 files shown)


src/chocolatey/infrastructure.app/services/PowershellService.cs
```diff
@@ -213,7 +213,7 @@ public bool run_action(ChocolateyConfiguration configuration, PackageResult pack
                     this.Log().Info(ChocolateyLoggers.Important, () => @"choco feature enable -n allowGlobalConfirmation");
 
                     var selection = InteractivePrompt.prompt_for_confirmation(@"Do you want to run the script?",
-                        new[] { "yes", "no", "print" },
+                        new[] { "yes", "all - yes to all", "no", "print" },
```
gep13: Would `always` be a better option here?

ferventcoder: This is in the context of that run only. Always implies that it would always do so on future runs.

ferventcoder: I wanted Yes to All, but I wanted the short answer `a` more so. The Interactive Prompt automatically uses the first letter to make the short answer and two "Y"s would cause it to be confused or produce unexpected results. This is effectively what it would produce: 
~~~
[Y]es
[A]ll - yes to all
[N]o
[P]rint
~~~

ferventcoder: To answer with the full bit, you would need to type back in either `a` or `all - yes to all`. The latter would be a bit weird and unexpected. Perhaps we should cut it off at `-` for matching.

ferventcoder:
This has been merged to stable at https://github.com/chocolatey/choco/commit/723f0dc2e04dba977fb8e6bde09be8d3f182db99 and has been released as part of Chocolatey v0.10.14.

PR #1886, opened 2019-07-20
(#1861) This improves the terminology in choco.exe help output

### LINKED ISSUE #1861: Terminology improvements in choco.exe help output
The following are produced by choco.exe, thus this isn't in documentation issues. All referenced output is via Chocolatey v0.10.15

```
Commands
 * install - installs packages from various sources
```
ENHANCEMENT: "various sources" just seems misleading. What percentage of free or commercial users actually install packages from multiple sources? I'd guess most use one source or two sources; i.e. internal repository and possibly the community repo as a 2nd. Maybe "installs packages" would be be better, simpler, cover the broader base.

```
* outdated - retrieves packages that are outdated. Similar to upgrade all --noop
```
BUG: It doesn't retrieve packages, the command retrieves package information about outdated packages/lists

````
* pack - packages up a nuspec to a compiled nupkg
````

ENHANCEMENT: Could say "packages nuspec, scripts, and other Chocolatey package resources into a nupkg file

````
* push - pushes a compiled nupkg
````

ENHANCEMENT: Could say "pushes a compiled nupkg to a repository (or maybe source)"

````
* new - generates files necessary for a chocolatey package from a template
````

ENHANCEMENT: new actually creates multiple template files as a package template. It could say "creates template files for creating a new Chocolatey package"

```
 * unpackself - have chocolatey set itself up
```

SUBJECTIVE: Might be be better stated as "Re-installs Chocolatey" or "Re-installs Chocolatey base files"

### PR DESCRIPTION
**Pull request by mprasanth**
This improves the terminology in choco.exe help output with the suggestions and new terminology.

Fixes #1861

### DIFF
--- a/src/chocolatey/infrastructure.app/commands/ChocolateyInstallCommand.cs
+++ b/src/chocolatey/infrastructure.app/commands/ChocolateyInstallCommand.cs
@@ -27,7 +27,7 @@ namespace chocolatey.infrastructure.app.commands
     using logging;
     using services;
 
-    [CommandFor("install", "installs packages from various sources")]
+    [CommandFor("install", "installs packages using configured sources")]
     public class ChocolateyInstallCommand : ICommand
     {
         private readonly IChocolateyPackageService _packageService;
--- a/src/chocolatey/infrastructure.app/commands/ChocolateyNewCommand.cs
+++ b/src/chocolatey/infrastructure.app/commands/ChocolateyNewCommand.cs
@@ -27,7 +27,7 @@ namespace chocolatey.infrastructure.app.commands
     using services;
     using templates;
 
-    [CommandFor("new", "generates files necessary for a chocolatey package from a template")]
+    [CommandFor("new", "creates template files for creating a new Chocolatey package")]
     public class ChocolateyNewCommand : ICommand
     {
         private readonly ITemplateService _templateService;
--- a/src/chocolatey/infrastructure.app/commands/ChocolateyPackCommand.cs
+++ b/src/chocolatey/infrastructure.app/commands/ChocolateyPackCommand.cs
@@ -26,7 +26,7 @@ namespace chocolatey.infrastructure.app.commands
     using logging;
     using services;
 
-    [CommandFor("pack", "packages up a nuspec to a compiled nupkg")]
+    [CommandFor("pack", "packages nuspec, scripts, and other Chocolatey package resources into a nupkg file")]
     public class ChocolateyPackCommand : ICommand
     {
         private readonly IChocolateyPackageService _packageService;
--- a/src/chocolatey/infrastructure.app/commands/ChocolateyOutdatedCommand.cs
+++ b/src/chocolatey/infrastructure.app/commands/ChocolateyOutdatedCommand.cs
@@ -24,7 +24,7 @@ namespace chocolatey.infrastructure.app.commands
     using logging;
     using services;
 
-    [CommandFor("outdated", "retrieves packages that are outdated. Similar to upgrade all --noop")]
+    [CommandFor("outdated", "retrieves information about packages that are outdated. Similar to upgrade all --noop")]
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

File: src/chocolatey/infrastructure.app/commands/ChocolateyPackCommand.cs [hunk 1]
```diff
@@ -26,7 +26,7 @@ namespace chocolatey.infrastructure.app.commands
     using logging;
     using services;
 
-    [CommandFor("pack", "packages up a nuspec to a compiled nupkg")]
+    [CommandFor("pack", "packages nuspec, scripts, and other Chocolatey package resources into a nupkg file")]
```
ferventcoder: 👍

src/chocolatey/infrastructure.app/commands/ChocolateyNewCommand.cs
```diff
@@ -27,7 +27,7 @@ namespace chocolatey.infrastructure.app.commands
     using services;
     using templates;
 
-    [CommandFor("new", "generates files necessary for a chocolatey package from a template")]
+    [CommandFor("new", "creates template files for creating a new Chocolatey package")]
```
ferventcoder: creates a new package, not a template. It generates a new package from a template.

src/chocolatey/infrastructure.app/commands/ChocolateyInstallCommand.cs
```diff
@@ -27,7 +27,7 @@ namespace chocolatey.infrastructure.app.commands
     using logging;
     using services;
 
-    [CommandFor("install", "installs packages from various sources")]
+    [CommandFor("install", "installs packages")]
```
ferventcoder: Various sources - whatever is set in the default - why remove this?

egit config merge.ff falsegit config merge.ff falsegit config merge.ff false
```diff
@@ -0,0 +1,475 @@
+[33mcommit 2ec31875f9865529afa399216096be0249b5787d[m[33m ([m[1;36mHEAD -> [m[1;32mhelp-command-enhancements[m[33m, [m[1;31morigin/master[m[33m, [m[1;31morigin/help-command-enhancements[m[33m, [m[1;31morigin/HEAD[m[33m, [m[1;32mmaster[m[33m)[m
```
ferventcoder: I think you might have gotten an extra file in here you'll need to remove.

src/chocolatey/infrastructure.app/commands/ChocolateyNewCommand.cs
AdmiringWorm: it kind of is template(d?) files though. It isn't really a package until the `.nupkg` file is created, right?

And usually the files isn't ready to be used as is without editing them. So IMHO, I think template (or templated) files would be more accurate?

src/chocolatey/infrastructure.app/commands/ChocolateyInstallCommand.cs
AdmiringWorm: Maybe

```suggestion
    [CommandFor("install", "installs packages using configured sources")]
```

would be somewhat better option?

gep13:
While the new cla-assistant.io may not have been signed, the original CLAHub one has been signed, as such, we can move forward with this PR:

image

src/chocolatey/infrastructure.app/commands/ChocolateyNewCommand.cs
gep13: I agree with @AdmiringWorm here. This is also in line with what is documented on the site here: https://docs.chocolatey.org/en-us/create/commands/new

gep13 (approved):
LGTM!

gep13:
@mprasanth thanks again for your help in getting this updated, really appreciate it!

PR #1902, opened 2019-08-21
(doc) Add security.md file

gep13:
This addition, changes the following:

image

To this:

image

Which, once you click on it, show the policy, rather than starting to create a new issue:

image

We might want to expand on the content a little bit.  For example, this is what @AdmiringWorm (who I am stealing the idea from) is using:

https://github.com/WormieCorp/.github/blob/master/.github/SECURITY.md

And also another example here:

https://github.com/standard/.github/blob/master/SECURITY.md

Thoughts?

### DIFF
--- a/.github/ISSUE_TEMPLATE/SecurityDisclosure.md
+++ b/.github/ISSUE_TEMPLATE/SecurityDisclosure.md
@@ -1,8 +1,0 @@
----
-name: Security Disclosure / Report
-about: Found a security issue?
----
-
-STOP RIGHT HERE - DO NOT CREATE A TICKET FOR A SECURITY FINDING IN THE OPEN (HERE OR IN ANY COMMUNITY)
-
-Security reports should never start out in the open. Please follow up directly with the team if you have a contact. If not you can always start with the information at https://chocolatey.org/security to see instructions on how to provide the disclosure. Thank you!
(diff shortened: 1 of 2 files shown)


gep13:
@pauby I actually thought the same. but from what I can see, I don't think we can control that 😢

pauby:
@gep13 I thought we could. I think you can do it by naming the templates so they sit in alphabetical order you need - see https://github.com/chocolatey-community/chocolatey-package-requests/tree/master/.github/ISSUE_TEMPLATE

gep13:
@pauby yes, but this file that I just added, sits outside the ISSUE_TEMPLATE folder.  It is at the root of the .github folder.

gep13:
So, it's not `technically` an issue template, since it appears elsewhere in the GitHub UI.

gep13:
@ferventcoder as a side note, we have the often of moving these files (once we are happy with them) to a new `.github` folder within the `chocolatey` organisation, and they would be applied to all repositories under this Organisation.

pauby:
> @pauby yes, but this file that I just added, sits outside the ISSUE_TEMPLATE folder. It is at the root of the .github folder.

To be fair, I did not check that 😄

AdmiringWorm:
I would also recommend adding a link to the security policy in the README.md (actually, github recommends it), as well as perhaps in the contribution guidelines.

ferventcoder:
LGTM!

Issue #1808, opened 2019-04-26
Don't show pinned package is pinned when upgrading


musm:
For packages that auto update I don't want to see 
`You have vscode v1.30.2 installed. Version 1.33.1 is available based on your source(s).` 
on every upgrade 

Can we disable this or just leave them in the warning portion?

musm:
bump

ferventcoder:
I believe there is a switch already for this. 

#580

ferventcoder:
Related to #994 and #580.

PR #1906, opened 2019-08-23
(GH-1808) Add ignore-pinned-messages option to upgrade command

RozzaysRed:
Previously the upgrade command would output messages of the upgrade for
a pinned package.  With this new ignore-pinned-messages option, there is a
check prior to outputting these messages to see if the package is pinned
and if this ignore message flag is set.  If it is, it will not output
the messages to the console.


ferventcoder:
Why was this closed?

Issue #798, opened 2016-06-14
install - pin package during installation


CWempe:
I just found out about chocolately and I think this is a great project!

Reading the issues here, I thought of a feature request.

I think it would be great if you could "pin" a package directly during (after) the installation.

Instead of 

```
choco install googlechrome
choco pin googlechrome
```

something like this
`choco install googlechrome --pin`

This would especialy come in handy when installing a list of packages via package.conf.

```
    <?xml version="1.0" encoding="utf-8"?>
    <packages>
      <package id="googlechrome" pin="true" />
    </packages>
```

Then again this might be obsolet (in some cases) when https://github.com/chocolatey/choco/issues/242 gets implemented.

ferventcoder:
@CWempe I believe this has been discussed before and it's a pretty great idea! I'm surprised that there wasn't already an existing ticket (I searched), perhaps from the prior discussions one was not created.

gep13:
Any change in this area will require an upstream change in the chocolatey.extension to ensure that any and all new logic is respected there.

gep13:
A follow up spike task has been created to confirm that this functionality continues to work as expected when the Chocolatey Licensed Extension is installed: https://gitlab.com/chocolatey/collaborators/choco-licensed/-/issues/631

PR #1915, opened 2019-09-11
(#798) Add pin-version parameter for choco install command to execute choco pin at the end

bergmeister:
Fixes https://github.com/chocolatey/choco/issues/798

This is the first minimum viable implementation. I haven't tested it yet as I'd like to see if it would pass the build validation and more importantly I would like to get early feedback on the implementation (as to whether the high level approach is OK) before writing more complex tests with which I'd appreciate help

### DIFF
--- a/src/chocolatey/infrastructure.app/commands/ChocolateyInstallCommand.cs
+++ b/src/chocolatey/infrastructure.app/commands/ChocolateyInstallCommand.cs
@@ -20,21 +20,27 @@ namespace chocolatey.infrastructure.app.commands
     using System.Collections.Generic;
     using System.Linq;
     using attributes;
+    using chocolatey.infrastructure.app.nuget;
     using commandline;
     using configuration;
     using infrastructure.commands;
     using infrastructure.configuration;
     using logging;
+    using NuGet;
     using services;
 
     [CommandFor("install", "installs packages from various sources")]
     public class ChocolateyInstallCommand : ICommand
     {
         private readonly IChocolateyPackageService _packageService;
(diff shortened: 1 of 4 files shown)


bergmeister:
Hmm, it seems I broke the IoC container mechanism that allows only 1 parameter in the constructor. Will need to have a look but would also appreciate some pointers

bergmeister:
The broken Travis build does not seem to be a problem of this PR but generally broken in master now.

src/chocolatey/infrastructure.app/commands/ChocolateyInstallCommand.cs
```diff
@@ -20,21 +20,27 @@ namespace chocolatey.infrastructure.app.commands
...
 
     [CommandFor("install", "installs packages from various sources")]
     public class ChocolateyInstallCommand : ICommand
     {
         private readonly IChocolateyPackageService _packageService;
+        private readonly ChocolateyPinCommand _chocolateyPinCommand;
+        private readonly ILogger _nugetLogger;
 
-        public ChocolateyInstallCommand(IChocolateyPackageService packageService)
+        public ChocolateyInstallCommand(IChocolateyPackageService packageService, ChocolateyPinCommand chocolateyPinCommand, ILogger nugetLogger)
```
ferventcoder: This should not be adjusted if possible. It will require more review on this aspect.

ferventcoder: oh wait, nevermind.

bergmeister:
No worries, I am glad we will get the functionality at least, which matters most.

PR #1941, opened 2019-10-20
(GH-1866) Restrict pending package removal to top level

### LINKED ISSUE #1866: remove pending packages task - search only top level level package folders to clean up
I have the problem that any choco command I issue takes quite a bit (multiple seconds) of time. This is even the case for something simple like "choco list --local-only".

Investigating this, I found that it is due to a combination of these factors: 
- The RemovePendingPackagesTask performs a full-depth file search of the PackagesLocation searching for pending package files.
- In some of my packages, I have directory symlinks that point to the actual location of the package data in the tools directory or that point to the source location of the package data on a network share.
- Some of my packages have large amount of package data (close to 1GB) and the network is not always extremely fast. This causes the full search to take up to few tens of seconds. 
- There is no way to prevent choco from searching full-depth, e.g. by marking the large or remote package data subdirectory somehow.

I wonder if it is really necessary to search a symlinked subdirectory for pending package files (since they are never copied around anyway). But in any case, I would be really happy if I could at least have the possibility to tell choco to not search a subdirectory (taking the responsibility myself if anything bad happens due to not searching, of course).

aerotog:
Resolves #1866.

Seemed like a pretty straight forward change. I considered adding a test, but there was no existing tests to mimic for the ITask implementing classes. Hopefully this change is small enough to not require additional code coverage.

### DIFF
--- a/src/chocolatey/infrastructure.app/tasks/RemovePendingPackagesTask.cs
+++ b/src/chocolatey/infrastructure.app/tasks/RemovePendingPackagesTask.cs
@@ -60,13 +60,13 @@ namespace chocolatey.infrastructure.app.tasks
         {
             this.Log().Debug(ChocolateyLoggers.Verbose, "[Pending] Removing all pending packages that should not be considered installed...");
 
-            var pendingFiles = _fileSystem.get_files(ApplicationParameters.PackagesLocation, ApplicationParameters.PackagePendingFileName, SearchOption.AllDirectories).ToList();
+            var pendingFiles = _fileSystem.get_files(ApplicationParameters.PackagesLocation, ApplicationParameters.PackagePendingFileName).ToList();
             foreach (var pendingFile in pendingFiles.or_empty_list_if_null())
             {
                 var packageFolder = _fileSystem.get_directory_name(pendingFile);
                 string packageFolderName = _fileSystem.get_directory_info_for(packageFolder).Name;
 
-                var pendingSkipFiles = _fileSystem.get_files(packageFolder, PENDING_SKIP_FILE, SearchOption.AllDirectories).ToList();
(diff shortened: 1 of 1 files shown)


src/chocolatey/infrastructure.app/tasks/RemovePendingPackagesTask.cs
```diff
@@ -60,13 +60,13 @@ private void handle_message(PreRunMessage message)
         {
             this.Log().Debug(ChocolateyLoggers.Verbose, "[Pending] Removing all pending packages that should not be considered installed...");
 
-            var pendingFiles = _fileSystem.get_files(ApplicationParameters.PackagesLocation, ApplicationParameters.PackagePendingFileName, SearchOption.AllDirectories).ToList();
+            var pendingFiles = _fileSystem.get_files(ApplicationParameters.PackagesLocation, ApplicationParameters.PackagePendingFileName).ToList();
```
ferventcoder: Is the default to search top level only? It would be better to make it explicit here - `SearchOption.TopDirectoryOnly` (or whatever the correct option is).

aerotog: Yes, SearchOption.TopDirectoryOnly is the default option. It might be better to make it explicit in case that ever changes, but Resharper does mark it as redundant code since it's the default already. Let me know if you want this changed or not.

aerotog:
> git commit message is lacking context of why

Commit body has been updated with more context.

ferventcoder:
Be pulling this in soon.

gep13:
@aerotog thank you for raising this PR.  I am going to move forward with merging in a related PR (https://github.com/chocolatey/choco/pull/2218) and will close this one out.

Issue #89, opened 2015-02-09
Remove the `cpack` command alias


rpavlik:
`cpack` is the command used by CMake to invoke its package generator. Not sure the polite way to resolve this.

And for reference, it also uses `cmake` and `ctest`.

(And apparently Microsoft is using CMake to build their open-source .NET on other platforms now...)

ferventcoder:
Aware of this. One reason we are moving away from the shorter ones (at least cpush and cpack).

dothebart:
Chocolatey should either prefix its binary or setup the path environment in a way that it doesn't disturb the packages it installed.

ferventcoder:
@dothebart Chocolatey itself does prefix the binary shortcuts with `c`. So it already does this. This is the only known case where that presents an issue.

solvingJ:
Just tried to use Cpack for Cmake on windows, chocolatey getting in the way.  What's the status of this?  Looks like a PR was merged 2 years ago?

ferventcoder:
@solvingJ Do you have a link to a PR?

ferventcoder:
@solvingJ If you are talking about https://github.com/chocolatey/choco/commit/e9dcb0cc12d49791511e8456b6e94438b81c369c,

deprecation != removal - deprecations keep functionality but announce it is being removed in the future. HTH

dothebart:
Maybe once the user installs cmake it could be clear, that he doesn't want a cpack alias and it could be removed then?

ferventcoder:
If you install CMake through Chocolatey using the portable version, it will overwrite cpack. However any actions to install packages or outside of Chocolatey's control, it becomes hard to know when those kind of occurrences happen. It's probably best to ensure there is documentation in troubleshooting about this, but over that, it's unknown what we can do until `cpack` is removed. We must follow a transition (deprecation) in removing things.

dothebart:
one sure thing is, that you usually would call the cmake - cpack with `-G <packageformat>` - maybe even inside of scrips. Unless you handle `-G` as a valid parameter in chocolateys cpack its a safe bet to search for cpack in the rest of the given paths and call that with $@ - or error out with a clear error message to the user that this is _not_ cmake cpack which he probably desired to invoke, and how he can remove the unwanted cpack with `del ...cpack.exe`

ferventcoder:
@dothebart you mean pass this on to a different cpack based on arguments. That is actually a really good idea!

ferventcoder:
[...]

PR #1946, opened 2019-10-23
(GH-89) Rename shims to an apt-style prefix

native-api:
'cpack' proved to conflict with CMake.
With a longer prefix, conflicts are practically impossible.

Fixes https://github.com/chocolatey/choco/issues/89


native-api:
Travis failure unrelated, caused by https://github.com/chocolatey/choco/issues/1020

ferventcoder:
Hi @native-api,
Many thanks for this, but this isn't something we've requested and not a direction we are going with respect to shims. Ultimately they are going to be removed, but that date is set more in line with v1, which you can see as the milestone for v1.

PR #1947, opened 2019-10-23
(GH-89) Remove shims

### LINKED ISSUE #89: Remove the `cpack` command alias

native-api:
Shims were deprecated since 26a98f86ce7e32946cb51a2631d77c07c5d0ce40.
Since `cpack` proved to conflict with CMake, this is an opportunity
to pull the plug on this feature.

Fixes #89.


native-api:
Travis failure unrelated, caused by https://github.com/chocolatey/choco/issues/1020

native-api:
Travis failure unrelated, caused by https://github.com/chocolatey/choco/issues/1020 and https://changelog.travis-ci.com/xenial-as-the-default-build-environment-99476

ferventcoder:
Hi @native-api,
Previously mentioned at https://github.com/chocolatey/choco/pull/1946#issuecomment-548127319. 

Just a note that #89 is linked to v1, not where we currently are at. There is work to be done in the interim where we need to tell folks those things are being removed. If you want to look at providing a warning when those are used, I'm all for that. However a removal or rename is not going to be considered at this time.

Issue #1716, opened 2019-01-22
Make PackageName default to ChocolateyPackageName


FranklinYu:
Following helpers have a parameter `-PackageName`:
- `Install-ChocolateyZipPackage`
- `Get-ChocolateyUnzip`

When they are used in install scripts (which is likely to be the case), this parameter seems redundant since we already have `$env:ChocolateyPackageName`. It would be great to let it default to `$env:ChocolateyPackageName` when this environment variable is available.

justinwalker4179:
I am a new contributor to this project and was wondering if I could be assigned this issue.

cowmanjoe:
I'm going to look into this as a first issue.

cowmanjoe:
Would this change not also apply to `Install-ChocolateyPackage` as well?

FranklinYu:
> Would this change not also apply to `Install-ChocolateyPackage` as well?

Yes I think so! Thank you for your Pull Request.

cowmanjoe:
Added this default value for several other scripts that use `$packageName`. If this is an optional parameter now that it has a default variable, shouldn't it not be in position 0? The problem is if we change the position, it breaks the API for other calling scripts.

FranklinYu:
> The problem is if we change the position, it breaks the API for other calling scripts.

Yes, that would be a breaking change. Although it sounds nice to have optional parameters after all the required ones, it’s not worth it to break packages, especially corporate internal ones (we are able to update the community feed ourselves).

Note that we already have some of such “wrong parameters”, like [`-Url` parameter of `Install-ChocolateyZipPackage`](https://chocolatey.org/docs/helpers-install-chocolatey-zip-package#url-string), for the same reason I guess.

PR #1974, opened 2019-12-30
(GH-1716) - User chocolateyPackageName as default parameter for packageName

cowmanjoe:
This sets the default value of `$packageName` to `$env:chocolateyPackageName` for` Install-ChocolateyZipPackage` and `Get-ChocolateyUnzip`. This is usually the value used as default in install scripts, so this can reduce the number of necessary parameters specified.

I followed what was stated in the ticket, but thought there might be some more scripts where this could be useful, such as `Install-ChocolateyPackage`. Should I add the defaults there as well?


cowmanjoe:
Closes #1716

PR #1979, opened 2020-01-05
(#1962) Use different message on error

### LINKED ISSUE #1962: When there is an error getting a package, do not report the package files upgrade aspect has completed
Hi all, it looks like the Chocolatey servers are (partially) down right now.  That's not the point of this enhancement request though.  Currently if the servers are (partially) down, you get the message shown below:
image

It doesn't really seem to make sense to report that an error has occurred and a package upgrade was not successful, before then saying the upgrade is complete and trying the same thing again.  Perhaps the whole process could fall over if the first network access fails?

cowmanjoe:
Currently, the method logs that the command was "completed" even if it failed. A log that says the step failed would make more sense to the user in this case.

Closes #1962.

### DIFF
--- a/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
+++ b/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
@@ -336,7 +336,9 @@ Did you know Pro / Business automatically syncs with Programs and
             EnvironmentSettings.reset_environment_variables(config);
             set_pending(packageResult, config);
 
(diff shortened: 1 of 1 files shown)


gep13 (approved):
LGTM!

gep13:
@cowmanjoe thank you very much for taking the time to make this change, really appreciate it!

Once the CI builds are completed, we can get this merged in.

gep13:
@cowmanjoe your changes have been merged, thanks for your contribution 👍

Issue #1986, opened 2020-01-09
refreshenv.cmd uses hard-coded temp files


jvuori:
### What You Are Seeing?
When running refreshenv at the same time under multiple command prompt windows it tries to write _envset.tmp and _env.cmd but ends up in a race condition situation as the files are locked by other process. This is not an issue when running refreshenv manually but when automating things and optimizing the automation by parallelism.

### What is Expected?
The temp files should be written with random file names.

### How Did You Get This To Happen? (Steps to Reproduce)
Write a script which launches multiple command prompts and runs refreshenv on each.

### Output Log
```
Refreshing environment variables from registry for cmd.exe. Please wait...The process cannot access the file because it is being used by another process.
The process cannot access the file because it is being used by another process.
The process cannot access the file because it is being used by another process.
The process cannot access the file because it is being used by another process.
The process cannot access the file because it is being used by another process.
The process cannot access the file because it is being used by another process.
C:\Users\Administrator\AppData\Local\Temp\_envset.tmp
C:\Users\Administrator\AppData\Local\Temp\_env.cmd
Finished..
```

PR #1987, opened 2020-01-12
(GH-1986) Use unique temp file names in RefreshEnv.cmd

jvuori:
Fix the issue with generating a GUID (with Powershell [guid]::NewGuid() function) and using it as a part of temporary file names.

Tried first %RANDOM% but it's not random enough. It didn't pass the test script which launches two refreshenv terminals almost at the same time. %TIME% worked better but but it's not 100% reliable either. Besides the format is locale specific so in certain countries it includes characters (colon) that can't be used in file names.

### DIFF
--- a/src/chocolatey.resources/redirects/RefreshEnv.cmd
+++ b/src/chocolatey.resources/redirects/RefreshEnv.cmd
@@ -11,20 +11,26 @@
 ::echo "RefreshEnv.cmd only works from cmd.exe, please install the Chocolatey Profile to take advantage of refreshenv from PowerShell"
 echo | set /p dummy="Refreshing environment variables from registry for cmd.exe. Please wait..."
 
+:: Generate unique file names
+for /F "tokens=* usebackq" %%F IN (`powershell -command "[guid]::NewGuid().ToString()"`) do set REFRESHENV_GUID=%%F
+set REFRESHENV_ENVSET_TMP="%TEMP%\refreshenv_envset_%REFRESHENV_GUID%.tmp"
+set REFRESHENV_ENVGET_TMP="%TEMP%\refreshenv_envget_%REFRESHENV_GUID%.tmp"
+set REFRESHENV_ENV_CMD="%TEMP%\refreshenv_env_%REFRESHENV_GUID%.cmd"
+
 goto main
 
 :: Set one environment variable from registry key
 :SetFromReg
-    "%WinDir%\System32\Reg" QUERY "%~1" /v "%~2" > "%TEMP%\_envset.tmp" 2>NUL
-    for /f "usebackq skip=2 tokens=2,*" %%A IN ("%TEMP%\_envset.tmp") do (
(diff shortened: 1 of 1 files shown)


src/chocolatey.resources/redirects/RefreshEnv.cmd
```diff
@@ -11,53 +11,63 @@
 ::echo "RefreshEnv.cmd only works from cmd.exe, please install the Chocolatey Profile to take advantage of refreshenv from PowerShell"
 echo | set /p dummy="Refreshing environment variables from registry for cmd.exe. Please wait..."
 
+:: Generate unique file names
+for /F "tokens=* usebackq" %%F IN (`powershell -command "[guid]::NewGuid().ToString()"`) do set REFRESHENV_GUID=%%F
```
ferventcoder: This would require at least a 2 second start up time for PowerShell. Not that I'm against that, but that is longer than the entire thing takes to run now. So this is more than doubling the time. Are there any other methods?

src/chocolatey.resources/redirects/RefreshEnv.cmd
```diff
@@ -11,53 +11,63 @@
 ::echo "RefreshEnv.cmd only works from cmd.exe, please install the Chocolatey Profile to take advantage of refreshenv from PowerShell"
 echo | set /p dummy="Refreshing environment variables from registry for cmd.exe. Please wait..."
 
+:: Generate unique file names
+for /F "tokens=* usebackq" %%F IN (`powershell -command "[guid]::NewGuid().ToString()"`) do set REFRESHENV_GUID=%%F
+set REFRESHENV_ENVSET_TMP="%TEMP%\refreshenv_envset_%REFRESHENV_GUID%.tmp"
+set REFRESHENV_ENVGET_TMP="%TEMP%\refreshenv_envget_%REFRESHENV_GUID%.tmp"
+set REFRESHENV_ENV_CMD="%TEMP%\refreshenv_env_%REFRESHENV_GUID%.cmd"
```
ferventcoder: Let's put this somewhere else besides temp. Somewhere more locked down, like a temp folder under the Chocolatey directories that is locked down.

src/chocolatey.resources/redirects/RefreshEnv.cmd
jvuori: I'm afraid pure CMD doesn't provide anything reliable we could really use. Sounds an odd idea but what if we just implemented some tiny win32 application (with C) that would only print its process ID and that would be used as the random part of the temporary file names? The executable would be so small that it could be even put into version control beside RefreshEnv.cmd.

jvuori: LOL! 13 824 bytes :)

```C
#include <stdio.h>
#include <processthreadsapi.h>

void main()
{
    printf("%d", GetCurrentProcessId());
}
```

`gcc pid.c -o pid.exe -s -Os`

PR #2006, opened 2020-02-24
(GH-2051) Fix: workingDirectory is always relpaced with the cache location

### LINKED ISSUE #2051: Get-ChocolateyUnzip / Start-ChocolateyProcessAsAdmin - Path always being replaced with cacheLocation
As a result of #1781, path is always being replaced with cacheLocation / TEMP for Get-ChocolateyUnzip.

ifreislich:
The previous fix to address UNC paths broke the ability to provide a 
working directory to the Start-ChocolateyProcessAsAdmin helper function.
The test "$workingDirectory.ProviderPath -eq $null" always evaluates to 
true because the String type does not have the member ProviderPath.

The fix is to obtain the location object only when no path is provided 
to the helper and to verify its validity before use, possibly falling 
back to the cache location.

Closes #2051


ferventcoder:
Rebased and merged into stable at https://github.com/chocolatey/choco/commit/8d78c0a6c31aaa1f63f6ee0dd58cda0ecca66da2. Thanks so much for this!

PR #2054, opened 2020-05-24
(#2052) Outdated: filter by package id

### LINKED ISSUE #2052: Outdated per package
`choco outdated mypackage`

This command should show only the outdated status of the one package provided.

Example output:
```
choco outdated mypackage
Chocolatey v0.10.16
Outdated Packages
 Output is package name | current version | available version | pinned?

mypackage|1.0.0|1.0.1|false
```

Basically it should check the outdated status of only that package.

cuichenli:
Resolves #2052 

The changed code enables the `outdated` command to run against one specific package.

### DIFF
--- a/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
+++ b/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
@@ -628,7 +628,11 @@ Would have determined packages that are out of date based on what is
  Output is package name | current version | available version | pinned?
 ");
 
-            config.PackageNames = ApplicationParameters.AllPackages;
+            if (config.PackageNames.Length == 0)
+            {
+                config.PackageNames = ApplicationParameters.AllPackages;
+            }
+
             config.UpgradeCommand.NotifyOnlyAvailableUpgrades = true;
 
             var output = config.RegularOutput;
--- a/src/chocolatey/infrastructure.app/services/NugetService.cs
+++ b/src/chocolatey/infrastructure.app/services/NugetService.cs
@@ -23,6 +23,7 @@ namespace chocolatey.infrastructure.app.services
     using System.IO;
     using System.Linq;
     using System.Net;
+    using System.Text;
     using NuGet;
     using adapters;
     using commandline;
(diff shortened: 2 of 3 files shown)


src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
```diff
@@ -628,7 +628,11 @@ public virtual void outdated_run(ChocolateyConfiguration config)
  Output is package name | current version | available version | pinned?
 ");
 
-            config.PackageNames = ApplicationParameters.AllPackages;
+            if (config.PackageNames.Length == 0)
```
ferventcoder, line 631: I thought at first we'd need to go back to handle additional argument parsing to ensure config.PackageNames is set, but I noticed the handle arg parsing method for the outdated command is already doing that. 👍

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -892,12 +893,19 @@ public virtual void remove_rollback_directory_if_exists(string packageName)
 
             var originalConfig = config;
 
+            var notFoundedMessage = new StringBuilder();
```
ferventcoder, line 896: We already have messaging on not found. Can you reuse that?

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -892,12 +893,19 @@ public virtual void remove_rollback_directory_if_exists(string packageName)
...
+
             foreach (var packageName in packageNames)
             {
                 // reset config each time through
                 config = originalConfig.deep_copy();
 
                 var installedPackage = packageManager.LocalRepository.FindPackage(packageName);
+                if (installedPackage == null)
+                {
+                    notFoundedMessage.AppendLine("Package {0} was not founded.".format_with(packageName));
```
ferventcoder, line 906: Typo at  "not founded", should be "not found". However, this doesn't come back like we expect. If someone is calling with limit output, we need to adjust to only provide the right output. We also have a particular exit code we set for enhanced exit codes when things are not found.

ferventcoder:
> The changed code enables the outdated command to run against one specific package.

It should actually enable `choco outdated git 7zip chocolatey`, not just one specific package.

src/chocolatey/infrastructure.app/services/NugetService.cs
cuichenli: Sorry I failed to find any message on not found except the one for `not found in source`, is that the one you are referring to? Otherwise, can you guide me where to find it? Thanks!

src/chocolatey/infrastructure.app/services/NugetService.cs
cuichenli: Thanks! One question, if the user specified `ignore-unfound`, should we still use the exit code you mentioned? Thanks!

gep13:
@cuichenli I wanted to follow up regarding this PR, are you in a position to address some of the comments from the review, and also to sign the CLA?

Let us know if you have any questions.

PR #2063, opened 2020-06-08
(doc) Changed cla links from clahub to cla-assistant

AdmiringWorm:
Previously the contribution documentation linked to clahub
which was used as a way for contributors to sign the CLA.
This PR changes this to cla-assistant which is now
being used.

Unfortunately, this also meant removing the link regarding what a CLA is, and why anyone should care as I could not find any equivalent page for this.


ferventcoder:
Thanks for making that adjustment!

PR #2064, opened 2020-06-08
(GH-1364) Template create .nuspec encoded without BOM

### LINKED ISSUE #1364: Choco new should output a .nuspec in UTF-8 without a byte order mark

TheCakeIsNaOH:
Fixes #1364 

Changes the encoding of `.nuspec` files created from templates to UTF8 with BOM to UTF8 without BOM according to best practices in the [Creating Chocolatey Packages](https://chocolatey.org/docs/create-packages#character-encoding) documentation. Applies to both the default template and to custom templates.

Adjusted template services tests to include UTF8 no BOM files, otherwise nuspec files would not be detected anymore.

I moved only the nuspec over to no BOM for both the default template and custom templates. The other `.txt` files from the default template could be moved over to no BOM very easily if that is wanted.

### DIFF
--- a/src/chocolatey.tests/infrastructure.app/services/TemplateServiceSpecs.cs
+++ b/src/chocolatey.tests/infrastructure.app/services/TemplateServiceSpecs.cs
@@ -231,6 +231,7 @@ namespace chocolatey.tests.infrastructure.app.services
             private readonly ChocolateyConfiguration config = new ChocolateyConfiguration();
             private readonly List<string> files = new List<string>();
             private readonly HashSet<string> directoryCreated = new HashSet<string>(StringComparer.InvariantCultureIgnoreCase);
+            private readonly UTF8Encoding utf8WithoutBOM = new UTF8Encoding(false);
 
             public override void Context()
             {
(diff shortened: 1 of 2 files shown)


AdmiringWorm:
> I moved only the nuspec over to no BOM for both the default template and custom templates. The other .txt files from the default template could be moved over to no BOM very easily if that is wanted.

I would agree with moving .txt files to be created without BOM as well. Honestly, all files not requiring BOM should be changed to not use it, and I think only PowerShell requires BOM (as it will otherwise be read with the encoding `Windows-1252` by powershell.exe).
But that is just my opinion, and it would be up to Rob if that should be changed or not.

TheCakeIsNaOH:
> Honestly, all files not requiring BOM should be changed to not use it, and I think only PowerShell requires BOM (as it will otherwise be read with the encoding Windows-1252 by powershell.exe).

For the default template, yes this makes sense IMO.

I would be hesitant to change the non-PowerShell files in custom templates over to default to no BOM.

AdmiringWorm:
> I would be hesitant to change the non-PowerShell files in custom templates over to default to no BOM.

From my own experience, I have had more problems with programs when BOM has been added for files they didn't expect than the other way around.

TheCakeIsNaOH:
Man, I guess I have to everything in git directly. I *thought* github desktop was wrapping correctly.

TheCakeIsNaOH:
@ferventcoder Bump, ready for a re-review.

gep13 (approved):
LGTM!

gep13:
@TheCakeIsNaOH thanks for getting this PR put together, really appreciate it!

TheCakeIsNaOH:
@gep13 Thanks for the reviews and merges.

Issue #2074, opened 2020-07-02
Control shim generation using a .shiminclude file


johnstevenson:
The following proposal is a simple way for package authors to specify which executables to create a shim for. It removes the need to create .ignore files to prevent an exe being shimmed and allows batch files (.bat and .cmd) to be shimmed without having to use `Install-BinFile`. It is completely compatible with the current behaviour.

## .shiminclude files
A `.shiminclude` file specifies executables that Chocolatey should generate a shim for. This mechanism provides an opt-in alternative to the default behaviour, which is to automatically shim all .exe files found in the package unless there is a matching file suffixed .ignore.

Each line in a .shiminclude file specifies a path relative to the package's `tools` directory. This can point to either a specific file type (.exe, .bat or .cmd), or to a directory or group of directories in which to search for .exe files.

### Format
- A `.shiminclude` file is UTF-8 encoded without a BOM.

- A blank line matches no directories or files.

- A line starting with `#` serves as a comment. Put a backslash `\` in  front of the first hash for paths that begin with a hash.

- An optional prefix `!` negates the path. Put a backslash `\` in front of the first `!` for paths that begin with a literal `!`.

- A backslash `\` or a forward slash `/` can be used as a directory separator.

- The `?` wildcard matches zero or one character except a directory separator.

- The `*` wildcard matches zero or more characters except a directory  separator.

- The order of entries is not considered: duplicates are skipped and negated paths are irrevocable.

### Usage
- A `.shiminclude` file must be located in the package's `tools` directory and its entries must be relative paths from this location. If an entry resolves to a location outside the package directory it will be skipped. If no .shiminclude file is found, Chocolatey will search for .exe files to shim.

- Chocolatey will not search for .exe files to shim if the .shiminclude file is empty, or contains only blank/comment entries.

- An entry is treated as a path to a file if it ends with an `.exe`, `.bat` or `.cmd` extension, otherwise it is taken as a path to
a directory.

- If a matching file suffixed `.ignore` is found, a shim will not be generated. 

### Examples
The following examples assume that a package has installed its software into the `<pkg>\tools\prog` directory, where `<pkg>` is the location of the package. The .shiminclude file is located at `<pkg>\tools\.shiminclude`.

- `prog\sbin\prog.exe` matches the `<pkg>\tools\prog\sbin\prog.exe` file.

- `prog\sbin` matches all .exe files in `<pkg>\tools\prog\sbin`.

[...]

PR #2075, opened 2020-07-02
(GH-2074) Control shim generation with a .shiminclude file

johnstevenson:
A .shiminclude file specifies executables that Chocolatey should
generate a shim for. This mechanism provides an opt-in alternative to
the default behaviour, which is to automatically shim all .exe files
found in the package unless there is a matching file suffixed .ignore.

Each line in a .shiminclude file specifies a path relative to the
package's tools directory. This can point to either a specific file
type (.exe, .bat or .cmd), or to a directory or group of directories in
which to search for .exe files.

Format:
- A .shiminclude file is UTF-8 encoded without a BOM.

- A blank line matches no directories or files.

- A line starting with `#` serves as a comment. Put a backslash `\` in
  front of the first hash for paths that begin with a hash.

- An optional prefix `!` negates the path. Put a backslash `\` in front
  of the first `!` for paths that begin with a literal `!`.

- A backslash `\` or a forward slash `/` can be used as a directory
  separator.

- The `?` wildcard matches zero or one character except a directory
  separator.

- The `*` wildcard matches zero or more characters except a directory
  separator.

- The order of entries is not considered: duplicates are skipped and
  negated paths are irrevocable.

Usage:
- A .shiminclude file must be located in the package's tools directory
  and its entries must be relative paths from this location. If an entry
  resolves to a location outside the package directory it will be
  skipped. If no .shiminclude file is found, Chocolatey will search for
  .exe files to shim.

- Chocolatey will not search for files to shim if the .shiminclude file
  is empty, or contains only blank/comment entries.

- An entry is treated as a path to a file if it ends with an .exe, .bat
  or .cmd extension, otherwise it is considered a path to a directory.

- If a matching file suffixed .ignore is found, a shim will not be
  generated.

Examples:
The following examples assume that a package has installed its software
into the `<pkg>\tools\prog` directory, where `<pkg>` is the location of
the package. The shiminclude file is located at `<pkg>\tools\.shiminclude`.

- `prog\sbin\prog.exe` matches the `<pkg>\tools\prog\sbin\prog.exe` file.

- `prog\sbin` matches all .exe files in `<pkg>\tools\prog\sbin`.

- `prog\*` matches all .exe files in all top level directories in
  `<pkg>\tools\prog`.

- `prog\sbin\prog?.exe` matches `<pkg>\tools\prog\sbin\prog.exe` and
   `<pkg>\tools\prog\sbin\prog2.exe`.

- The file extension must be included to match other file types. For
  example `prog\sbin\prog3.bat` matches the `<pkg>\tools\prog\sbin\prog3.bat`
  file, while `prog\*\*.bat` matches all .bat files in all top level
  directories in `<pkg>\tools\prog`.

### DIFF
--- a/src/chocolatey.tests.integration/chocolatey.tests.integration.csproj
+++ b/src/chocolatey.tests.integration/chocolatey.tests.integration.csproj
@@ -99,6 +99,7 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="infrastructure.app\services\FilesServiceSpecs.cs" />
+    <Compile Include="infrastructure.app\shimtarget\ShimTargetSpecs.cs" />
     <Compile Include="infrastructure\commands\CommandExecutorSpecs.cs" />
     <Compile Include="infrastructure\cryptography\CrytpoHashProviderSpecs.cs" />
     <Compile Include="infrastructure\filesystem\DotNetFileSystemSpecs.cs" />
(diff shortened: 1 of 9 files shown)


ferventcoder:
Thanks @johnstevenson - will review this in the next few weeks (definitely before we release the next version of choco!).

johnstevenson:
@ferventcoder That's good to know (it sounds like you might like it!).

PR #2093, opened 2020-08-07
(GH-2092) Take all registry keys into account

### LINKED ISSUE #2092: When reporting installed applications, not all currently installed packages are correctly removed from list
It is possible that a package, which installs an application into Add/Remove programs, is installed, however, when running `choco list -li` the installed application is still listed.

When a Chocolatey package is installed, the associated installation application, should _not_ be included in this output.  This is an indication that this application is under the management of Chocolatey.

gep13:
It is possible that the installation of an application can generate
multiple registry keys within the snapshot file that is craeted.  Rather
than looking at only the first registry key when deciding if an installed
application is covered by a Chocolatey package, look at all keys within
the snapshot file.

Fixes #2092


gep13:
@ferventcoder as requested, this have been rebased onto the head of the stable branch.

Issue #2083, opened 2020-07-20
Error was 'Cannot process argument transformation on parameter 'Encoding'. 'byte' is not a supported encoding name.


doggy8088:
### What You Are Seeing?

When I install Chocolatey using PowerShell Core, I get an warning below:

> WARNING: Unable to add Chocolatey to the profile. You will need to do it manually. Error was 'Cannot process argument transformation on parameter 'Encoding'. **'byte' is not a supported encoding name**. For information on defining a custom encoding, see the documentation for the Encoding.RegisterProvider method. (Parameter 'name')'

### What is Expected?

I expected to install the following script into `$PROFILE`.

```ps1
# Chocolatey profile
$ChocolateyProfile = "$env:ChocolateyInstall\helpers\chocolateyProfile.psm1"
if (Test-Path($ChocolateyProfile)) {
  Import-Module "$ChocolateyProfile"
}
```

### How Did You Get This To Happen? (Steps to Reproduce)

1. Install PowerShell Core v7 on Windows 10 (2004)

2. Create `$env:USERPROFILE\Documents\PowerShell` folder

```ps1
mkdir $env:USERPROFILE\Documents\PowerShell
```

3. Run the following script:

    ```ps1
    Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))
    ```

### Output Log

<details>
<summary>Full Log Output</summary>

```
PS C:\> Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))
Getting latest version of the Chocolatey package for download.
Getting Chocolatey from https://chocolatey.org/api/v2/package/chocolatey/0.10.15.
Extracting C:\Users\will\AppData\Local\Temp\chocolatey\chocInstall\chocolatey.zip to C:\Users\will\AppData\Local\Temp\chocolatey\chocInstall...
Installing chocolatey on this machine
Creating ChocolateyInstall as an environment variable (targeting 'Machine')
  Setting ChocolateyInstall to 'C:\ProgramData\chocolatey'
[...]

PR #2119, opened 2020-09-18
Change Get-Content $Path -Encoding byte to Get-Content $Path -AsByteS…

tonyping:
…tream for compatibility with Powershell Core

### DIFF
--- a/nuget/chocolatey/tools/chocolateysetup.psm1
+++ b/nuget/chocolatey/tools/chocolateysetup.psm1
@@ -549,7 +549,7 @@ param(
 
 # Adapted from http://www.west-wind.com/Weblog/posts/197245.aspx
 function Get-FileEncoding($Path) {
-    $bytes = [byte[]](Get-Content $Path -Encoding byte -ReadCount 4 -TotalCount 4)
+    $bytes = [byte[]](Get-Content $Path -AsByteStream -ReadCount 4 -TotalCount 4)
 
     if(!$bytes) { return 'utf8' }
 
--- a/nuget/chocolatey/tools/chocolateysetup.psm1
+++ b/nuget/chocolatey/tools/chocolateysetup.psm1
@@ -549,7 +549,14 @@ param(
 
 # Adapted from http://www.west-wind.com/Weblog/posts/197245.aspx
 function Get-FileEncoding($Path) {
+  if ($PSVersionTable.PSVersion.Major -lt 6) {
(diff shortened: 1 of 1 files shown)


vexx32:
Given that the newer syntax is completely incompatible with older versions of PowerShell (which choco still supports), we'd need to manage this with an if/else based on the PowerShell version; we can't simply use newer syntaxes at the present without making sure older versions will still work.

tonyping:
Ah sorry, I didn't know that. So would doing a simple version check and then selecting which parameter would work (-Encoding Byte for older, or -AsByteStream for newer) be enough, or are there any further consideration that Powershell Core needs to be fully compatible with the installer? 

Because I saw this[issue](https://github.com/chocolatey/choco/issues/2083), that mentions that Core does not have OS environment variables, but I'm not really sure what the replier meant by that.

vexx32:
Yeah I've no idea what they mean by that, but I think they misunderstood the issue / underlying problem.

I would... hmm. Yeah, probably a version gated check for anything less than 6 should use the old form and otherwise use the new switch is probably the safest bet. 🙂

gep13:
@vexx32 are you in a position to re-review this?

@tonyping was an issue created that this PR can be linked to?

nuget/chocolatey/tools/chocolateysetup.psm1
```diff
@@ -549,7 +549,14 @@ param(
...
+  if ($PSVersionTable.PSVersion.Major -lt 6) {
+    Write-Debug "Detected Powershell version < 6 ; Using -Encoding byte parameter"
     $bytes = [byte[]](Get-Content $Path -Encoding byte -ReadCount 4 -TotalCount 4)
+  }
+  else {
+    Write-Debug "Detected Powershell version >= 6 ; Using -AsByteStream parameter"
+    $bytes = [byte[]](Get-Content $Path -AsByteStream -ReadCount 4 -TotalCount 4)
+    }
 
     if(!$bytes) { return 'utf8' }
```
vexx32, line 561: This looks good to me, apart from the indentation seemingly not matching the rest of the function here. Can you bump the indentation so it matches the rest of this function?

vexx32 (approved):
Looks good to me with one minor formatting tweak.

gep13:
I am going to close this PR, as this has been updated in a new PR here:

https://github.com/chocolatey/choco/pull/2416

Let's use that one going forward.  Correct attribution will still be applied in the git commits that are merged into the develop branch.

PR #2129, opened 2020-10-02
(#2049) Note --version as a switch in the help menu

### LINKED ISSUE #2049: Include information about --version option in choco help documentation
Ensure when running the top level help that `--version` is mentioned as an option for outputting the version information.

Reference: https://twitter.com/hunterz0/status/1261487482006315018?s=21

DeeHechEss:
Resolves #2049 

I added the --version switch to the help menu.
I'm not very familiar with C# but digging through the code it seems the version doesn't need any config options to be set, so I put an empty function in the option_set not sure if that's the right way of doing it.

I tried adding the -v shorthand but it complained about duplicate key, as it is also used for --verbose. But in ConfigurationOptions.cs on line 105 it looks like the shorthand can also be used for --version.

### DIFF
--- a/src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
+++ b/src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
@@ -355,7 +355,7 @@ namespace chocolatey.infrastructure.app.builders
                         .Add("trace",
                              "Trace - Show trace messaging. Very, very verbose trace messaging. Avoid except when needing super low-level .NET Framework debugging. Available in 0.10.4+.",
                              option => config.Trace = option != null)
-                        .Add("v|version",
+                        .Add("version",
                              "Version - Prints out the Chocolatey version.",
                              option => {})
                         .Add("nocolor|no-color",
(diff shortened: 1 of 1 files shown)


src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
```diff
@@ -355,6 +355,9 @@ private static void set_global_options(IList<string> args, ChocolateyConfigurati
                         .Add("trace",
                              "Trace - Show trace messaging. Very, very verbose trace messaging. Avoid except when needing super low-level .NET Framework debugging. Available in 0.10.4+.",
                              option => config.Trace = option != null)
+                        .Add("version",
+                             "Version - Prints out the Chocolatey version.",
+                             option => {})
```
ferventcoder, line 360: IIRC this is may conflict with install / upgrade / uninstall / search as is. It would be worth checking if the `version` of those just nicely overrides this or not.

gep13:
@DeeHechEss thank you very much for taking the time to create this PR, we really appreciate it!

Based on the feedback here, and some internal discussion, I have taken the liberty to make a commit here: https://github.com/chocolatey/choco/commit/e8f76b6cb5352c976a837611e8bfd8e845794490 which adds in the required information.

I would normally have added this to the PR branch, and merge the PR directly, however, the fact that you created a PR directly from the master branch makes this difficult.  As a result, I am going to close this PR, but I have added attribution to the commit to show that you helped with this work.

Thanks again for your help!

PR #2130, opened 2020-10-03
feature: Add automatic process elevation

tapika:
This is automatic elevation request for chocolatey.

Feature itself more or less manual, and cannot be tested automatically - that's why did not any tests on this one.
The advantage of automatic elevation is that you can automatically install without need to run another choco instance,
disadvantage - is that console is opened in new window, and that might exit before user being able to see the output.

This is my first pull request from series, I hope that we can agree / discuss on features / functionalities which will be present in next pull request chain - see also my mail sent to you previously.


gep13:
@tapika thank you for taking the time to raise this PR, and for your input in other issues for Chocolatey CLI.

At this current time, we don't have any plans to provide a mechanism to do automatic process elevation.  Chocolatey CLI works the way it does, since it needs to interact with a number of different systems, including other Chocolatey products like Chocolaty Agent and Chocolatey Central Management.  Any changes to the elevation abilities of choco.exe would need to be done in conjunction with work in those other products.

With this in mind, I am going to go ahead and close this issue.  If you wanted to discuss this further, please feel free to respond here, or create a [discussion](https://github.com/chocolatey/choco/discussions).

Issue #2044, opened 2020-05-06
Temp location can change unexpectedly


teknowledgist:
### What You Are Seeing?

The value of `$env:Temp` when running as SYSTEM changes for the session in which `Get-ToolsLocation` is first run.  (*I encountered this problem with `Get-ToolsLocation` and wrote most of this around that, but really any other script/function that calls `Update-SessionEnvironment` will create the same problem.*)

Running as SYSTEM, the `$env:Temp` variable is normally `C:\Windows\Temp`.  If the first package to ever use `Get-ToolsLocation` happens to be running as SYSTEM, the location of `$env:Temp` changes to `C:\Windows\system32\config\systemprofile\AppData\Local\Temp` for the duration of the session.  This appears to cause some packages to fail (for unknown reasons).  

Future sessions have the correct `$env:Temp` regardless of calls to `Get-ToolsLocation`.

### What is Expected?

`$env:Temp` shouldn't change (especially in the middle of the ChocolateyInstall.ps1) unless it is requested by the user.

### How Did You Get This To Happen? (Steps to Reproduce)

Everyday users won't be running Chocolatey as SYSTEM, but most centralized desktop management software will.  To test this manually requires opening cmd or PoSh windows as SYSTEM as follows:
1. Start with a clean Win10 (snapshotted) VM
2. Install Chocolatey
3. Install [pstools](https://chocolatey.org/packages/pstools) (or [runassystem](https://chocolatey.org/packages/runassystem)) package
4. From an elevated command prompt, run `pstools /s /i cmd.exe` (or `pstools /s /i powershell`).  (If using runassystem, double click and enter either cmd.exe or powershell.)

From the new console window now running as SYSTEM, you can attempt to install the [anaconda3](https://chocolatey.org/packages/anaconda3) package (`choco install anaconda3`).  It will fail with the new path displaying.  (If you try it again, in a new window -- as SYSTEM -- it will work.) (There may be other packages broken by this, but it would be a PITA to test.)

Easier to see the change is to reset the system and:

1) Install the [chocolateypowershell](https://chocolatey.org/packages/chocolateypowershell) utility package
2) Open the *chocolateypowershell* console (as SYSTEM), check the `$env:temp` path, run `Get-ToolsLocation` and again check `$env:temp`.  The output of that is as follows:

    [ChocolateyPS] C:\> $env:temp
    C:\Windows\TEMP
    [ChocolateyPS] C:\> get-toolslocation
    C:\tools
[...]

PR #2154, opened 2020-11-27
(GH-2044) Fix for changing $env:Temp

teknowledgist:
(GH-2044) This fixes a problem when a package is installed under the System credentials (typically how centralized management software -- e.g. SCCM -- operate):

Running as `SYSTEM`, the `$env:Temp` variable is normally *C:\Windows\Temp*. If the **first** package to ever use `Update-SessionEnvironment.ps1` happens to be running as `SYSTEM`, the location of `$env:Temp` changes to *C:\Windows\system32\config\systemprofile\AppData\Local\Temp* for the duration of the session.  This change to `$env:Temp` does not occur in subsequent calls to `Update-SessionEnvironment.ps1`.  It is unknown why this change occurs at all and likewise is unknown why it occurs for only the first time.

The `Update-SessionEnvironment.ps1` normally gives priority to user-scope environment variables over machine-scope environment variables.  Changes in this commit eliminates the gathering of the `User` scope environment variables when running as the `SYSTEM` account on the belief that reason to be operating under the `SYSTEM` account is to do machine-wide activities, and there is no reason to see `SYSTEM` as a "user".  On a clean system, the only User-scope environment variables under `SYSTEM` are *path*, *temp*, and *tmp* which are already defined under the Machine-scope, and in general, there shouldn't be any new User-scope variables for the SYSTEM account because it's not a "normal" account.  Basically, there should be little reason to expect this change to break anything.

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Update-SessionEnvironment.ps1
+++ b/src/chocolatey.resources/helpers/functions/Update-SessionEnvironment.ps1
@@ -80,10 +80,10 @@ None
 
   #Path gets special treatment b/c it munges the two together
   $paths = 'Machine', 'User' |
-    % {
+    ForEach-Object {
       (Get-EnvironmentVariable -Name 'PATH' -Scope $_) -split ';'
     } |
-    Select -Unique
+    Select-Object -Unique
   $Env:PATH = $paths -join ';'
 
   # PSModulePath is almost always updated by process, so we want to preserve it.
--- a/src/chocolatey.resources/helpers/functions/Update-SessionEnvironment.ps1
+++ b/src/chocolatey.resources/helpers/functions/Update-SessionEnvironment.ps1
@@ -65,15 +65,18 @@ None
   $architecture = $env:PROCESSOR_ARCHITECTURE
   $psModulePath = $env:PSModulePath
 
-  #ordering is important here, $user comes after so we can override $machine
-  'Process', 'Machine', 'User' |
-    % {
-      $scope = $_
-      Get-EnvironmentVariableNames -Scope $scope |
-        % {
-          Set-Item "Env:$($_)" -Value (Get-EnvironmentVariable -Scope $scope -Name $_)
+  #ordering is important here, $user should override $machine...
+  $ScopeList = 'Process', 'Machine'
+  if ($userName -notin 'SYSTEM', "${env:COMPUTERNAME}`$") {
+    # but only if not running as the SYSTEM/machine in which case user can be ignored.
+    $ScopeList += 'User'
+  }
+  foreach ($Scope in $ScopeList) {
+    Get-EnvironmentVariableNames -Scope $Scope |
+        ForEach-Object {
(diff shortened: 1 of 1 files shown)


gep13:
@vexx32 any chance you can take a look at this one, thanks?

src/chocolatey.resources/helpers/functions/Update-SessionEnvironment.ps1
```diff
@@ -56,31 +56,34 @@ None
...
-  #ordering is important here, $user comes after so we can override $machine
-  'Process', 'Machine', 'User' |
-    % {
-      $scope = $_
-      Get-EnvironmentVariableNames -Scope $scope |
-        % {
-          Set-Item "Env:$($_)" -Value (Get-EnvironmentVariable -Scope $scope -Name $_)
+  #ordering is important here, $user should override $machine...
+  $Scopes = 'Process', 'Machine'
+  if (($userName -ne 'SYSTEM') -and ($userName -ne "$($env:COMPUTERNAME)`$")) {
```
vexx32, line 70: This might be clearer to write as:
```suggestion
  if ($userName -notin 'SYSTEM', "${env:COMPUTERNAME}`$") {

```

src/chocolatey.resources/helpers/functions/Update-SessionEnvironment.ps1
```diff
@@ -56,31 +56,34 @@ None
...
+    # but only if not running as the SYSTEM/machine in which case user can be ignored.
+    $Scopes += 'User'
+  }
+  foreach ($Scope in $Scopes) {
```
vexx32, line 74: I generally recommend avoiding plural variables be used alongside singular counterparts; it's very easy to mess things up with a single-character typo and not notice that the logic is totally broken until later. Maybe use `$ScopeList` instead for the latter variable?

src/chocolatey.resources/helpers/functions/Update-SessionEnvironment.ps1
```diff
@@ -56,31 +56,34 @@ None
...
+    Get-EnvironmentVariableNames -Scope $Scope |
+        ForEach-Object {
+          Set-Item "Env:$($_)" -Value (Get-EnvironmentVariable -Scope $Scope -Name $_)
```
vexx32, line 77: Unnecessary subexpression wrapper in this string
```suggestion
          Set-Item "Env:$_" -Value (Get-EnvironmentVariable -Scope $Scope -Name $_)
```

vexx32 (changes requested):
Overall looks good to me. A few things here and there I think we should clean up a bit before we merge. 🙂

teknowledgist:
I agree with all the comments by @vexx32 and I *think* I've made the adjustments so they roll into this PR.  (I'm not very good with GitHub and how the change management flow works.)

vexx32 (approved):
Looks good to me, thanks for fixing that up! 💖

gep13 (approved):
LGTM!

gep13:
@teknowledgist thank you very much for taking the time to submit this PR, we really appreciate it!

PR #2171, opened 2021-01-11
Fix date format from "Sep 09 2020 18:01:35" to "Mon 2021-01-11 11:30:41"

ypid-geberit:
In [other places of the docs](https://docs.chocolatey.org/en-us/features/package-audit) you refer to ISO 8601 yourself. I don’t suggest anything fancy here like read the preferred date format of the user from the OS. Instead I am proposing to just change the format string to a more sensible one.

The weekday at first is outside of the spec of both ISO 8601 and RFC 3339 but can be a nice addition (and is common https://www.freedesktop.org/software/systemd/man/systemd.time.html#Displaying%20Timestamps).

Ref: https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings
Ref: https://tools.ietf.org/html/rfc3339

PS: I assume this (https://github.com/chocolatey/choco/blob/master/CONTRIBUTING.md#prerequisites)

> Ensure you have signed the Contributor License Agreement (CLA) - without this we are not able to take contributions that are not trivial.

Is true here. Signing the CLA for this trivial change is too much. Just to avoid any issue here, in case you think any copyright applies to my change (which I don’t think so), I license it under https://creativecommons.org/publicdomain/zero/1.0/deed.en.


gep13:
@ypid-geberit apologies for not coming back to you sooner about this.

I am really not sure if I agree with this change, if anything, I would argue that we should modify the other locations, since in my opinion 2020-09-09 is far more confusing than Sep 09 2020.

If you would like to bring this up again, can I first suggest that you open an issue where we can discuss the problem that you are seeing, but you suggest to fix the issue, so that we can get agreement on a path forward, before any changes are made.

Thanks

PR #2182, opened 2021-01-21
Add feature of "Manual install by user" if auto install fails for any reason.

umarabdullahmansoori:
Chocolatey packages fails many times due to different reasons, therefor anyone cannot fully rely on them. If package installation fails, user download same program from google and install it manually. 

So, if chocolatey fails to install package, it run the downloaded exe file itself so user will manually install it easily (instead of going to web and download it).


pauby:
@umarabdullahmansoori I'm assuming this was a mistake, and you meant to merge other changes?

Issue #886, opened 2016-08-01
Packages.config should include all options (including ignorechecksums)


cori:
### What You Are Seeing?

cinst on command-line using a .config file disregards ignorechecksum for specific packages.
### What is Expected?

an entry in a .config file that includes `ignorechecksums="true" or another of the aliases from https://github.com/chocolatey/choco/wiki/CommandsInstall#options-and-switches should honor the ignorechecksums directive and allow, for that specific package,an installer that fails the checksums verification to be installed.
### How Did You Get This To Happen? (Steps to Reproduce)

ran `cinst rt.config -y` using the attached rt.config.txt file.
[rt.config.txt](https://github.com/chocolatey/choco/files/394245/rt.config.txt)
### Output Log

chocolatey.log: https://gist.github.com/cori/01ff60e1638b75f38a25cbe41e7a0a16
choco.summary.log: https://gist.github.com/cori/b51791c5e22e19180d1e31e0eee717f1

cori:
On a side note, running `cinst rt.config -y --ignore-checksums` with a .config file with checksum-failing packages *does* enable them to be installed,but in a scripted install environment that's unsatisfactory - I need to be able to control when to ignore checksums on a package-by-package basis so that I'm only ignoring them in cases that I've investigated, not for all the packages I'm installing arbitrarily,which would be pretty insecure.

cori:
I guess I wouldn't have thought of this as an enhancement, but maybe the problem is that https://github.com/chocolatey/choco/wiki/CommandsInstall#packagesconfigdoesn't call out which options work in a .config file and which ones don't - without some clarity there it seems like a sensible assumption that all of the options listed in the following section should be expected to work as advertised.

Obviously, though, you need to manage your workload as you see fit. Thanks for the swift attention.

ferventcoder:
@cori As I mentioned over in the disqus notes, I think this should include all options that can be passed.

ferventcoder:
> I guess I wouldn't have thought of this as an enhancement

@cori I could go either way. Unexpected behavior could make it a bug, but as we did not add that functionality yet I think of it as an enhancement. Forgotten functionality could make it a bug. Like I said, I could go either way. 👍

cori:
don't see a note on the disqus thread, but probably more important to get it here!

ferventcoder:
yes, more important to get it here. Perhaps it wasn't on disqus. Where did we talk about this already?

cori:
[...]

PR #2187, opened 2021-01-27
(#886) added support for force and timeout in packages.config file

boussaffawalid:
Added support to timeout and force in packages.config

This is now supported
```
<package id="packagename" timeout="1000" force="true" />
```

Relates to #886.

Other PR's will be required to bring in the other options that can then be controlled via the packages.config file.

### DIFF
--- a/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
+++ b/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
@@ -737,6 +737,8 @@ Would have determined packages that are out of date based on what is
                     if (pkgSettings.ApplyPackageParametersToDependencies) packageConfig.ApplyPackageParametersToDependencies = true;
                     SourceType sourceType;
                     if (Enum.TryParse(pkgSettings.Source, true, out sourceType)) packageConfig.SourceType = sourceType;
+                    if (pkgSettings.Force) packageConfig.Force = true;
+                    packageConfig.CommandExecutionTimeoutSeconds = pkgSettings.CommandExecutionTimeoutSeconds == -1 ? packageConfig.CommandExecutionTimeoutSeconds : pkgSettings.CommandExecutionTimeoutSeconds;
 
                     this.Log().Info(ChocolateyLoggers.Important, @"{0}".format_with(packageConfig.PackageNames));
                     packageConfigs.Add(packageConfig);
--- a/src/chocolatey/infrastructure.app/commands/ChocolateyInstallCommand.cs
+++ b/src/chocolatey/infrastructure.app/commands/ChocolateyInstallCommand.cs
@@ -370,7 +370,7 @@ NOTE: The filename is only required to end in .config, the name is not required 
       <package id=""alloptions"" version=""0.1.1""
                source=""https://somewhere/api/v2/"" installArguments=""""
                packageParameters="""" forceX86=""false"" allowMultipleVersions=""false""
-               ignoreDependencies=""false""
+               ignoreDependencies=""false"" commandExecutionTimeoutSeconds=""1000"" force=""false""
                />
     </packages>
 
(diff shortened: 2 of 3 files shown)


src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
```diff
@@ -722,6 +722,8 @@ private IEnumerable<ChocolateyConfiguration> get_packages_from_config(string pac
                     if (pkgSettings.IgnoreDependencies) packageConfig.IgnoreDependencies = true;
                     if (pkgSettings.ApplyInstallArgumentsToDependencies) packageConfig.ApplyInstallArgumentsToDependencies = true;
                     if (pkgSettings.ApplyPackageParametersToDependencies) packageConfig.ApplyPackageParametersToDependencies = true;
+                    if (pkgSettings.Force) packageConfig.Force = true;
+                    packageConfig.CommandExecutionTimeoutSeconds = string.IsNullOrWhiteSpace(pkgSettings.Timeout.ToString()) ? packageConfig.CommandExecutionTimeoutSeconds : pkgSettings.Timeout;
```
AdmiringWorm, line 726: Just as a note, this won't work as you expect it though.
The `Timeout` even when converted to a string will never be null or empty.
The default of an integer in `0` and not empty (even when converted to a string).

ferventcoder:
Hi @boussaffawalid! Thanks so much for taking the time to improve Chocolatey. It looks like a really good start! 

Note that @AdmiringWorm dropped a note in on some work to be done with the code, I'll note that the commit message/body needs to talk about the why of the change. If you haven't read over CONTRIBUTING, please take a look at that to allow this to be easier for us to incorporate the change without a lot of back and forth - especially this section: https://github.com/chocolatey/choco/blob/master/CONTRIBUTING.md#prepare-commits. 

Once you've fixed up those sections, also take a look at the rest of the document as it details how to push back the changes on the same branch/commit. Let us know if you have any questions. 👍

docs/generated/CommandsInstall.md
```diff
@@ -127,7 +127,7 @@ Alternative to PackageName. This is a list of packages in an xml manifest for Ch
       <package id="alloptions" version="0.1.1"
                source="https://somewhere/api/v2/" installArguments=""
                packageParameters="" forceX86="false" allowMultipleVersions="false"
-               ignoreDependencies="false"
+               ignoreDependencies="false" timeout="1000" force="false"
```
gep13, line 130: @boussaffawalid this file is automatically generated by another part of the Chocolatey, and shouldn't be updated directly.  As you will hopefully notice, this file no longer exists in this repository, as it was recently removed.  The location where you would want to update this would be here:

https://github.com/chocolatey/choco/blob/master/src/chocolatey/infrastructure.app/commands/ChocolateyInstallCommand.cs#L368

Once changed there, the next time we generate the docs, this section would be updated in the generated markdown files.

gep13 (changes requested):

src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
boussaffawalid: fixed, I set a default value for pkgSettings.Timeout and used it for comparison. I used -1 as default value is 0 is a valid timeout.

AdmiringWorm: It looks like you made a small mistake, unfortunately.
You set the default on the `Force` Parameter instead of the `Timeout` parameter.

boussaffawalid: @AdmiringWorm  oops that was a mistake, it should be fixed now.

gep13 (approved):
LGTM!

gep13:
The failing build here is due to a known issue with the build running on Ubuntu, so for now, I am going to move forward with the merge.

gep13:
@boussaffawalid thank you for taking the time to raise this PR, and congratulations on your first time contribution to Chocolatey! We really appreciate your help here!

PR #2204, opened 2021-02-23
(GH-2203) Get-ChocolateyUnzip add unzipLocation as an alias to Destination

### LINKED ISSUE #2203: Get-ChocolateyUnzip - Support UnzipLocation as an alias to Destination
It is really annoying when I forget which helper has which argument, and I have to go back and fix it.

`Install-ChocolateyZipPackage` already has a `Destination` alias to `UnzipLocation`.

It would be really nice if `Get-ChocolateyUnzip` gets the same treatment, with `UnzipLocation` as an alias to `Destination`.

TheCakeIsNaOH:
Fixes #2203

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
+++ b/src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
@@ -88,7 +88,7 @@ Install-ChocolateyZipPackage
 #>
 param(
   [alias("file")][parameter(Mandatory=$false, Position=0)][string] $fileFullPath,
-  [parameter(Mandatory=$true, Position=1)][string] $destination,
+  [alias("unzipLocation")][parameter(Mandatory=$true, Position=1)][string] $destination,
   [parameter(Mandatory=$false, Position=2)][string] $specificFolder,
   [parameter(Mandatory=$false, Position=3)][string] $packageName,
   [alias("file64")][parameter(Mandatory=$false)][string] $fileFullPath64,


ferventcoder:
Howdy @TheCakeIsNaOH ! This pull request looks fine for code, however the commit needs a bit more. Have you read over the CONTRIBUTING?

Even though this isn't much more than adding an alias, the commit message body should have all of the reasoning in it so when we look back in three years (and we do this all the time) we'll understand why the change was made.

ferventcoder:
Specifically this section - https://github.com/chocolatey/choco/blob/master/CONTRIBUTING.md#prepare-commits

TheCakeIsNaOH:
@ferventcoder fixed.

src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
```diff
@@ -88,7 +88,8 @@ Install-ChocolateyZipPackage
 #>
 param(
   [alias("file")][parameter(Mandatory=$false, Position=0)][string] $fileFullPath,
-  [parameter(Mandatory=$true, Position=1)][string] $destination,
+  [parameter(Mandatory=$true, Position=1)]
+  [alias("unzipLocation")][string] $destination,
```
gep13, line 92: @TheCakeIsNaOH is there a reason that you chose to break this parameter onto two lines? The other parameters in this function are all on a single line, with:

`alias - parameter - type - name`

I think it would make sense for this to remain the same format.  Thoughts?

gep13 (changes requested):

src/chocolatey.resources/helpers/functions/Get-ChocolateyUnzip.ps1
TheCakeIsNaOH: In `Install-ChocolateyZipPackage`, it was broken up onto two lines, so I assumed that was the format to use.

I can switch it to being on one line.

TheCakeIsNaOH: Done.

gep13: Gotcha.  Depending on the history of any given function, the formatting may have changed, but within any given file I think it makes sense to keep the formatting consistent.  Thanks for getting that fixed up!

gep13 (approved):
LGTM!

gep13:
@TheCakeIsNaOH thanks for getting this added!  Once the CI builds are finished we can get this merged.

PR #2207, opened 2021-03-01
(doc) Update to prefer # rather than GH-

gep13:
GH- is specific to GitHub, where as, # works across different version
control systems, so it makes sense to get into the habit of using
something that works everywhere.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -223,7 +223,7 @@ A commit should observe the following:
 
 A commit message should observe the following (based on ["A Note About Git Commit Messages"](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)):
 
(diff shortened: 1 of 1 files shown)


vexx32 (approved):

CONTRIBUTING.md
```diff
@@ -223,7 +223,7 @@ A commit should observe the following:
 
 A commit message should observe the following (based on ["A Note About Git Commit Messages"](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)):
 
-  * The first line of the commit message should be a short description around 50 characters in length and be prefixed with the GitHub issue it refers to with parentheses surrounding that. If the GitHub issue is #25, you should have `(GH-25)` prefixed to the message.
+  * The first line of the commit message should be a short description around 50 characters in length and be prefixed with the GitHub issue it refers to with parentheses surrounding that. If the GitHub issue is #25, you should have `(#25)` prefixed to the message.
```
ferventcoder, line 226: This is great - we should mention here that we previously requested GH, but that is deprecated. Approved from me otherwise.

PR #2208, opened 2021-03-01
(#2051) Make parameter description consistent

gep13:
In an earlier commit, the default value of the WorkingDirectory
parameter was changed to be $null, but the parameter description wasn't
updated.

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Start-ChocolateyProcessAsAdmin.ps1
+++ b/src/chocolatey.resources/helpers/functions/Start-ChocolateyProcessAsAdmin.ps1
@@ -65,8 +65,8 @@ Array of exit codes indicating success. Defaults to `@(0)`.
 
 .PARAMETER WorkingDirectory
 The working directory for the running process. Defaults to
-`Get-Location`. If current location is a UNC path, uses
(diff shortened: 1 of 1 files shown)


src/chocolatey.resources/helpers/functions/Start-ChocolateyProcessAsAdmin.ps1
```diff
@@ -65,8 +65,8 @@ Array of exit codes indicating success. Defaults to `@(0)`.
 
 .PARAMETER WorkingDirectory
 The working directory for the running process. Defaults to
-`Get-Location`. If current location is a UNC path, uses
-`$env:TEMP` for default as of 0.10.14.
+`$null`. If current location is a UNC path, uses `$env:TEMP` for default
+ as of 0.10.14.
```
ferventcoder, line 69: If null is passed in, it will set it to `Get-Location` - that note needs to be addressed here me thinks?

ferventcoder (changes requested):
questions on clarity of what we changed

gep13:
Having discussed this with @pauby today, we don't believe that this PR is required, since the new applied logic in the previous PR actually matches what the help text said previously.

PR #2218, opened 2021-03-25
(#1866) Avoid recursive search for pending package removal

### LINKED ISSUE #1866: remove pending packages task - search only top level level package folders to clean up

jetersen:
fixes #1866
fixes #1941
fixes https://github.com/actions/virtual-environments/issues/3010

Created a repro over at https://github.com/jetersen/test-choco-pack-github-action

Opened this PR as #1941 was not signed, which I assume was blocking.
This should help alleviate the initial choco startup on CI systems.

### DIFF
--- a/src/chocolatey/infrastructure.app/tasks/RemovePendingPackagesTask.cs
+++ b/src/chocolatey/infrastructure.app/tasks/RemovePendingPackagesTask.cs
@@ -60,13 +60,13 @@ namespace chocolatey.infrastructure.app.tasks
         {
             this.Log().Debug(ChocolateyLoggers.Verbose, "[Pending] Removing all pending packages that should not be considered installed...");
 
-            var pendingFiles = _fileSystem.get_files(ApplicationParameters.PackagesLocation, ApplicationParameters.PackagePendingFileName, SearchOption.AllDirectories).ToList();
+            var pendingFiles = _fileSystem.get_files(ApplicationParameters.PackagesLocation, ApplicationParameters.PackagePendingFileName).ToList();
             foreach (var pendingFile in pendingFiles.or_empty_list_if_null())
             {
                 var packageFolder = _fileSystem.get_directory_name(pendingFile);
                 string packageFolderName = _fileSystem.get_directory_info_for(packageFolder).Name;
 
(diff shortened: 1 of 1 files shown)


gep13:
@jetersen as you have pointed out, the CLA hasn't been signed on the other PR, so yes, I feel we can move forward with this PR, but I will update the commit message here to provide some additional context.

jetersen:
@gep13 much appreciated

gep13:
@jetersen thanks again for taking the time to raise this PR!

jetersen:
@gep13 the last chocolatey release was quite a while ago. What are the chances for seeing a release of this in the near future?

TheCakeIsNaOH:
@jetersen here are the items that are intended to be completed for 0.10.16, which is the next release:
https://github.com/chocolatey/choco/milestone/43

jetersen:
So, that's a hard no? Would more frequent releases be better?

TheCakeIsNaOH:
> So, that's a hard no?

I'm just a contributor, so I am not the one making the decisions. 
If there is say a security vulnerability discovered or something, then a release might be made sooner.
But fixing those issues are the current plan.

jetersen:
@gep13 I can see v0.10.16 is TBD https://github.com/chocolatey/choco/issues/2266 
This PR fixes an annoying issue do we have a ETA for the release? 😅

pauby:
We don't have an ETA on the 0.10.16 release at the moment.

PR #2234, opened 2021-04-15
[IGNORE] try travis without manual mono-devel install

TheCakeIsNaOH:


gep13:
@TheCakeIsNaOH thanks for doing this test, I will go ahead and close out this PR.

Issue #2236, opened 2021-04-16
Mono build test failure on systems with a merged usr


TheCakeIsNaOH:
On some linux systems, system binaries are not under `/bin`, but under `/usr/bin`, along with `sbin` and `lib` moving to under `/usr`. The binaries may be symlinked to their original location under `/bin` as well for backwards compatibility, but when looking up the path to the binary via the `$PATH`, they are under `/usr/bin`.
https://wiki.debian.org/UsrMerge

This is an issue as the filesystem test hardcodes `ls` to be `/bin/ls`, and fails if it actually located in `/usr/bin/ls`
https://github.com/chocolatey/choco/blob/master/src/chocolatey.tests/infrastructure/filesystem/DotNetFileSystemSpecs.cs#L219

This should be switched to allow both `/bin/ls` and `/usr/bin/ls`

PR #2237, opened 2021-04-16
(#2236) Mono test work with merged usr systems

TheCakeIsNaOH:
The should_find_existing_executable() test has a hardcoded path to the
ls executable, which is incorrect on systems with a merged usr
directory, as they have binaries under /usr/bin instead of under /bin
Switches the test to work with both path.

Fixes: #2236 

I'm pretty sure `ShouldBeInRange` is the correct thing to use, but feel free to correct if there is a better method to use.

### DIFF
--- a/src/chocolatey.tests/infrastructure/filesystem/DotNetFileSystemSpecs.cs
+++ b/src/chocolatey.tests/infrastructure/filesystem/DotNetFileSystemSpecs.cs
@@ -214,10 +214,17 @@ namespace chocolatey.tests.infrastructure.filesystem
             [Fact]
             public void GetExecutablePath_should_find_existing_executable()
             {
-                FileSystem.get_executable_path("ls").ShouldEqual(
-                    Platform.get_platform() != PlatformType.Windows
-                        ? "/bin/ls"
-                        : "ls");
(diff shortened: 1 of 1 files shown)


gep13:
@TheCakeIsNaOH Would it not be better to use:

https://docs.shouldly.io/documentation/enumerable/oneof

The `ShouldBeInRange` seems to apply more to a range of numbers, and confirming that a given number is contained within that upper and lower range limits.

TheCakeIsNaOH:
@gep13 Choco has `should` not `shouldly`

I can't find `ShouldBeOneOf` in `should`.

TheCakeIsNaOH:
I do agree that `ShouldBeInRange` is not quite correct, looking at it again.

I've switched this to use `ShouldContain` and inverted the comparison, so the strings to compare to are on the left.

gep13 (approved):
LGTM!

gep13:
>@TheCakeIsNaOH said...
Choco has should not shouldly

Doh! Good catch! I made an assumption here, and didn't verify.

gep13:
@TheCakeIsNaOH your changes have been merged, thanks for your contribution 👍

Issue #2238, opened 2021-04-16
Mono build on non-windows cannot find xbuild if it is not in /usr/bin


TheCakeIsNaOH:
Uppercut appears to hardcode the xbuild path to `/usr/bin/xbuild`:
https://github.com/chocolatey/choco/blob/master/.build/compile.step#L118

This is fine most of the time, however, if xbuild is not available there, the mono build will fail.
For example, the Github actions MacOS environment has `xbuild` in `/Library/Frameworks/Mono.framework/Versions/Current/Commands`

Removing the `basedir` appears to work, as then it just uses whatever `xbuild` is on the path, if I understand what is going on correctly. 

Log of failing build:
```
     [echo] Removing and adding /Users/runner/work/choco/choco/build_output.
     [echo] Build for chocolatey started at 04/16/2021 20:13:58 with UppercuT.
     [echo] Using UppercuT version: 1.4.2.0.
     [echo] ====================
     [echo] UppercuTBuild
     [echo] ====================
                 [echo] ====================
                 [echo] GitVersioner
                 [echo] ====================
                 [echo] Getting revision number from working directory '/Users/runner/work/choco/choco' or remotely at https://github.com/chocolatey/choco.
                 [exec] fatal: No names found, cannot describe anything.
                 [exec] /Users/runner/work/choco/choco/.build/versioners/git.step(26,6):
                 [exec] External Program Failed: git (return code was 128)
     [echo] Using revision number: fatal: No names found, cannot describe anything. and hash: fatal: No names found, cannot describe anything..
     [echo] Using build number 0.
     [echo] Version full updated to 0.10.16.0.
     [echo] Compiled code drop directory will be /Users/runner/work/choco/choco/code_drop.
                 [echo] ====================
                 [echo] VersionBuilder
                 [echo] ====================
                 [echo] Generating a strong name key (/Users/runner/work/choco/choco/chocolatey.snk) using /Applications/Microsoft SDKs/Windows/v6.0A/Bin/sn.exe if the key file doesn't already exist.
                 [echo] Generating SolutionVersion.cs with assembly version 0.10.16.0.
                 [echo] ====================
                 [echo] BuildInformation
                 [echo] ====================
                 [echo] Generating XML file with some information to possibly get merged in at /Users/runner/work/choco/choco/build_output/_BuildInfo.xml.
                 [echo] ====================
                 [echo] Compiler
                 [echo] ====================
                        [echo] 
                        [echo] ====================
[...]

PR #2239, opened 2021-04-16
(#2238) Mono build don't hardcode xbuild path

TheCakeIsNaOH:
Removes basedir from Mono build on non-windows platforms.
This lets xbuild be located at other locations, as long at it's on the path

Fixes: #2238 

I'm not %100 sure that this is the best solution, but it seems to work everwhere I have tested it.
Ubuntu, Debian, Docker, and MacOS (in Github actions) all build.

I'm not super familiar with uppercut, so if there is a better way to do this, I'm all ears.

### DIFF
--- a/.build/compile.step
+++ b/.build/compile.step
@@ -108,14 +108,12 @@
       <!-- Do clean and build in two steps since xbuild has a bug with using OutputPath and Rebuild together Bug #628525 at Novell -->
       <echo level="Warning" message="Cleaning build..." if="${microsoft.framework.specific != 'mono-4.5'}" />
       <exec program="${app.xbuild}"
-          basedir="/usr/bin"
           workingdir="${dirs.build}"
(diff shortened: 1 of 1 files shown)


gep13:
@ferventcoder would you be in a position to provide some guidance here?

The idea behind this is that we would like to build Chocolatey on GitHub Actions, running on Windows, Mac and Linux, and there is an issue when trying this due to the location of the xbuild application on some runners.

I know that directly editing the UppercuT files is not recommended, so I am curious as to what the best course of action is here.

gep13 (approved):
LGTM!

gep13:
@TheCakeIsNaOH thanks for investigating this, and for finding a solution.  Will get this merged once the CI builds complete.

gep13:
There is an open GitHub Actions incident which I believe is why the Windows build isn't working correctly:

https://www.githubstatus.com/incidents/cj7gzzj30411

Since this change only impacts the Posix build, which has passed, I am going to go ahead and merge this in.

gep13:
@TheCakeIsNaOH thanks again for getting this fixed up!

Issue #1443, opened 2017-11-06
Upgrade all reuses overridden package parameters when useRememberedArgumentsForUpgrades feature is turned on


dinfinity:
### What You Are Seeing?
I have a multitude of packages installed via chocolatey, a number of which I have set custom install directories for. Chocolatey regularly shows error messages that indicate that the overridden parameters for another package are used. For instance:

```
ERROR: Running ["C:\Users\Dual Infinity\AppData\Local\Temp\chocolatey\TotalCommander\9.10\INSTALL.exe" INSTALLDIR=C:\Frameworks\NodeJS] was not successful. Exit code was '1'. See log for possible error messages.
The upgrade of totalcommander was NOT successful.
Error while running 'C:\ProgramData\chocolatey\lib\TotalCommander\tools\chocolateyInstall.ps1'.
 See log for details.
```

It seems to use the last overridden parameters encountered.

Addendum 2017-11-07:
The issue is usually resolved by running the upgrade all command multiple times. It does not seem random, but it does seem that if a package with custom parameters was installed successfully its parameters are not read and do not conflict in the subsequent run of upgrade all.

### What is Expected?
Chocolatey should (obviously) not reuse package parameters for other packages.

### How Did You Get This To Happen? (Steps to Reproduce)
In essence, this would entail setting custom install dirs for multiple packages and running chocolatey upgrade all.

### Output Log
See zip:
[dinfinity.chocolatey.log.2017-11-06.zip](https://github.com/chocolatey/choco/files/1446353/dinfinity.chocolatey.log.2017-11-06.zip)

### Related 
* https://github.com/chocolatey/choco/issues/797

ferventcoder:
This sounds like a bug, and a major one.

ferventcoder:
This isn't going to be quite as easy to find the cause of as hoped. Thanks for the log file. The code in question doesn't reuse parameters (on first glance, there could be a small bug in there). 

I think to really find this, we're going to need to understand how you installed everything and possibly take a much closer look (if you are up for it).  

How did you install things? One at a time or did you run multiple choco.exe processes at the same time?

dinfinity:
Taking a closer look is no problem. I like chocolatey a lot.

I have been ignoring this issue for a couple of months now, so I can't say for sure I did not run multiple simultaneous choco commands, although I can't remember doing so and deem it quite unlikely.

I could provide a zip of (parts of) my ProgramData\chocolatey directory if that helps.

I'm adding the following to the issue description:
[...]

PR #2240, opened 2021-04-17
(#2111) Deep copy config

TheCakeIsNaOH:
Always deep copy the config when making a backup originalConfig. This
ensures that when the config is reset it gets fully reset. Fixes issues
with reuse of arguments when useRememberedArgumentsForUpgrades is
enabled.

Hopefully fixes: #1443
Fixes: #2111

I have tested this with the procedure in [this comment](https://github.com/chocolatey/choco/issues/2111#issuecomment-759124896), and with this patch, it no longer upgrades python3 to a prerelease build, while without this patch it does.

### DIFF
--- a/src/chocolatey/infrastructure.app/services/NugetService.cs
+++ b/src/chocolatey/infrastructure.app/services/NugetService.cs
@@ -435,7 +435,7 @@ folder.");
                 uninstallSuccessAction: null,
                 addUninstallHandler: true);
 
-            var originalConfig = config;
+            var originalConfig = config.deep_copy();
 
             foreach (string packageName in packageNames.or_empty_list_if_null())
             {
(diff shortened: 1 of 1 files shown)


gep13:
@TheCakeIsNaOH in principal, I don't have an issue with this change, and there is a precedence for this change here:

https://github.com/chocolatey/choco/blob/master/src/chocolatey/infrastructure.app/services/NugetService.cs#L990

I am curious as to why you didn't make the same change on this line though:

https://github.com/chocolatey/choco/blob/master/src/chocolatey/infrastructure.app/services/NugetService.cs#L1360

TheCakeIsNaOH:
I just missed that one, I think I screwed up my `git add`, I've added it.

gep13 (approved):
LGTM!

gep13:
@TheCakeIsNaOH thanks a lot of digging into this, and coming up with the steps to reproduce and diagnose this.

PR #2286, opened 2021-06-07
(#1541) Add priority ordering for repository optimizations

### LINKED ISSUE #1541: choco source priority is ignored
### What You Are Seeing?

If you have multiple chocolatey sources, the priority which you give them is not applied.

Quote from the Choco Wiki: 

> Priority - The priority order of this source as compared to other 
>        sources, lower is better. Defaults to 0 (no priority). All priorities 
>        above 0 will be evaluated first, then zero-based values will be 
>        evaluated in config file order

This issue happens if you have the default source set to 0 or to a higher number than your prefered source.

### What is Expected?

The expectation is that every source with priority > 0 is prefered and that the lower the the priority number, the higher the priority. Nevertheless this does not seem to be the case.

Another Quote from Gary Ewan Parker regarding the default behaviour:

> Each source can have a priority, which defines the order that Chocolatey searches the sources to find a package. If a package is found on a higher priority source, that package version is used **even** if there is a newer version on another lower priority source.

In this Example we have ccleaner **5.35.6210 on our own source**, the **public source provides the version 5.41.6446**. Here you can see, that the "chocolatey" priority is set to 2 (same behaviour with 0) and our "tkcloud" source is set to 1 (lower so it should be prefered):

2018-04-18_10h56_00
Nevertheless, the public version is selected.

If we disable the default repository, the correct version from our source is selected:
2018-04-18_10h58_13

We could verify this faulty behaviour on multiple systems. Even a remove and readd of the sources does not resolve this

### Output Log
On another note, the choco list commands provides the packages in the correct order:

2018-04-18 12_21_58-auswahlen windows powershell

The only problem seems to be the install commands.

### Additional Information
[Zendesk ticket 26658](https://chocolatey.zendesk.com/agent/tickets/26658)

┆Issue is synchronized with this [Clickup task](https://app.clickup.com/t/868h60c7q) by [Unito](https://www.unito.io)

AdmiringWorm:
This pull requests re-adds the ability to order sources by priority when installing and upgrading packages.

fixes #1541

### DIFF
--- a/src/chocolatey.tests.integration/scenarios/InstallScenarios.cs
+++ b/src/chocolatey.tests.integration/scenarios/InstallScenarios.cs
@@ -2993,6 +2993,154 @@ namespace chocolatey.tests.integration.scenarios
         }
 
         [Concern(typeof(ChocolateyInstallCommand))]
+        public class when_installing_new_package_from_priority_source_with_repository_optimization : ScenariosBase
+        {
+            public override void Context()
+            {
+                base.Context();
+                Configuration.PackageNames = Configuration.Input = "upgradepackage";
+                Configuration.Features.UsePackageRepositoryOptimizations = true;
+                Configuration.Sources = Configuration.Sources + ";chocolatey"; // Needed, otherwise we can't do repository optimizations
+                Configuration.MachineSources.Add(new MachineSourceConfiguration
+                {
+                    Name = "chocolatey",
+                    Key = "https://community.chocolatey.org/api/v2"
+                });
+                Scenario.add_packages_to_source_location(Configuration, "upgradepackage.1.1.0*" + Constants.PackageExtension);
+                Scenario.add_packages_to_priority_source_location(Configuration, "upgradepackage.1.0.0*" + Constants.PackageExtension);
+            }
+
+            public override void Because()
+            {
+                Results = Service.install_run(Configuration);
+            }
+
+            [Fact]
+            public void should_install_where_install_location_reports()
+            {
+                foreach (var packageResult in Results)
+                {
+                    Directory.Exists(packageResult.Value.InstallLocation).ShouldBeTrue();
+                }
+            }
+
+            [Fact]
+            public void should_install_a_package_in_the_lib_directory()
+            {
+                var packageDir = Path.Combine(Scenario.get_top_level(), "lib", Configuration.PackageNames);
+
+                Directory.Exists(packageDir).ShouldBeTrue();
+            }
+
+            [Fact]
+            public void should_install_lower_version_of_package()
+            {
+                foreach (var packageResult in Results)
+                {
+                    packageResult.Value.Version.ShouldEqual("1.0.0");
+                }
+            }
+
+            [Fact]
+            public void should_have_a_successful_package_result()
+            {
+                foreach (var packageResult in Results)
+                {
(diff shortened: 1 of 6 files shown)


src/chocolatey/infrastructure.app/nuget/NugetList.cs
```diff
@@ -180,6 +187,27 @@ public static IPackage find_package(string packageName, SemanticVersion version,
             // find the package based on version using older method
             if (version != null) return repository.FindPackage(packageName, version, config.Prerelease, allowUnlisted: false);
 
+            var priorityRepository = repository as PriorityAggregateRepository;
+            if (priorityRepository != null)
+            {
+                // Returns an IEnumerable of Tuple
+                // Item1 = the priory
```
gep13, line 194:
```suggestion
                // Item1 = the priority

```

src/chocolatey/infrastructure.app/nuget/NugetList.cs
```diff
@@ -180,6 +187,27 @@ public static IPackage find_package(string packageName, SemanticVersion version,
             // find the package based on version using older method
             if (version != null) return repository.FindPackage(packageName, version, config.Prerelease, allowUnlisted: false);
 
+            var priorityRepository = repository as PriorityAggregateRepository;
+            if (priorityRepository != null)
+            {
+                // Returns an IEnumerable of Tuple
+                // Item1 = the priory
+                // Item2 = the repository list for that priory
```
gep13, line 195:
```suggestion
                // Item2 = the repository list for that priority

```

src/chocolatey/infrastructure.app/nuget/PriorityAggregateRepository.cs
```diff
@@ -0,0 +1,160 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using chocolatey.infrastructure.app.configuration;
+using NuGet;
+
+namespace chocolatey.infrastructure.app.nuget
+{
+    /// <summary>
+    /// Helper file for grouping repositories with a distinct priory set.
```
gep13, line 12:
```suggestion
    /// Helper file for grouping repositories with a distinct priority set.
```

src/chocolatey.tests.integration/scenarios/InstallScenarios.cs
```diff
@@ -2992,6 +2992,154 @@ public void should_have_an_error_package_result()
...
+            [Fact]
+            public void should_not_have_warning_package_result()
+            {
+                foreach (var packageResult in Results)
+                {
+                    packageResult.Value.Warning.ShouldBeFalse();
+                }
+            }
+        }
+
```
gep13, line 3071: There is an extra unnecessary line here.

gep13 (changes requested):
A couple minor changes based on a quick review.

gep13:
@AdmiringWorm would it be possible for you to update this PR with the new template that we are using in other repositories, i.e. include information about the testing that was performed (in addition to the new tests which have been added - thanks for getting those added!)

AdmiringWorm:
@gep13 I have fixed the requested changes, updated the template for the PR and changed the PR to target the develop branch (instead of the old master branch).

The testing steps added was added from memory, and not the same exact steps I had done back when I opened the PR.

Issue #2287, opened 2021-06-08
Improvements to Tab Expansion (Multiple Packages, allow dash)


cpmcgrath:
When using tab completion there's two issues I often run into:

* I want to update multiple packages, but tab expansion only works for the first package listed
* I type something like **choco upgrade dotnet-** the expansion stops working.

Luckily, these are pretty easy fixes, that can be achieved by updating the regexes:

```
^upgrade\s+(?<package>[^\.][^-\s]*)$
^upgrade.*\s+(?<package>[^\.-\s][^\s]*)$
```
Basically adding .* means that the package name becomes the last word in the string instead of the first.
By moving the dash to the first part of the package name, it changes it from saying a package name cannot have a dash, to a package name cannot start with a dash. Also worth putting \s in there to be clear that a package cannot start with a space. This also makes a noticeable speed improvement on **choco upgrade -** as the original regex allows - as a valid package name, searches, fails and then will look for the switches.

then rinse and repeat for install, uninstall

PR #2288, opened 2021-06-08
(#2287) Improvements to Tab Expansion

cpmcgrath:
Previously, when installing, updating, or uninstalling packages, the
tab expansion would only occur for the first item. Further, the tab
expansion would not work if the search term contained a dash.

Now tab expansion works for all items, and items can contain dashes
as long as it does not start with dash.


gep13:
@cpmcgrath thank you very much for taking the time to make this contribution, and congratulations on this being your first contribution to Chocolatey.

Would you be able to update the description of this PR to include information about how this has been tested?

Although not pulled in yet, we are going to be moving towards a template like the following:

https://github.com/chocolatey/.github/pull/1/files#diff-479755504b0aad0fed26047714591511b4d07c76bc8521171b87a5f16100f850

If you could update this PR to use this template, it would greatly help move it towards getting merged in.

Thanks

PR #2295, opened 2021-06-21
(#2124) Add ability to initialize without logging

### LINKED ISSUE #2124: API - Allow Lets.GetChocolatey to not initialize logging
In some cases you want to call Lets.GetChocolatey without initializing logging (or making any changes to how it might currently be set).

AdmiringWorm:
This PR makes changes to the API initialization method to allow
Chocolatey to allow the initialization to happen without enabling
logging.
Additionally changes have been made to allow adding extra loggers to be
added without overriding the logging
to the applications main chocolatey.log file.

These changes were necessary to allow programs (like ChocolateyGUI)
to add its own logging infrastructure without disabling logging
to the main log file.

fixes #2124

### DIFF
--- a/src/chocolatey/GetChocolatey.cs
+++ b/src/chocolatey/GetChocolatey.cs
@@ -18,6 +18,7 @@ namespace chocolatey
 {
     using System;
     using System.Collections.Generic;
+    using System.IO;
     using System.Reflection;
     using infrastructure.licensing;
     using SimpleInjector;
@@ -47,16 +48,21 @@ namespace chocolatey
     {
         private static readonly log4net.ILog _logger = LogManager.GetLogger(typeof(Lets));
 
-        private static GetChocolatey set_up()
+        private static GetChocolatey set_up(bool initializeLogging)
         {
             add_assembly_resolver();
 
-            return new GetChocolatey();
+            return new GetChocolatey(initializeLogging);
         }
 
         public static GetChocolatey GetChocolatey()
         {
-            return GlobalMutex.enter(() => set_up(), 10);
+            return GetChocolatey(initializeLogging: true);
+        }
+
+        public static GetChocolatey GetChocolatey(bool initializeLogging)
+        {
+            return GlobalMutex.enter(() => set_up(initializeLogging), 10);
         }
 
         private static ResolveEventHandler _handler = null;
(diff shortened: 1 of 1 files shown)


src/chocolatey/GetChocolatey.cs
```diff
@@ -37,6 +37,7 @@ namespace chocolatey
     using Assembly = infrastructure.adapters.Assembly;
     using IFileSystem = infrastructure.filesystem.IFileSystem;
     using ILog = infrastructure.logging.ILog;
+    using System.IO;
```
gep13, line 40: Can this be moved up with the other `System.*` namespaces?

src/chocolatey/GetChocolatey.cs
```diff
@@ -47,16 +48,21 @@ public static class Lets
...
             add_assembly_resolver();
 
-            return new GetChocolatey();
+            return new GetChocolatey(initializeLogging);
         }
 
         public static GetChocolatey GetChocolatey()
         {
-            return GlobalMutex.enter(() => set_up(), 10);
+            return GetChocolatey(true);
```
gep13, line 60:
```suggestion
            return GetChocolatey(initializeLogging: true);
```

src/chocolatey/GetChocolatey.cs
```diff
@@ -122,10 +128,25 @@ public class GetChocolatey
...
+        public GetChocolatey(bool initializeLogging)
         {
-            Log4NetAppenderConfiguration.configure(null, excludeLoggerNames: ChocolateyLoggers.Trace.to_string());
-            Bootstrap.initialize();
-            Log.InitializeWith(new AggregateLog(new List<ILog>() { new Log4NetLog(), _logSinkLogger }));
+            if (initializeLogging)
+            {
+                string loggingLocation = ApplicationParameters.LoggingLocation;
+                
+                if (!Directory.Exists(loggingLocation))
```
gep13, line 141: I am just wondering about the direct usage of the Directory class here.  Normally, where possible we use the FileSystem abstraction when IO operations are performed.  While we don't have a direct IFileSystem instance here, we do have access to the `_container` where we could get one from.  Thoughts?

gep13 (changes requested):

src/chocolatey/GetChocolatey.cs
AdmiringWorm: I'm not sure, the main reason for using the `Directory` class here was due to Rob also did this himself in the Work he had done.
I only assumed it was because at this point chocolatey hadn't yet been set up, but I'll run some tests to check if it would be possible for us to use `IFileSystem` already here.

AdmiringWorm: I did some testing around using the `IFileSystem` resolved from the container, and that works as it should.

The PR have been updated to make use of it instead of raw calls to the `Directory` class.

gep13 (approved):
LGTM!

gep13:
@AdmiringWorm any thoughts on why this is failing the ubuntu build?

vexx32:
Seems the nuget restore failed there with SSL issues, similar results for all dependent packages:
```
                 [exec] Restoring NuGet package NUnit.Runners.2.6.4.
                 [exec] WARNING: Unable to find version '2.6.4' of package 'NUnit.Runners'.
                 [exec]   https://api.nuget.org/v3/index.json: Unable to load the service index for source https://api.nuget.org/v3/index.json.
                 [exec]   The SSL connection could not be established, see inner exception.
                 [exec]   Authentication failed, see inner exception.
                 [exec]   Ssl error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED
                 [exec]     at /build/mono-6.12.0.122/external/boringssl/ssl/handshake_client.c:1132
                 [exec]   https://www.nuget.org/api/v2/: Unable to load the service index for source https://www.nuget.org/api/v2/.
                 [exec]   The SSL connection could not be established, see inner exception.
                 [exec]   Authentication failed, see inner exception.
                 [exec]   Ssl error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED
                 [exec]     at /build/mono-6.12.0.122/external/boringssl/ssl/handshake_client.c:1132
```

gep13:
@vexx32 yip, saw the same thing, and was just pinging @AdmiringWorm about it on Slack 😄

gep13:
@vexx32 @AdmiringWorm I have opened a PR attempting to address this here: https://github.com/chocolatey/choco/pull/2314 Can you both have a review, and let me know what you think?  Thanks!

gep13 (approved):
LGTM!

PR #2307, opened 2021-08-02
(#877) Set SourceType for packages parsed from config

### LINKED ISSUE #877: Installing from packages.config should support windowsfeatures (and other metapackages)
_(I did some digging for an issue or mention of this behaviour and could not find anything; my apologies if something does exist. Many thanks for your help)_
### What You Are Seeing?

I have a packages.config with the following entry:
`<package id="IIS-WebServerRole" source="windowsfeatures" />`
and it fails to install with the error: "**The package was not found with the source(s) listed**"
### What is Expected?

Anything that can be installed via "choco install" (which the above can) should be supported in the packages.config. E.g. `choco install IIS-WebServerRole -s windowsfeatures` succeeds on the same machine.
- Create a packages.config with the above entry (I also had a "notepadplusplus" entry but I don't think that changes the result).
- Run choco install packages.config
### Output Log

https://gist.github.com/chrispellett/af88b0b4eb1969e783e9a87c300d6b9e

AdmiringWorm:
This commit updates the handling of packages when they are read from
a `.config` file to make sure that the SourceType property will be
correctly set.

This was needed for subsequent code to resolve the Source Runner correctly
when the specified source is not a normal Chocolatey package source.

fixes #877

## How has this been tested?

The following packages.config file was created:

```
<?xml version="1.0" encoding="utf-8"?>
<packages>
  <package id="IIS-WebServerRole" source="windowsfeatures" />
  <package id="windirstat" version="1.1.2.20161210" />
</packages>
```

And then the command `choco install test.config` file was run.

When running with current released version of Chocolatey, this causes an error, since the IIS-WebServerRole package can't be found.

When running with the code from this branch, the feature was correctly installed, as well as the windirstat package.

### DIFF
--- a/src/chocolatey.tests/chocolatey.tests.csproj
+++ b/src/chocolatey.tests/chocolatey.tests.csproj
@@ -108,6 +108,7 @@
     <Compile Include="infrastructure.app\configuration\ConfigurationOptionsSpec.cs" />
     <Compile Include="infrastructure.app\nuget\NugetCommonSpecs.cs" />
     <Compile Include="infrastructure.app\services\AutomaticUninstallerServiceSpecs.cs" />
+    <Compile Include="infrastructure.app\services\ChocolateyPackageServiceSpecs.cs" />
     <Compile Include="infrastructure.app\services\FilesServiceSpecs.cs" />
     <Compile Include="infrastructure.app\services\NugetServiceSpecs.cs" />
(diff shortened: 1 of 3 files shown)


gep13 (approved):
LGTM!

gep13:
@AdmiringWorm thanks for getting this fixed!

PR #2337, opened 2021-08-31
(#2336) Remove quotes from InstallLocation

### LINKED ISSUE #2336: AutomaticUninstallerService doesn't handle quoted InstallLocation
### What You Are Seeing?

Stemming from a [Chocolatey.Extension issue](https://github.com/chocolatey/chocolatey-licensed-issues/issues/48), when the `InstallLocation` is quoted, the FileSystem object doesn't detect a directory as existing if the path is quoted. This means the test for directory existing is failing even when the directory does in fact exist.

### What is Expected?

When checking for the existence of the `InstallLocation`, if there are quotes in the path they should be handled appropriately.

1. Install the chocolatey package lastpass
1. Try uninstalling it with choco uninstall lastpass --fromprograms

### Output Log

https://gist.github.com/AdmiringWorm/8998e1703a6ba07b34e5dd0b6836e033

corbob:
## Description

The InstallLocation can sometimes have quotes around the file path. When this happens we don't detect the directory as existing even when it does. As such, the uninstall doesn't actually happen. This change removes the quotes before we ask the File System Provider if the directory exists.

## Testing

Tested this by Installing the Lastpass Chocolatey package then running `choco uninstall --fromprograms` (This is a licensed feature) and confirmed that it successfully called the uninstaller.

I have created a test package `uninstall-from-programs` that can be used for testing this as well (it creates an install entry with quotes and does not successfully uninstall without this change).

Fixes #2336 

The manual testing documentation for chocolatey.extension has been updated to include testing the `uninstall-from-programs` package.

### DIFF
--- a/src/chocolatey/infrastructure.app/services/AutomaticUninstallerService.cs
+++ b/src/chocolatey/infrastructure.app/services/AutomaticUninstallerService.cs
@@ -131,7 +131,7 @@ namespace chocolatey.infrastructure.app.services
 
             this.Log().Debug(() => " Preparing uninstall key '{0}' for '{1}'".format_with(key.UninstallString.to_string().escape_curly_braces(), key.DisplayName.to_string().escape_curly_braces()));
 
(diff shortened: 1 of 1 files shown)


gep13:

PR #2352, opened 2021-09-08
Fix typo in setting Authorization header

glatuske:
The explicitly added Authorization header does not get added to the request object (re instead of req).

That means downloading a file by choco, which requires an Authorization header to bet set, is currently not possible.
An error message like this will be shown: "The property 'Authorization' cannot be found on this object. Verify that the property exists and can be set."

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Get-WebFile.ps1
+++ b/src/chocolatey.resources/helpers/functions/Get-WebFile.ps1
@@ -186,7 +186,7 @@ param(
         'Cookie' {$req.CookieContainer.SetCookies($uri, $options.headers.$key)}
         'Referer' {$req.Referer = $options.headers.$key}
         'User-Agent' {$req.UserAgent = $options.headers.$key}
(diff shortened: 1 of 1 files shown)


gep13 (approved):
LGTM!

gep13:
@glatuske thank you very much for taking the time to fix this, we really appreciate it!

Congratulations on your first time contribution to this repository!

This change will be going out in the 0.11.2 release of Chocolatey.

gep13:
@glatuske actually, after doing some more testing, there is more to this problem than it would first appear.

I have written up some notes on it here:

https://github.com/chocolatey/choco/issues/2374#issue-1002051805

Long story short is, the fix that you provided doesn't actually solve the problem.  Looks like there wasn't actually a problem to begin with.

I am going to create a follow up commit to remove the code that was introduced in v0.11.0, which should put us back to a place where Authorization headers can once again be sent along to allow download of assets that are secured.

PR #2353, opened 2021-09-09
(test) Add unit tests for ArgumentsUtility class

gep13:
This is starting to be used in other places, so we should have some
tests around the execution of the logic contained within this class to
ensure that there are no regressions if/when editing this class.

For now, all the current expectations are covered within the different
TestFixtures, but this should be extended if any changes are made to
the class.

### DIFF
--- a/src/chocolatey.tests/infrastructure.app/utility/ArgumentsUtilitySpecs.cs
+++ b/src/chocolatey.tests/infrastructure.app/utility/ArgumentsUtilitySpecs.cs
@@ -1,0 +1,62 @@
+﻿namespace chocolatey.tests.infrastructure.app.utility
+{
+    using chocolatey.infrastructure.app.utility;
+    using NUnit.Framework;
+    using Should;
+
+    public class ArgumentsUtilitySpecs
+    {
+        public abstract class ArgumentsUtilitySpecsBase : TinySpec
+        {
+            public override void Context()
+            {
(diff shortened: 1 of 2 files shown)


src/chocolatey.tests/infrastructure.app/utility/ArgumentsUtilitySpecs.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            public override void Context()
+            {
+            }
+        }
+
+        [TestFixture("choco install bob --package-parameters-sensitive=\"/test=bill\"", true)]
+        [TestFixture("choco install bob -package-parameters-sensitive=\"/test=bill\"", true)]
+        [TestFixture("choco install bob --install-arguments-sensitive=\"/test=bill\"", true)]
+        [TestFixture("choco install bob -install-arguments-sensitive=\"/test=bill\"", true)]
```
AdmiringWorm, line 19: Should maybe test for the same with `upgrade` and `uninstall`? (I know parameters can be used in `uninstall`, but unsure if install arguments can though).

AdmiringWorm (approved):
As the utility do not take into account the command names, ignore my previous comment.

LGTM 👍

gep13:
@AdmiringWorm thanks for checking this, going to get this merged in.

Issue #2358, opened 2021-09-17
Add scripts for debug and official build targets


TheCakeIsNaOH:
To allow an official docker container, a mono build script that targets `ReleaseOfficial` is required.
It would make sense to also include a script that creates a `debug` build as well.

Required for https://github.com/chocolatey/choco/issues/1718

PR #2359, opened 2021-09-17
(#2358) Add official and debug Mono build scripts

TheCakeIsNaOH:
Fixes: #2358

I think we may have to dig into uppercut as well to get the code signing working on Mono.

### DIFF
--- a/build.debug.sh
+++ b/build.debug.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+# stty -echo
+
+# ::Project UppercuT - http://uppercut.googlecode.com
+# ::No edits to this file are required - http://uppercut.pbwiki.com
+
+function usage
+{
+    echo ""
+    echo "Usage: build.debug.sh"
+    exit
+}
+
+function displayUsage
+{
+    case $1 in
+        "/?"|"-?"|"?"|"/help") usage ;;
+    esac
+}
+
+displayUsage $1
+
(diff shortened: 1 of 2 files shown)


gep13:
@TheCakeIsNaOH I have just taken these for a spin on my Mac and they work, although I had to `chmod` them before they would execute.  Is that expected?

TheCakeIsNaOH:
@gep13 No, that is not expected, I think? I'll look into it.

------
> I think we may have to dig into uppercut as well to get the code signing working on Mono.

Ok, yup, this is still building with the unofficial strongname key, and is not trying to codesign. I think the script is fine? But uppercut will have to be fixed to set the official configuration using the official strongnaming key. And I don't quite know enough about authenticode to even know where to start on getting that signing working on linux.

TheCakeIsNaOH:
I think I've fixed it.

gep13:
> But uppercut will have to be fixed to set the official configuration using the official strongnaming key.

Some of this is driven by the presence of certain files on the file system, and environment variables.  Leave this part with me, and I will let you know how I get on.

gep13:
> I think I've fixed it.

Just tested again, and this is now working.

gep13 (approved):
LGTM!

gep13:
@TheCakeIsNaOH thanks for getting these added!

Issue #2364, opened 2021-09-17
Linux docker includes source code


TheCakeIsNaOH:
The Linux Dockerfile currently produces an image that includes the choco source code and all build artifacts. 

It would be a good idea to only include the final build artifacts the in the official image.

Issue #2363, opened 2021-09-17
Linux Docker container has warning about missing lib directory


TheCakeIsNaOH:
### What You Are Seeing?

When running choco in the linux docker container, it complains about not being able to find the `$env:chocolateyInstall/lib` directory. 

After installing a package, the warning disappears, since the `lib` directory has been created.

### What is Expected?

That it does not display a warning about that.

### How Did You Get This To Happen? (Steps to Reproduce)

1. Build `mono-choco` docker container
2. Run container
3. Run `choco`

### Suggested fix

Pre-create the `lib` directory inside the dockerfile

### Output Log

<details>
<summary>Full Log Output</summary>

<p>

~~~sh
Chocolatey v0.11.1-10-g2f554cbf
Unable to set trace logging:
 Object reference not set to an instance of an object
Chocolatey is running on Linux v 5.10.0.0
Attempting to delete file "usr/local/src/choco/code_drop/chocolatey/console/chocolatey.dll.old".
Attempting to delete file "/opt/chocolatey/choco.exe.old".
Attempting to create directory "/opt/chocolatey/helpers".
Attempting to create directory "/opt/chocolatey/helpers/functions".
Attempting to create directory "/opt/chocolatey/redirects".
Attempting to create directory "/opt/chocolatey/tools".
Command line: /opt/chocolatey/choco.exe --verbose --debug --trace --allow-unofficial
Received arguments: --verbose --debug --trace --allow-unofficial
RemovePendingPackagesTask is now ready and waiting for PreRunMessage.
Sending message 'PreRunMessage' out if there are subscribers...
[Pending] Removing all pending packages that should not be considered installed...
Directory '/opt/chocolatey/lib' does not exist.
Sending message 'PostRunMessage' out if there are subscribers...
Not able to cleanup NuGet temp folders. Failure was Path to combine cannot be empty. Tried to combine null with ''. Method called from 'delete_directory'
Exiting with 1

~~~

</p>

</details>

Issue #2362, opened 2021-09-17
Docker Mono add support for running choco.exe not just choco


TheCakeIsNaOH:
When running in a Linux docker container, it currently only adds `choco` to the path. 
On windows, running either `choco` or `choco.exe` works.
Therefore, the linux docker container should add `choco.exe` to the path to help compatibility with preexisting commands and scripts.

Issue #2361, opened 2021-09-17
Dockerfile add metadata with LABEL


TheCakeIsNaOH:
It would be good to add metadata to the docker container. The opencontainer spec for labeling probably should be used: https://github.com/opencontainers/image-spec/blob/main/annotations.md

Part of this is to update the maintainer information, as per the [Dockerfile reference](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated), the `MAINTAINER` instruction is deprecated. Instead, the maintainer should be set through the `LABEL` instruction.

PR #2365, opened 2021-09-17
(#2364) Rework Linux Dockerfile

TheCakeIsNaOH:
Rework the Dockerfile for Linux. Move to a multi-stage build so the
source code is not brought along into the final image (#2364). Add
build args to select the build script and mono version to use. Remove
deprecated MAINTAINER instruction and add metadata with LABEL (#2361).
Add "choco.exe" to path, so both "choco" and "choco.exe" can be used,
like Windows supports (#2362). Change the install location to
"/opt/chocolatey" as that is more in line with the Linux filesystem
hierarchy (#2360). Create the "lib" directory to prevent the warning
about the directory not being found (#2363).

The choco_official_wrapper is used for official builds, where the
"--allow-unofficial" argument is not needed.

Fixes: #2360
Fixes: #2361 
Fixes: #2362 
Fixes: #2363 
Fixes: #2364 

The debug and official build scripts are in #2359

### DIFF
--- a/docker/Dockerfile.linux
+++ b/docker/Dockerfile.linux
@@ -1,21 +1,43 @@
-FROM mono:6.12
-
-MAINTAINER Justin Phelps <linuturk@onitato.com>
+ARG monoversion="6.12"
+FROM mono:${monoversion} as build
 
 RUN apt-get update && apt-get install git -y
 
 COPY . /usr/local/src/choco/
 
 WORKDIR /usr/local/src/choco
-RUN chmod +x build.sh
-RUN chmod +x zip.sh
-RUN ./build.sh
+RUN chmod +x *.sh
 
-WORKDIR /usr/local/bin
-RUN ln -s /usr/local/src/choco/code_drop/chocolatey
+ARG buildscript="build.sh"
+RUN ./$buildscript
 
-COPY docker/choco_wrapper /usr/local/bin/choco
+RUN if [ "$buildscript" = "build.official.sh" ]; then \
+    cp docker/choco_official_wrapper code_drop/chocolatey/console/choco_wrapper; \
+  else \
+    cp docker/choco_wrapper code_drop/chocolatey/console/choco_wrapper; \    
+  fi;
 
-ENV ChocolateyInstall /usr/local/bin/chocolatey
+
+ARG monoversion="6.12"
+FROM mono:${monoversion} as install
+
+ARG monoversion="6.12"
+LABEL org.opencontainers.image.url="https://chocolatey.org/"
+LABEL org.opencontainers.image.documentation="https://docs.chocolatey.org/"
+LABEL org.opencontainers.image.source="https://github.com/chocolatey/choco"
+LABEL org.opencontainers.image.vendor="Chocolatey Software, Inc"
+LABEL org.opencontainers.image.licenses="Apache-2.0"
+LABEL org.opencontainers.image.title="Chocolatey"
+LABEL org.opencontainers.image.description="Chocolatey Client running on Mono"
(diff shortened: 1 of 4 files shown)


docker/Dockerfile.linux
```diff
@@ -1,21 +1,43 @@
...
+RUN ./$buildscript
+
+RUN if [ "$buildscript" = "build.official.sh" ]; then \
+    cp docker/choco_official_wrapper code_drop/chocolatey/console/choco_wrapper; \
+  else \
+    cp docker/choco_wrapper code_drop/chocolatey/console/choco_wrapper; \    
+  fi;
+
+
+ARG monoversion="6.12"
```
gep13, line 21: Does this need to be duplicated?

docker/Dockerfile.linux
```diff
@@ -1,21 +1,43 @@
...
+LABEL org.opencontainers.image.url="https://chocolatey.org/"
+LABEL org.opencontainers.image.documentation="https://docs.chocolatey.org/"
+LABEL org.opencontainers.image.source="https://github.com/chocolatey/choco"
+LABEL org.opencontainers.image.vendor="Chocolatey Software, Inc"
+LABEL org.opencontainers.image.licenses="Apache-2.0"
+LABEL org.opencontainers.image.title="Chocolatey"
+LABEL org.opencontainers.image.description="Chocolatey Client running on Mono"
+LABEL org.opencontainers.image.authors="https://chocolatey.org/support/"
```
gep13, line 32:
```suggestion
LABEL org.opencontainers.image.authors="https://chocolatey.org/support"
```
Otherwise, it goes to a 404

gep13 (changes requested):

docker/Dockerfile.linux
TheCakeIsNaOH: Yes, as the variable goes out of scope for every build stage (i.e. each `FROM), as per:
https://docs.docker.com/engine/reference/builder/#scope

docker/Dockerfile.linux
TheCakeIsNaOH: Whoops, I screwed that one up. Fixed.

docker/Dockerfile.linux
gep13: Gotcha, makes sense. Thanks for confirming.

gep13 (approved):

docker/README.md
```diff
@@ -8,5 +8,9 @@ To build this image yourself, follow these steps:
 1. Clone down the repository using `git clone https://github.com/chocolatey/choco.git`.
 1. Change directories to the root of the repository.
 1. Run the docker build command. `docker build -t mono-choco -f docker/Dockerfile.linux .` (the trailing . is important)
```
gep13, line 10: What is the significance of `mono-choco` here?  Is this the tag? Should this be something else, like `chocolatey/choco`?

gep13:
@TheCakeIsNaOH I rebased this PR on the head of the stable branch.

docker/README.md
TheCakeIsNaOH: That is the image name. The format is `<user>/<image name>:<tag>`, so that would be `chocolatey/mono-choco:latest`.

It probably should be changed to `choco:latest` or `choco:dev`. Then with the official build, you would use `chocolatey/choco:0.11.1`

I was holding off on changing it in the documentation until the naming had settled down a bit. Specifically, what tags to use.

gep13: Ok, makes sense. Thanks for the information!

gep13 (approved):
LGTM!

Issue #2366, opened 2021-09-17
Add a Windows Dockerfile


TheCakeIsNaOH:
Add a Windows Dockerfile.

The big question is what base image to pull from, as version compatibility appears to be a headache.
https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility

TheCakeIsNaOH:
Looking at the [available windows container images](https://docs.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/container-base-images), it looks like `mcr.microsoft.com/windows/servercore` is probably the right choice, as that is the one suggested for .Net framework applications. 

And looking at the version compatibility, newer windows versions can generally run older container images, but generally not vise-versa. Since windows containers started with server 2016, I presume that using the `ltsc2016` tag is the best tag to use.

TheCakeIsNaOH:
I'm thinking that the `Dockerfile` should be fairly simple, where choco is build on the host. Then, choco would be unpacked and installed inside the container from the `.nupkg` produced in `code_drop\nuget\`. 

The advantage of this would be a simple and quick container build, and the build of choco installed inside the container could be the exact same one that is uploaded elsewhere.

PR #2367, opened 2021-09-17
(#2366) Add Dockerfile for Windows

TheCakeIsNaOH:
Uses servercore ltsc2016 image for .Net framework compatibility.
Adds Install-ChocolateyInContainer helper script to install choco
directly from the nupkg. The script is based on what is used in the
test evironment, but it uses 7z.exe as the windows com object fails
when run in this container.

Fixes: #2366 

This is a draft at the moment as I'm not sure that this is the best way to go about creating a windows container, but I'd still be happy for feedback.


TheCakeIsNaOH:
What happened to these branches? Looks like stable and develop have different commits, but with the same changes.

gep13:
@JPRuskin did you have any input on this?  Thanks!

StefanScherer:
There are three major versions / tags of the servercore base image: `ltsc2016`, `ltsc2019` and now also `ltsc2022`.
Microsoft stops listing the old 2016 image at https://hub.docker.com/_/microsoft-windows-servercore
Maybe if you plan to provide an image it would make sense to provide a choco image for the more modern OS versions as well.

TheCakeIsNaOH:
> Maybe if you plan to provide an image it would make sense to provide a choco image for the more modern OS versions as well.

But containers built on older base images can run on newer versions of windows, correct? So the `ltsc2016` tag should be able to run on just about any version of windows that has containers?

StefanScherer:
> But containers built on older base images can run on newer versions of windows, correct? So the ltsc2016 tag should be able to run on just about any version of windows that has containers?

Yes, that's correct. But as ltsc2016 was the first version of Windows containers, Microsoft has improved a lot over time and shrank down the base image from > 4GB to about 2GB, so running on a newer Windows machine it just takes much much longer and I as a user think twice before pulling such old base images 😅 

But sure, we could start with just ltsc2016 and see what users will ask for.

PR #2381, opened 2021-09-27
(docs) Change incorrect version number.

mkevenaar:
When creating the PR to alow ChocolateyExitOnRebootDetected, it was
planned to be released on 0.10.16+, but that version was never released
Changed to the next version number that is actually released.


gep13:
@mkevenaar thank you for taking the time to get this updated!

PR #2403, opened 2021-10-14
(#2289) Update Rhino.Licensing dependency

### LINKED ISSUE #2289: Expired license causing FOSS choco to fail to run commands
### What You Are Seeing?

If you're running a licensed edition, and your license expires it is expected that Chocolatey falls back to FOSS operational mode. We raised and fixed this issue in https://github.com/chocolatey/choco/issues/1500, but it looks like we need to backport some of that work into the open source codebase.

### What is Expected?

An expired license should not prevent `choco install` etc to fail

vexx32:
## Summary

Update Rhino.Licensing dependency to get the fix for #2289 where it would crash the application after 5 minutes due to a timer not being properly deactivated and disposed.

This also includes an initial commit that replaces the current unversioned Rhino.Licensing dll with an identical but versioned dll so we can be more sure about where it came from & pull that version from our internal repository if we need to at a later date.

## Testing 

(Copied from https://github.com/chocolatey/rhino-licensing/pull/5)

As this was identified in Chocolatey with an expired license installed, the testing of this needs to be done with Chocolatey, an expired Chocolatey license, and Chocolatey.Extension installed.

This is the reproduction of the issue.

1. Create a test package:
```sh
git clone https://gist.github.com/d79ce5d3643a5be47ebcbea619c5cb2c.git
cd d79ce5d3643a5be47ebcbea619c5cb2c
choco pack .\test.nuspec
```
1. Ensure your Chocolatey license has expired.
1. Run: `choco install test -s . -y`
1. Note that prior to the package completing installation Chocolate throws an exception.

To apply the fix to a debug build of chocolatey and chocolatey.extension, what I ended up doing:

1. Use [everything](https://community.chocolatey.org/packages/everything) to find all files called `Rhino.Licensing.dll`
1. Delete all found `Rhino.Licensing.dll` files (make sure you're not deleting any you actually need. You likely only need to delete the ones within the `chocolatey` and `chocolatey.extension`, but I didn't have any other ones I needed so it was just easier to nuke them all.)
1. run `.\psake.ps1`
1. copy the `net40` build of `Rhino.Licensing.dll` to both `chocolatey` and `chocolatey.extension` lib folders
1. In `chocolatey.extension` code run `.\Get-ChocoUpdatedDebugVersion.ps1` to build and gather the Chocolatey executable
1. Build and run `chocolatey` from `chocolatey.extension` codebase with parameters: `install test -s <path to test.nupkg> -y`
1. Note that exception is no longer thrown

## Related Issue

Fixes #2289

## Changes Made

- [x] Dependency uplift
- [x] Bugfix

## Checklist

Due to the nature of the bug, I don't think it's reasonable to add a test to the codebase that always takes > 5 minutes to run for this.

### DIFF
--- a/lib/Rhino.Licensing.1.4.1/lib/net40/acknowledgements.txt
+++ b/lib/Rhino.Licensing.1.4.1/lib/net40/acknowledgements.txt
@@ -1,0 +1,1 @@
+Rhino Licensing is making use of the excellent log4net logging framework
--- a/lib/Rhino.Licensing.1.4.1/lib/net40/license.txt
+++ b/lib/Rhino.Licensing.1.4.1/lib/net40/license.txt
@@ -1,0 +1,25 @@
+Copyright (c) 2005 - 2009 Ayende Rahien (ayende@ayende.com)
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification,
+are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright notice,
+    this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright notice,
+    this list of conditions and the following disclaimer in the documentation
+    and/or other materials provided with the distribution.
+    * Neither the name of Ayende Rahien nor the names of its
+    contributors may be used to endorse or promote products derived from this
+    software without specific prior written permission.
+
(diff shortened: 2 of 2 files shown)


gep13:
@vexx32 I have taken this for a spin this morning, and it works exactly as described.  Thanks to you and @corbob for getting this fixed up.

I did add one additional commit into this PR to add back in the acknowledgements and license files, as it is important that we don't lose those.  Can you please update the nupkg so that future releases also contain those files?

PR #2430, opened 2021-10-29
(#62) Default push source config

### LINKED ISSUE #62: Ability to specify a default source when doing choco push
Use Case:

Default, out of the box, when you do `choco push` you get a warning about missing source

As a User, I want to be able to specify default source, so I don't have to always do `choco push -source http://chocolatey.org/`

TheCakeIsNaOH:
## Description Of Changes

Adds a new config value DefaultPushSource. This config value is used to set the default source used for
choco push. It goes into config.PushCommand.DefaultSource

If DefaultPushSource is set, it should set the source assuming that
a source is not passed in explicitly with --source. If a source is
passed in explicitly, then DefaultPushSource should not override it.

If --api-key is not passed and a api key has not been saved for the
DefaultPushSource, it will fail with the standard message about an api
key being required.

If DefaultPushSource is set to "disabled", it should throw an error if
there is not an explicit source passed in. But if an explicit source is
passed in, it should not error.

## Motivation and Context

The DefaultPushSource is useful for users that have an internal repository that they want to normally want to push to.

The behavior if DefaultPushSource is set to disabled is useful if a user wants has multiple repositories and wants to always have to explicitly specify a source so as to make sure to push package to the correct source.

## Testing

1. Build choco
1. Copy a nupkg next to choco.exe (so as to not have to specify it every time)
1. Run `choco push --allow-unofficial` to validate that it still defaults to `https://push.chocolatey.org` if `DefaultPushSource` is not set.
1. Run `choco config set DefaultPushSource --value=disabled  --allow-unofficial` to force passing an explicit source
1. Run `choco push --allow-unofficial` and it should fail, asking for you to pass an explicit `--source`
1. Run `choco push --source=https://push.chocolatey.org/ --allow-unofficial` to validate that it will not throw if an explicit source is passed
1. Run `choco config set DefaultPushSource --value=chocolatey  --allow-unofficial` 
1. Run `choco push --allow-unofficial` to validate that the name `chocolatey` will resolve to `https://community.chocolatey.org/api/v2/`
1. Run `choco config set DefaultPushSource --value=https://community.chocolatey.org/api/v2/ --allow-unofficial` 
1. Run `choco push --allow-unofficial` to validate that it will try to push to the `https://community.chocolatey.org/api/v2/`

## Change Types Made

## Related Issue

Fixes #62

Depends on #1511

## Change Checklist

### DIFF
--- a/src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
+++ b/src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
@@ -267,6 +267,7 @@ namespace chocolatey.infrastructure.app.builders
             config.Proxy.BypassOnLocal = set_config_item(ApplicationParameters.ConfigSettings.ProxyBypassOnLocal, configFileSettings, "true", "Bypass proxy for local connections. Available in 0.10.4+.").is_equal_to(bool.TrueString);
             config.UpgradeCommand.PackageNamesToSkip = set_config_item(ApplicationParameters.ConfigSettings.UpgradeAllExceptions, configFileSettings, string.Empty, "A comma-separated list of package names that should not be upgraded when running `choco upgrade all'. Defaults to empty. Available in 0.10.14+.");
             config.DefaultTemplateName = set_config_item(ApplicationParameters.ConfigSettings.DefaultTemplateName, configFileSettings, string.Empty, "Default template name used when running 'choco new' command. Available in 0.12.0+.");
+            config.PushCommand.DefaultSource = set_config_item(ApplicationParameters.ConfigSettings.DefaultPushSource, configFileSettings, string.Empty, "Default source to push packages to when running 'choco push' command. Available in 0.12.0 +.");
         }
 
         private static string set_config_item(string configName, ConfigFileSettings configFileSettings, string defaultValue, string description, bool forceSettingValue = false)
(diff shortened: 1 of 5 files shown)


TheCakeIsNaOH:
I've rebased this on top of the commits from #1511, and it actually works without any further code changes. I've added a unit test for it.

vexx32 (approved):
This looks great, thanks for sorting this out! 💖

src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
```diff
@@ -267,6 +267,7 @@ private static void set_config_items(ChocolateyConfiguration config, ConfigFileS
             config.Proxy.BypassOnLocal = set_config_item(ApplicationParameters.ConfigSettings.ProxyBypassOnLocal, configFileSettings, "true", "Bypass proxy for local connections. Available in 0.10.4+.").is_equal_to(bool.TrueString);
             config.UpgradeCommand.PackageNamesToSkip = set_config_item(ApplicationParameters.ConfigSettings.UpgradeAllExceptions, configFileSettings, string.Empty, "A comma-separated list of package names that should not be upgraded when running `choco upgrade all'. Defaults to empty. Available in 0.10.14+.");
             config.DefaultTemplateName = set_config_item(ApplicationParameters.ConfigSettings.DefaultTemplateName, configFileSettings, string.Empty, "Default template name used when running 'choco new' command. Available in 0.12.0+.");
+            config.PushCommand.DefaultSource = set_config_item(ApplicationParameters.ConfigSettings.DefaultPushSource, configFileSettings, string.Empty, "Default source to push packages to when running 'choco push' command. Available in 0.12.0 +.");
```
gep13, line 270: @vexx32 can you update this line with the correct version number?

vexx32: *blinks*

Dang, good catch. Will do, thank you!

gep13:
This PR included a commit associated with #63, and that commit needs to be removed.  The reason for this being removed in covered in this [comment](https://github.com/chocolatey/choco/issues/63#issuecomment-1514404424).

PR #2433, opened 2021-10-30
(#2432) Actions workflow run integration tests

### LINKED ISSUE #2432: Run integration tests on github actions workflow.
Currently, the Appveyor CI is running the integration tests, but the GitHub actions workflow is not. Let's get both running the integration tests.

TheCakeIsNaOH:
## Description Of Changes

- Adds a `full` target to the uppercut testing, which runs all tests but `NotWorking` and `Ignore` tests. It can be called with `test.bat full`
- Adds a `test.sh` script for non-Windows systems. 
- Adds a scheduled workflow to run `test.bat full` on the Windows, runs `test.sh` on non-Windows. 

## Motivation and Context

Running the integration tests on the Actions workflow is needed for feature parity with the Appveyor CI run.

## Testing

N/A, see github actions workflow

## Change Types Made

## Related Issue

Fixes #2432

## Change Checklist

### DIFF
--- a/.build/analyzers/nunit.test.step
+++ b/.build/analyzers/nunit.test.step
@@ -24,6 +24,8 @@
   <target name="go" depends="run_tasks" />
 
   <target name="run_all_tests" depends="prepare, set_nunit_failure_mode, get_test_dlls_all, run_tests, generate_report" description="Like go but runs all tests" />
+  
+  <target name="run_integration_tests" depends="prepare, set_nunit_failure_mode, get_test_dlls_integration, run_tests, generate_report" description="Like go but runs the integration tests" />
 
   <!-- run_normal_tasks is called from run_tasks unless replace extension exists -->
   <target name="run_normal_tasks"
@@ -96,6 +98,33 @@
 
     <property name="args.test_runner" value="${dlls.test} ${test.args}" />
   </target>
+
+  <target name="get_test_dlls_integration">
+    <echo level="Warning" message="Getting all test dlls (including integration tests) based on name in directory ${path::get-full-path(dirs.build.code)}." />
+    <property name="dll.names" value="" />
+    <foreach item="File" property="dll.filename">
+      <in>
+        <items>
+          <exclude name="${dirs.build.code}/lib/**" />
+          <exclude name="${dirs.build.code}/_PublishedApplications/**" />
+          <exclude name="${dirs.build.code}/**/TestFu.dll" />
+          <include name="${dirs.build.code}/**/*Test*.Integration*dll" />
+          <include name="${dirs.build.code}/**/*test*.integration*dll" />
+          <include name="${dirs.build.code}/**/*Spec*.Integration*dll" />
+          <include name="${dirs.build.code}/**/*spec*.integration*dll" />
+        </items>
+      </in>
+      <do>
+        <property name="dll.names" value="${dll.names + ' ' + string::replace(string::replace(dll.filename,path::get-full-path(dirs.build.code) + '\',''),'\',path.separator) + ''}" />
+      </do>
+    </foreach>
+
+    <property name="dlls.test" value="${dll.names}" />
+    <echo level="Warning" message="Running NUnit against these test dlls - ${dlls.test}." />
+
+    <property name="args.test_runner" value="${dlls.test} ${test.args} ${nunit.separator}exclude=&quot;NotWorking,Ignore,notworking,ignore&quot;" />
+  </target>
+
 
   <target name="run_tests" depends="prepare" description="Running Unit Tests">
     <if test="${dlls.test != ''}">
--- a/.build/analyzers/test.step
+++ b/.build/analyzers/test.step
@@ -35,6 +35,11 @@
     <call target="set_normal_failure_mode" />
   </target>
 
+  <target name="integration">
+    <nant buildfile="${dirs.current.file}${path.separator}nunit.test.step" target="run_integration_tests" inheritall="true" if="${test.framework == 'nunit'}" />
+    <call target="set_normal_failure_mode" />
+  </target>
+
   <target name="open_results">
     <nant buildfile="${dirs.current.file}${path.separator}mbunit2.test.step" target="open_results" inheritall="true" if="${test.framework == 'mbunit2'}" />
     <nant buildfile="${dirs.current.file}${path.separator}gallio.test.step" target="open_results" inheritall="true" if="${test.framework == 'gallio'}" />
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,15 +1,9 @@
 name: Chocolatey Builds
 
 on:
-  # Trigger on pushes to master and develop, or with pull requests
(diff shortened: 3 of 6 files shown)


test.sh
```diff
@@ -0,0 +1,30 @@
...
+# stty -echo
+
+# ::Project UppercuT - http://uppercut.googlecode.com
+# ::No edits to this file are required - http://uppercut.pbwiki.com
+
+function usage
+{
+	echo ""
+	echo "Usage: test.sh"
+    echo "Usage: test.sh all - to run all tests"
```
gep13, line 11: The formatting here seems a little out

test.bat
```diff
@@ -29,6 +29,7 @@ goto finish
 echo.
 echo Usage: test.bat
 echo Usage: test.bat all - to run all tests
+echo Usage: test.bat full - to run all tests but notworking and ignored tests
```
gep13, line 32: I am not sure of the description here, both say that they run all tests.  Should we mention here specifically the execution of integration tests?

.github/workflows/build.yml
```diff
@@ -40,6 +44,8 @@ jobs:
        run: |
           chmod +x build.sh
           $GITHUB_WORKSPACE//build.sh -v
+     - name: Test with Nunit on Mono
+       run: $GITHUB_WORKSPACE//test.sh
```
gep13, line 48: Do we need the `full` argument here, like on line 37 above?

.github/workflows/build.yml
```diff
@@ -24,13 +24,17 @@ jobs:
       run: |
         chmod +x build.sh
         $GITHUB_WORKSPACE//build.sh -v
+    - name: Test with Nunit on Mono
+      run: $GITHUB_WORKSPACE//test.sh
```
gep13, line 28: Do we need the full argument as per line 37 below?

.build/analyzers/test.step
```diff
@@ -35,6 +35,11 @@
     <call target="set_normal_failure_mode" />
   </target>
 
+  <target name="full">
+    <nant buildfile="${dirs.current.file}${path.separator}nunit.test.step" target="run_full_tests" inheritall="true" if="${test.framework == 'nunit'}" />
```
gep13, line 39: Is there a reason that you didn't add the other buildfile's here?  Granted, they are unlikely to be used, but just curious...

.build/analyzers/nunit.test.step
```diff
@@ -24,6 +24,8 @@
   <target name="go" depends="run_tasks" />
 
   <target name="run_all_tests" depends="prepare, set_nunit_failure_mode, get_test_dlls_all, run_tests, generate_report" description="Like go but runs all tests" />
+  
+  <target name="run_full_tests" depends="prepare, set_nunit_failure_mode, get_test_dlls_full, run_tests, generate_report" description="Like go but runs a full set of the working tests" />
```
gep13, line 28: Part of me wonders if this should be renamed to something like `run_integration_tests`, and narrow the scope of the test assemblies that it finds to only the ones with integration in the name.

This would mean that we could then choose to run all unit tests, and then all integration tests.

Thoughts?

gep13 (changes requested):

.github/workflows/build.yml
TheCakeIsNaOH: The integration tests need more work on non-windows platforms. 
I am planning to add it here: https://github.com/chocolatey/choco/pull/2435

.github/workflows/build.yml
TheCakeIsNaOH: The integration tests need more work on non-windows platforms. 
I am planning to add it here: https://github.com/chocolatey/choco/pull/2435

test.bat
TheCakeIsNaOH: Yup, sounds good.

test.sh
TheCakeIsNaOH: Yup, will fix

.build/analyzers/test.step
TheCakeIsNaOH: I just didn't want to try to implement `run_full_tests` for the files for the other test framework.

.build/analyzers/nunit.test.step
TheCakeIsNaOH: Hmm, that could work. Now that I am thinking about it, that would be good.

It would speed up the CI builds by a couple of seconds, as the unit tests are already run by `build.sh/cmd`, so they would not be run twice.

.build/analyzers/test.step
gep13: Understood.  Just wanted to check.  I think that this is fine.

test.sh
TheCakeIsNaOH: Done

test.bat
TheCakeIsNaOH: Done

.build/analyzers/nunit.test.step
TheCakeIsNaOH: This has been updated, hopefully it is working correctly.

TheCakeIsNaOH: Ok, that is better, the casing was screwed up for the linux build.

gep13:
@TheCakeIsNaOH so, in principal, this looks really good to me.  I have taken this for a spin locally, and I get the following results:

* test.bat - 631 tests
* test.bat integration - 757 tests
* test.bat all - 1388 tests

The only remaining concern I have is adding the integration tests to each build execution.  Given the increased amount of time that the integration tests take to complete, I am more inclined to create another time trigger based workflow that runs the integration tests (say every evening), and which can also be triggered manually.  That way, we continue to have the reasonably quick execution of the build and the unit tests, and then we get the knowledge that integration tests will be executed later, to ensure that nothing has broken there.

What do you think?

TheCakeIsNaOH:
Hmm, that sounds ok.  What about appveyor?

gep13:
@TheCakeIsNaOH I am actually starting to wonder whether we need to AppVeyor build.  What are your thoughts here?

TheCakeIsNaOH:
@gep13 The reason it was kept around when the github actions workflow was added was because it uploaded artifacts and code coverage results. I am actually working on getting artifacts uploaded right now.

Are the code coverage results still being used?

TheCakeIsNaOH:
I've added a commit here to upload the build artifacts, and it seems to be working correctly.

gep13:
> @TheCakeIsNaOH said...
Are the code coverage results still being used?

Having the code coverage results being corrected reported is definitely something that I would like to see, and ideally, we would look to add a badge to the readme to indicate what the current coverage is.

Looking at the coveralls page for this project:

https://coveralls.io/github/chocolatey/choco

It looks like this hasn't been updated since 2019, so I think that there is some work required here in order to get this working again.  I would see this as a separate issue though.

gep13 (approved):
LGTM!

gep13:
@TheCakeIsNaOH your changes have been merged, thanks for your contribution 👍

PR #2482, opened 2021-12-04
(#2481) Only allow installing template packages on non-Windows

### LINKED ISSUE #2481: Warn about unuspported packages on non-Windows platforms
After some discussion with @gep13, there is a concern that users running on non-Windows system will install packages, assume that they are for the native system instead of for Windows systems, and then create an extra support burden.

Given that only template packages are currently intended to be installed on non-Windows platforms, other package types should not be able to be installed by default. If any other types of packages are attempted to be installed/upgraded, a helpful error message should be displayed. 

This error should be able to be bypassed if `--force` is used, so if someone knows what they are doing, they can still install other types of packages.

Depending on what happens long term with the licensed extension, support for `.extension` packages could be added as well. The installation of `.extension` packages is working, but the usage of extensions is not. This is where [`.license` packages](https://github.com/chocolatey/choco/issues/2480) would be useful, as they could also be put on the allow list.

TheCakeIsNaOH:
## Description Of Changes

This adds a check for non-Windows systems that only allows package ids
ending in .template to install. 

The check can be overridden if --force is specified. Because
upgrade_run calls install_run if the package is not already installed,
there is no need to duplicate this check in upgrade_run.

A bool in application parameters was also added which is used to disable the
check during install for supported package IDs. This bool is shifted
during testing so the integration tests will still run as normal,
without failing due to the package ID check.

## Motivation and Context

This check is added to help provide guidance to users, as templates are the only package type currently supported on non-Windows systems.

In the future, if the licensed extension becomes supported on
non-Windows systems, then the check will need to be amended to support
installing packages relating to licensed extension.

## Testing

- Ran `choco install wget` on Windows to validate that the check is bypassed on Windows systems
- Ran `choco install curl` on Linux to validate that it fails with the specified error message.
- Ran `choco install msi.template` on Linux to validate that it installs without the error.
- Ran `choco install curl -f` on Linux to validate that the check can be bypassed if force is specified.
- Ran integration tests on both Windows and a Linux system.

## Change Types Made

* [x] Breaking change (fix or feature that could cause existing functionality to change) (Breaks non-Windows systems only)

## Related Issue

Fixes #2481

## Change Checklist

* [x] Documentation has been updated https://github.com/chocolatey/docs/pull/280

### DIFF
--- a/src/chocolatey.tests.integration/NUnitSetup.cs
+++ b/src/chocolatey.tests.integration/NUnitSetup.cs
@@ -102,6 +102,9 @@ namespace chocolatey.tests.integration
             field = typeof(ApplicationParameters).GetField("LockTransactionalInstallFiles");
             field.SetValue(null, false);
 
+            field = typeof(ApplicationParameters).GetField("CheckPackageIdsNonWindows");
+            field.SetValue(null, false);
+
             // we need to speed up specs a bit, so only try filesystem locking operations twice
             field = fileSystem.GetType().GetField("TIMES_TO_TRY_OPERATION", BindingFlags.Instance | BindingFlags.NonPublic);
             if (field != null)
--- a/src/chocolatey/infrastructure.app/ApplicationParameters.cs
+++ b/src/chocolatey/infrastructure.app/ApplicationParameters.cs
@@ -150,6 +150,11 @@ namespace chocolatey.infrastructure.app
         /// </summary>
         public static readonly bool AllowPrompts = true;
 
+        /// <summary>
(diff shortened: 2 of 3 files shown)


TheCakeIsNaOH:
I'm not sure if there any any tests I should add for this.

TheCakeIsNaOH:
After discussion with @gep13 via @ferventcoder, only a warning should be shown, instead of throwing. Will create a new PR, will leave this one for posterity if someone needs to come back to it.

PR #2484, opened 2021-12-05
(#1443) Reset config via action after every package upgrade

### LINKED ISSUE #1443: Upgrade all reuses overridden package parameters when useRememberedArgumentsForUpgrades feature is turned on

TheCakeIsNaOH:
## Description Of Changes

This adds a parameter to the upgrade_run method, which
is used to pass an action which resets the configuration in the
function that calls upgrade_run.

## Motivation and Context

This fixes a bug when useRememberedArguments is enabled that causes
arguments to be reused in the chocolateyInstall.ps1 when multiple
packages are upgraded. It happens because OptionSet.parse() sets the
configuration via the actions specified in the UpgradeCommand optionSet
setup (and likewise in configuration builder as well). It only sets
options that are saved, so if a later package does not have an option,
it is not set, and the previous option is reused.

This fixes the bug because it resets the configuration at the
ChocolateyPackageService level, which is enough to reset the
configuration for action that runs the PowerShell as well.

## Testing

1. Build choco
1. Run these setup steps:
```
.\choco.exe feature enable -n useRememberedArgumentsForUpgrades --allow-unofficial -y
.\choco.exe feature enable -n allowGlobalConfirmation --allow-unofficial -y
.\choco.exe install curl --package-parameters="'/CurlOnlyParam'" --version=7.77.0 --allow-unofficial -y -f --ia="'/CurlIAParam'" --forcex86
.\choco.exe install wget --version=1.21.1 --allow-unofficial -y -f
```
Note that curl has package parameters, install arguments and forcex86 set, while wget does not have any of those.

3. Run this: `.\choco.exe upgrade all --allow-unofficial -y --debug`

The curl install script should be run with the package parameters, install arguments and forcex86 similar to this:
```
& '.\helpers\chocolateyScriptRunner.ps1' -packageScript '.\lib\curl\tools\chocolateyInstall.ps1' -installArguments '/CurlIAParam' -packageParameters '/CurlOnlyParam' -forceX86']
```
And the wget install script should be run without those arguments:
```
& '.\helpers\chocolateyScriptRunner.ps1' -packageScript '.\lib\Wget\tools\chocolateyinstall.ps1' -installArguments '' -packageParameters ''']
```

## Change Types Made

* [ ] Breaking change (fix or feature that could cause existing functionality to change)s

## Related Issue

Fixes #1443

## Change Checklist

### DIFF
--- a/src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
+++ b/src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
@@ -30,6 +30,9 @@ namespace chocolatey.infrastructure.app.configuration
     [Serializable]
     public class ChocolateyConfiguration
     {
+        [NonSerialized]
+        private ChocolateyConfiguration _originalConfiguration;
+
         public ChocolateyConfiguration()
         {
             RegularOutput = true;
@@ -56,6 +59,86 @@ namespace chocolatey.infrastructure.app.configuration
 #if DEBUG
             AllowUnofficialBuild = true;
 #endif
+        }
+
+        /// <summary>
+        /// Creates a backup of the current version of the configuration class.
+        /// </summary>
+        /// <exception cref="System.Runtime.Serialization.SerializationException">One or more objects in the class or child classes are not serializable.</exception>
+        public void start_backup()
+        {
+            // We do this the easy way to ensure that we have a clean copy
+            // of the original configuration file.
+            _originalConfiguration = this.deep_copy();
+        }
+
+        /// <summary>
+        /// Restore the backup that has previously been created to the initial
+        /// state, without making the class reference types the same to prevent
+        /// the initial configuration class being updated at the same time if a
+        /// value changes.
+        /// </summary>
+        /// <param name="removeBackup">Whether a backup that was previously made should be removed after resetting the configuration.</param>
+        /// <exception cref="InvalidOperationException">No backup has been created before trying to reset the current configuration, and removal of the backup was not requested.</exception>
+        /// <remarks>
+        /// This call may make quite a lot of allocations on the Gen0 heap, as such
+        /// it is best to keep the calls to this method at a minimum.
+        /// </remarks>
+        public void reset_config(bool removeBackup = false)
+        {
+            if (_originalConfiguration == null)
+            {
+                if (removeBackup)
+                {
+                    // If we will also be removing the backup, we do not care if it is already
+                    // null or not, as that is the intended state when this method returns.
+                    return;
+                }
+
+                throw new InvalidOperationException("No backup has been created before trying to reset the current configuration, and removal of the backup was not requested.");
+            }
+
+            var t = this.GetType();
+
+            foreach (var property in t.GetProperties(BindingFlags.Public | BindingFlags.Instance))
+            {
+                try
+                {
+                    var originalValue = property.GetValue(_originalConfiguration, new object[0]);
+
+                    if (removeBackup || property.DeclaringType.IsPrimitive || property.DeclaringType.IsValueType || property.DeclaringType == typeof(string))
+                    {
+                        // If the property is a primitive, a value type or a string, then a copy of the value
+                        // will be created by the .NET Runtime automatically, and we do not need to create a deep clone of the value.
+                        // Additionally, if we will be removing the backup there is no need to create a deep copy
+                        // for any reference types, as such we also set the reference itself so it is not needed
+                        // to allocate more memory.
+                        property.SetValue(this, originalValue, new object[0]);
+                    }
+                    else if (originalValue != null)
+                    {
+                        // We need to do a deep copy of the value so it won't copy the reference itself,
+                        // but rather the actual values we are interested in.
+                        property.SetValue(this, originalValue.deep_copy(), new object[0]);
+                    }
+                    else
+                    {
+                        property.SetValue(this, null, new object[0]);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    throw new ApplicationException("Unable to restore the value for the property '{0}'.".format_with(property.Name), ex);
+                }
+            }
+
+            if (removeBackup)
+            {
+                // It is enough to set the original configuration to null to
+                // allow GC to clean it up the next time it runs on the stored Generation
+                // Heap Table.
+                _originalConfiguration = null;
+            }
         }
 
         // overrides
@@ -237,6 +320,7 @@ NOTE: Hiding sensitive configuration data! Please double and triple
 
         [Obsolete("Side by Side installation is deprecated, and is pending removal in v2.0.0")]
         public bool AllowMultipleVersions { get; set; }
+
         public bool AllowDowngrade { get; set; }
         public bool ForceDependencies { get; set; }
         public string DownloadChecksum { get; set; }
--- a/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
+++ b/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
@@ -1,4 +1,4 @@
-// Copyright © 2017 - 2022 Chocolatey Software, Inc
+﻿// Copyright © 2017 - 2022 Chocolatey Software, Inc
 // Copyright © 2011 - 2017 RealDimensions Software, LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -22,7 +22,6 @@ namespace chocolatey.infrastructure.app.services
     using System.IO;
     using System.Linq;
     using System.Text;
-    using builders;
     using commandline;
     using configuration;
     using domain;
@@ -31,8 +30,8 @@ namespace chocolatey.infrastructure.app.services
     using infrastructure.events;
     using infrastructure.services;
     using logging;
-    using NuGet;
     using nuget;
+    using NuGet;
     using platforms;
     using results;
     using tolerance;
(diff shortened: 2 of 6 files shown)


src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
```diff
@@ -839,7 +841,7 @@ public void upgrade_noop(ChocolateyConfiguration config)
                 get_environment_before(config, allowLogging: true);
 
                 var beforeUpgradeAction = new Action<PackageResult>(packageResult => before_package_modify(packageResult, config));
-                var results = perform_source_runner_function(config, r => r.upgrade_run(config, action, beforeUpgradeAction));
+                var results = perform_source_runner_function(config, r => r.upgrade_run(config, action, resetConfigAction, beforeUpgradeAction));
```
AdmiringWorm, line 844: Any reason you decided to change the `ISourceRunner` interface instead of resetting the configuration inside the `perform_source_runner_function` method?

It sounds a bit wrong to me having to update the interface just to reset the configuration in the `NugetService`

TheCakeIsNaOH: It's been long enough since I made this PR that I don't remember exactly, but I think I tried that, and it did not work out.
I'd agree that not changing the interface would be better, so if you can get it to work that would be great.

AdmiringWorm: I have replaced the commit with alternative handling (added you as the co-maintainer).

If you could have a test run of the change to verify it still works as you expect, I would appreciate it.

I used the testing steps you have in the PR body as the base of the changes.

If you check it out sometime today, could you reference @corbob to make the review and final merge of the changes?

I have backed up the changes you made and pushed them up to a different branch just in case.

TheCakeIsNaOH: Your commit seems to work as well as my previous version did, however, I found another bug that affects both versions, see below comment.

TheCakeIsNaOH:
It seems like this only remembers bool switches (like `--forcex86`) for the first package that uses them.

```
.\choco.exe feature enable -n useRememberedArgumentsForUpgrades --allow-unofficial -y
.\choco.exe feature enable -n allowGlobalConfirmation --allow-unofficial -y
.\choco.exe install curl --package-parameters="'/CurlOnlyParam'" --version=7.77.0 --allow-unofficial -y -f --ia="'/CurlIAParam'" --forcex86
.\choco.exe install wget --version=1.21.1 --allow-unofficial -y -f --forcex86
.\choco.exe upgrade all --allow-unofficial -y --debug
```

And during that `upgrade all`, the curl package will extract it's 32 bit archive, but the wget package will extract it's 64 bit archive. The wget package should have the `--forcex86` argument set, but it appears like that it was not set.

AdmiringWorm:
> It seems like this only remembers bool switches (like `--forcex86`) for the first package that uses them.
> 
> ```
> .\choco.exe feature enable -n useRememberedArgumentsForUpgrades --allow-unofficial -y
> .\choco.exe feature enable -n allowGlobalConfirmation --allow-unofficial -y
> .\choco.exe install curl --package-parameters="'/CurlOnlyParam'" --version=7.77.0 --allow-unofficial -y -f --ia="'/CurlIAParam'" --forcex86
> .\choco.exe install wget --version=1.21.1 --allow-unofficial -y -f --forcex86
> .\choco.exe upgrade all --allow-unofficial -y --debug
> ```
> 
> And during that `upgrade all`, the curl package will extract it's 32 bit archive, but the wget package will extract it's 64 bit archive. The wget package should have the `--forcex86` argument set, but it appears like that it was not set.

Hmm, that sounds like we would maybe want to wait to pull this PR into the repository.

I'll bring it up internally tomorrow to see where we want to go.

AdmiringWorm:
I have moved this PR into draft, while the code and tests are being worked on to prevent any merges happening

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
```diff
@@ -58,6 +61,86 @@ public ChocolateyConfiguration()
 #endif
         }
 
+        /// <summary>
+        /// Creates a backup of the current version of the configuration class.
+        /// </summary>
+        /// <exception cref="System.Runtime.Serialization.SerializationException">One or more objects in the class or child classes are not serializable.</exception>
+        public void start_backup()
+        {
+            // We de this the easy way to ensure that we have a clean copy
```
corbob, line 70:
```suggestion
            // We do this the easy way to ensure that we have a clean copy

```

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
```diff
@@ -58,6 +61,86 @@ public ChocolateyConfiguration()
...
+        /// <summary>
+        /// Restore the backup that has previously been created to the initial
+        /// state, without making the class reference types the same to prevent
+        /// the initial configuration class being updated at the same time if a
+        /// value changes.
+        /// </summary>
+        /// <param name="removeBackup">Whether a backup that was previously made should be removed after resetting the configuration.</param>
+        /// <exception cref="InvalidOperationException">No backup has been created before trying to reset the current configuration, and removal of the backup was not requested.</exception>
+        /// <remarks>
+        /// This call may make quite alot of allocations on the Gen0 heap, as such
```
corbob, line 84:
```suggestion
        /// This call may make quite a lot of allocations on the Gen0 heap, as such

```

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
```diff
@@ -58,6 +61,86 @@ public ChocolateyConfiguration()
...
+                try
+                {
+                    var originalValue = property.GetValue(_originalConfiguration, new object[0]);
+
+                    if (removeBackup || property.DeclaringType.IsPrimitive || property.DeclaringType.IsValueType || property.DeclaringType == typeof(string))
+                    {
+                        // If the property is a primitive, a value type or a string, then a copy of the value
+                        // will be created by the .NET Runtime automatically, and we do not need to create a deep clone of the value.
+                        // Additionally, if we will be removing the backup there is no need to create a deep copy
+                        // for any reference types, as such we also sets the reference itself so it is not needed
```
corbob, line 114:
```suggestion
                        // for any reference types, as such we also set the reference itself so it is not needed

```

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
```diff
@@ -58,6 +61,86 @@ public ChocolateyConfiguration()
...
+                        property.SetValue(this, originalValue.deep_copy(), new object[0]);
+                    }
+                    else
+                    {
+                        property.SetValue(this, null, new object[0]);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    this.Log().Debug("Unable to restore the value for property '{0}' with the message '{1}'", property.Name, ex.Message);
```
corbob, line 131: Is this possibly a scenario where a more present message should be presented?

I guess the overarching question I have is: what are the possible consequences if a value is unable to be restored?

corbob (commented):
Just a few spelling points in the comments.

Also a question: Does this need to be updated in chocolatey licensed extension as well?

AdmiringWorm:
> Also a question: Does this need to be updated in chocolatey licensed extension as well?

No, there is no need to update anything in the Licensed codebase.

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
AdmiringWorm: > Is this possibly a scenario where a more present message should be presented?
> 
> I guess the overarching question I have is: what are the possible consequences if a value is unable to be restored?

The consequence if the value can not be restored would be that it would still contain the value that was changed.
However, I was not able to produce a case where this would fail in the current codebase, as such I only added a debug logging for this in the rare case it could fail.

AdmiringWorm: I have changed this to rethrow the exception as an application exception instead of silently ignoring it.

corbob (dismissed):
LGTM!

corbob (approved):

corbob:
Thanks for getting this added @AdmiringWorm. I'll let the GitHub Actions finish, then kick the one that failed and once that's all green I'll get this merged 😁

PR #2504, opened 2022-01-05
(#886) Add support for more arguments in packages.config

TheCakeIsNaOH:
## Description Of Changes

This adds the remaining arguments in the install command's OptionSet and some more arguments in the global OptionSet as elements to the packages.config xml serialization schema. 

## Motivation and Context

This is useful for users wanting arguments that were previously not
available. It is also a pre-requisite for exporting remembered 
arguments, as some of the remembered arguments were not available
in the previous available schema.

Some of the global arguments were not added:
- debug
- verbose 
- trace
- nocolor
- noop
- allowunofficial
- proxy
- proxy-user
- proxy-password
- proxy-bypass-list
- proxy-bypass-on-local
- log-file

All of these appeared to either not be applicable to a single package (e.g. proxy related) or required earlier than the packages.config would get read (like log-file).

## Testing

Save this as a packages.config file:
```xml
<?xml version="1.0" encoding="utf-8"?>
<packages>
   <package id="iperf2" prerelease="true" overrideArguments="true"
     notSilent="true" allowDowngrade="true" forceDependencies="true"
     skipAutomationScripts="true" user="string" password="string" cert="string"
     certPassword="string" ignoreChecksums="true" allowEmptyChecksums="true" 
     allowEmptyChecksumsSecure="true" requireChecksums="true" 
     downloadChecksum="string" downloadChecksum64="string" 
     downloadChecksumType="sha512" downloadChecksumType64="sha512"
     ignorePackageExitCodes="true" usePackageExitCodes="true"
     stopOnFirstFailure="true" exitWhenRebootDetected="true"
     ignoreDetectedReboot="true" disableRepositoryOptimizations="true"
     acceptLicense="true" confirm="true" limitOutput="true" cacheLocation="Z:\"
     failOnStderr="true" useSystemPowershell="true" noProgress="true"
   />
</packages>
```

Set the debugging arguments as `install \path\to\packages.config`

Then set a breakpoint at `get_packages_from_config` in `ChocolateyPackageService`, and step through the method, ensuring that the settings in the `packageConfig` get updated correctly.

## Change Types Made

* [ ] Bug fix (non-breaking change)
* [x] Feature / Enhancement (non-breaking change)
* [ ] Breaking change (fix or feature that could cause existing functionality to change)

## Related Issue

Fixes #886
Depends on #2187

## Change Checklist

* [x] Requires a change to the documentation
* [ ] Documentation has been updated (Tracked in https://github.com/chocolatey/docs/issues/403)
[...]

PR #2506, opened 2022-01-05
Export remembered args

### LINKED ISSUE #2503: Add option to include install-args in the `choco export` command
It would be nice to have a flag similar to the `--include-version-numbers` flag for the `choco export` command but for including the installation arguments of a given package (assuming the user has the configuration `useRememberedArgumentsForUpgrades` enabled).

Consider the following scenario where I install Git with various installation arguments applied:

```powershell
cinst -y git.install --installargs "'/COMPONENTS=icons,assoc,assoc_sh,autoupdate,windowsterminal,scalar'" --params "/GitAndUnixToolsOnPath /NoAutoCrlf /WindowsTerminal /NoShellIntegration /NoGuiHereIntegration /NoShellHereIntegration /WindowsTerminalProfile /DefaultBranchName:main /Editor:VisualStudioCode"
```

then run `choco export -o="choco-packages.config"`, however the output XML only contains:

```XML
<package id="git.install"/>
```

whereas it could contain something like so:

```XML
<package id="git.install" arguments=" --install-arguments=&quot;&apos;/COMPONENTS=icons,assoc,assoc_sh,autoupdate,windowsterminal,scalar&apos;&quot; --package-parameters=&quot;&apos;/GitAndUnixToolsOnPath /NoAutoCrlf /WindowsTerminal /NoShellIntegration /NoGuiHereIntegration /NoShellHereIntegration /WindowsTerminalProfile /DefaultBranchName:main /Editor:VisualStudioCode&apos;&quot;" />
```

This way, when I run `cinst -y choco-packages.config` it would apply the initial install-args and params for Git.

Proposed Solution: add `--include-install-args` flag to the `choco export` command.

TheCakeIsNaOH:
## Description Of Changes

This adds the --include-remembered-arguments option which is used to
export any remembered arguments. This reuses the
set_package_config_for_upgrade method in the NugetService to read and
parse the saved arguments.

Changes the set_package_config_for_upgrade method in the nuget service
to be public so it can be accessed for purposes of parsing the saved
arguments so they can be exported.

## Motivation and Context

This is rounding out the feature set of the export command, so it is able to export the saved arguments.

## Testing

TODO

## Change Types Made

## Related Issue

Fixes #2503
Depends on #2504 
Depends on #2187 

## Change Checklist

* [x] Documentation has been updated


TheCakeIsNaOH:
This is ready for a basic review to make sure I am on the right track, then I can look at adding unit tests and testing information.

TheCakeIsNaOH:
I'm open to suggestions on more unit tests that should be added.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -161,6 +180,49 @@ public void run(ChocolateyConfiguration configuration)
                                     xw.WriteAttributeString("version", packageResult.Package.Version.ToString());
                                 }
 
+                                if (configuration.ExportCommand.IncludeRememberedPackageArguments)
+                                {
+                                    var pkgInfo = _packageInfoService.get_package_information(packageResult.Package);
+                                    configuration.Features.UseRememberedArgumentsForUpgrades = true;
+                                    _nugetService.set_package_config_for_upgrade(configuration, pkgInfo);
+
+                                    //Mirrors the arguments captured in ChocolateyPackageService.capture_arguments()
```
gep13, line 189: Can we add a space here after the `//`?

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -161,6 +180,49 @@ public void run(ChocolateyConfiguration configuration)
...
+
+                                    if (configuration.AllowDowngrade) xw.WriteAttributeString("allowDowngrade", "true");
+                                    if (configuration.AllowMultipleVersions) xw.WriteAttributeString("allowMultipleVersions", "true");
+
+                                    if (!string.IsNullOrWhiteSpace(configuration.SourceCommand.Username)) xw.WriteAttributeString("user", configuration.SourceCommand.Username);
+                                    if (!string.IsNullOrWhiteSpace(configuration.SourceCommand.Password)) xw.WriteAttributeString("password", configuration.SourceCommand.Password);
+                                    if (!string.IsNullOrWhiteSpace(configuration.SourceCommand.Certificate)) xw.WriteAttributeString("cert", configuration.SourceCommand.Certificate);
+                                    if (!string.IsNullOrWhiteSpace(configuration.SourceCommand.CertificatePassword)) xw.WriteAttributeString("certPassword", configuration.SourceCommand.CertificatePassword);
+
+                                    //Arguments from the global options set
```
gep13, line 210: Can we add a space here after the `//`?

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -161,6 +180,49 @@ public void run(ChocolateyConfiguration configuration)
...
+                                    if (configuration.CommandExecutionTimeoutSeconds != ApplicationParameters.DefaultWaitForExitInSeconds)
+                                    {
+                                        xw.WriteAttributeString("timeout",configuration.CommandExecutionTimeoutSeconds.to_string());
+                                    }
+
+                                    //TODO, should this be exported? It seems rather system specific
```
gep13, line 216: Can we add a space here after the `//`?

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -161,6 +180,49 @@ public void run(ChocolateyConfiguration configuration)
...
+                                    //if (!string.IsNullOrWhiteSpace(configuration.CacheLocation)) xw.WriteAttributeString("cacheLocation", configuration.CacheLocation);
```
gep13, line 217: Can we add a space here after the `//`?

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -161,6 +180,49 @@ public void run(ChocolateyConfiguration configuration)
...
+
+                                    if (configuration.Features.FailOnStandardError) xw.WriteAttributeString("failOnStderr", "true");
+                                    if (!configuration.Features.UsePowerShellHost) xw.WriteAttributeString("useSystemPowershell", "true");
+
+                                    //Make sure to reset the configuration
```
gep13, line 222: Can we add a space here after the `//`?

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -133,13 +141,24 @@ public bool may_require_admin_access()
...
             var packageResults = _nugetService.get_all_installed_packages(configuration);
             var settings = new XmlWriterSettings { Indent = true, Encoding = new UTF8Encoding(false) };
+            var originalConfiguration = configuration.deep_copy();
+
+            if (configuration.ExportCommand.IncludeRememberedPackageArguments)
+            {
+                // The -o argument from the export command options set interferes with the -o argument from the install command options set.
+                ConfigurationOptions.OptionSet.Remove("o");
+
+                //Add the options set from the install command.
```
gep13, line 158: Can we add a space here after the `//`?

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -36,11 +36,14 @@ public class ChocolateyExportCommand : ICommand
     {
         private readonly INugetService _nugetService;
         private readonly IFileSystem _fileSystem;
+        private readonly IChocolateyPackageInformationService _packageInfoService;
+        private readonly IChocolateyPackageService _packageService;
```
gep13, line 40: This is defined here, and it is used further down on line 159, but it is never instantiated via the constructor. Is this an omission, or what is the thinking here?

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -133,13 +141,24 @@ public bool may_require_admin_access()
...
         public void run(ChocolateyConfiguration configuration)
         {
             var packageResults = _nugetService.get_all_installed_packages(configuration);
             var settings = new XmlWriterSettings { Indent = true, Encoding = new UTF8Encoding(false) };
+            var originalConfiguration = configuration.deep_copy();
+
+            if (configuration.ExportCommand.IncludeRememberedPackageArguments)
+            {
+                // The -o argument from the export command options set interferes with the -o argument from the install command options set.
+                ConfigurationOptions.OptionSet.Remove("o");
```
gep13, line 156: Doh! This is an interesting one! I distinctly remember going through the options that were used when setting `-o` for the export command, and I couldn't find any mention of it, but now that I check again, it is clearly there for the override command 😢

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -133,13 +141,24 @@ public bool may_require_admin_access()
...
+
+                //Add the options set from the install command.
+                var installCommand = new ChocolateyInstallCommand(_packageService);
```
gep13, line 159: `_packageService` isn't set in the constructor for this class, and it feels like it should be. I am surprised that this isn't causing an error when trying to run it.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
```diff
@@ -161,6 +180,49 @@ public void run(ChocolateyConfiguration configuration)
...
 
+                                if (configuration.ExportCommand.IncludeRememberedPackageArguments)
+                                {
+                                    var pkgInfo = _packageInfoService.get_package_information(packageResult.Package);
+                                    configuration.Features.UseRememberedArgumentsForUpgrades = true;
+                                    _nugetService.set_package_config_for_upgrade(configuration, pkgInfo);
+
+                                    //Mirrors the arguments captured in ChocolateyPackageService.capture_arguments()
+
+                                    if (configuration.Prerelease) xw.WriteAttributeString("prerelease", "true");
```
gep13, line 191: These strings, for example `prerelease` are starting to feel a little bit like "magic" strings, as they are being used in multiple places as bare strings. Being used here, as well as when naming the attributes in the `PackagesConfigFilePackageSetting` class.  Should we break these out into a new class, of const string's, or similar, so we can define them once, and use in multiple places?

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
gep13: Regarding the actual question on this one...

This is a good question! When actually using the remembered arguments during an upgrade procedure, using the same cachelocation used during the installation makes sense. However, if exporting the configuration and running  on a new machine, that could indeed be different.  I would be inclined to leave this commented out, and replace the TODO comment with another comment that says that we at least thought about this, and decided to not export it, and left it here commented out to ensure that it wasn't forgotten about in future.

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
```diff
@@ -572,6 +572,7 @@ public sealed class ProxyConfiguration
     public sealed class ExportCommandConfiguration
     {
         public bool IncludeVersionNumbers { get; set; }
+        public bool IncludeRememberedPackageArguments { get; set; }
```
gep13, line 575: Can we also get this added to the ChocolateyTabCompletion.ps1 script, so that folks can get tab completion for this new option. It would go somewhere in here: https://github.com/chocolatey/choco/blob/develop/src/chocolatey.resources/helpers/ChocolateyTabExpansion.ps1#L66

We would need a similar addition to the ChocolateyTabCompletion.ps1 file for the new `pinpackage` option on the install and upgrade command.  Can you also get that added?  No objection to that getting added to this PR, as long as it is in a separate commit, and tagged against #798

gep13 (changes requested):
Noticed a couple things on this one. Let me know if you have any questions about my comments.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
TheCakeIsNaOH: It is required when instantiating the `ChocolateyInstallCommand`, but when setting up to parse install arguments (via `ChocolateyInstallCommand.configure_argument_parser`), there is no usage/check that the passed in `IChocolateyPackageService` is actually instantiated. 

If you think it would be best, I could instantiate it?

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
TheCakeIsNaOH: Yup, at least there was an error when I tried to do it, so it's not like it became a silent bug.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
TheCakeIsNaOH: I'll go ahead and instanciate it.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
TheCakeIsNaOH: Done

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
TheCakeIsNaOH: Done

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
TheCakeIsNaOH: Done

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
TheCakeIsNaOH: Done

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
TheCakeIsNaOH: Done

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
TheCakeIsNaOH: Done

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
TheCakeIsNaOH: Done, both added.

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
TheCakeIsNaOH: Hmm, why not use the `PackagesConfigFilePackageSetting` (or `PackagesConfigFileSetting`) class to do the xml serialization, that way the strings would be in only one place?

TheCakeIsNaOH:
Converted to draft, as I found some inconsistent behavior with some arguments not being included, will do some debugging.

src/chocolatey.resources/helpers/ChocolateyTabExpansion.ps1
```diff
@@ -64,7 +64,7 @@ $commandOptions = @{
   sync = "--output-directory='' --id='' --package-id='' -?" + $allcommands
   optimize = "--deflate-nupkg-only --id='' -?" + $allcommands
   export = "--include-version-numbers --output-file-path='' -?" + $allcommands
-  template = "--name=''" + $allcommands
+  template = "--name='' --include-remembered-arguments" + $allcommands
```
gep13, line 67: Shouldn't this be attached to the export command, and not the template command?

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
gep13: Hmm, that is not a bad idea. That would mean that the serialized XML could be used elsewhere. Although, I don't immediately see a need for it anywhere else, but putting it in one of these classes would mean that we "simplify" the command class, and do the work elsewhere.

src/chocolatey.resources/helpers/ChocolateyTabExpansion.ps1
TheCakeIsNaOH: Fixed

src/chocolatey/infrastructure.app/commands/ChocolateyExportCommand.cs
TheCakeIsNaOH: I've switched to serializing that class to create the xml. It required adding members so the boolean options are not serialized if they are not specified, but otherwise I like the implementation.

TheCakeIsNaOH:
I've fixed the inconsistent behavior I think, and updated the manual testing. This is ready for a re-review.

TheCakeIsNaOH:
This is draft until #3003 is merged/closed, as the implementation of this depends on what happens with that PR.

gep13:
Need to discuss what would happen here with regard to sensitive arguments that might have been passed in with Chocolatey Licensed Extension, i.e. --package-parameters-sensitive.  We can't allow those to be included in the exported file.

nopeless:
@TheCakeIsNaOH thanks for the work. Makes it easier to export my current package config

Issue #2501, opened 2021-12-28
Complete pending Pin Command specs


TheCakeIsNaOH:
There are a couple of tests in `ChocolateyPinCommandSpecs` that are marked as pending because the nuget mocking needs to be completed.
https://github.com/chocolatey/choco/blob/d8ca7bad2ad64538a0c1026a9317700dbb7f03db/src/chocolatey.tests/infrastructure.app/commands/ChocolateyPinCommandSpecs.cs#L427-L457

These nuget mocking methods should be completed, and the tests enabled.

gep13:
@AdmiringWorm did you have any thoughts on these pending tests? Anything that you can think about adding here that would help?

TheCakeIsNaOH:
IMHO, this would be best served by waiting for the nuget uplift, and completing these tests after that is merged, because work spent completing these tests now would likely be mostly not be applicable for newer nuget library versions.

PR #2385, opened 2021-10-01
(#449) Add template command

TheCakeIsNaOH:
## Description Of Changes

Renames noop to generate_noop in the template service.

Adds `template` command, and associated functions in the template service.

## Motivation and Context

This allows a user to get information about all installed templates, or
a specific template. The list subcommand lists all installed templates.
The info subcommand lists verbose information about a specific template

Usage with -r should output a machine parsable list of the installed
templates. Usage with verbose (implied with the info subcommand) 
outputs more information about the template(s).

The information displayed here can be gathered by running the list
command, config command and inspecting the contents of the
TemplatesLocation. However, this command allows the information to be
displayed in one location.

Both manually installed templates and templates installed via .template
packages are displayed. 

## Testing

There have not been unit tests added for the template service `list` method. This is because the correct methods to mock out the nuget information need to be found, see #2501.

Steps to setup templates to test with (run in CMD):
```
REM Clean out anything old
DEL /S /Q "C:\path\to\choco\src\chocolatey.console\bin\debug\lib"
DEL /S /Q "C:\path\to\choco\src\chocolatey.console\bin\debug\.chocolatey"
DEL /S /Q "C:\path\to\choco\src\chocolatey.console\bin\debug\templates"
REM Setup config
C:\path\to\choco\src\chocolatey.console\bin\debug\choco.exe feature disable -n showNonElevatedWarnings --allow-unofficial -y
C:\path\to\choco\src\chocolatey.console\bin\debug\choco.exe feature enable -n allowGlobalConfirmation --allow-unofficial -y
C:\path\to\choco\src\chocolatey.console\bin\debug\choco.exe feature enable -n useRememberedArgumentsForUpgrades
REM install msi and zip templates
C:\path\to\choco\src\chocolatey.console\bin\debug\choco.exe install msi.template
C:\path\to\choco\src\chocolatey.console\bin\debug\choco.exe install zip.template
REM copy msi template to emulate templates installed manually (e.g. outside of package)
xcopy "C:\path\to\choco\src\chocolatey.console\bin\debug\templates\msi" "C:\path\to\choco\src\chocolatey.console\bin\debug\templates\non-pkg\" /s /e
xcopy "C:\path\to\choco\src\chocolatey.console\bin\debug\templates\msi" "C:\path\to\choco\src\chocolatey.console\bin\debug\templates\default\" /s /e
```

Then run various commands to validate it lists out the correct information:
```
REM check that the command selects list as the default option
[...]

PR #2507, opened 2022-01-05
(#2500) List parameters in templates

### LINKED ISSUE #2500: List parameters in templates
Follow on to this issue: https://github.com/chocolatey/choco/issues/449

Parameters in templates should be listed. This entails code to find the parameters, and code to display the parameters in the `list` command. Assuming that the parameters are cached in a file, a format would need to be chosen, and code to write/read these parameters would be needed.

TheCakeIsNaOH:
## Description Of Changes

This adds the ability for the template command to list the parameters
in templates. If the template is installed via a package, the
parameters get cached in a file called .parameters in the template's
directory. This file is ignored by calls to choco new, as it is
excluded from being copied.

There another method added to TokenReplacer to get the parameters via
regex. This keeps the regex in the TokenReplacer class.

## Motivation and Context

It it useful to be able to list the available parameters in a template. 

The parameters are cached to reduce the number of files that need to be read to get the parameters. Only templates installed via a `.template` package have the parameters cached. This is because those cache files would get deleted as a part of the package upgrade process, so any cache files from those templates would be up to date. However, for packages that are installed manually, there is not an efficient way to check that the cache is still valid, so that it why parameters from those templates are always read manually.

## Testing

Just like in #2385, unit tests cannot be added yet, due to #2501

Manual testing instructions:
1. Install `msi.template` and `zip.template`
2. Make a copy of the msi template folder and rename to "default"
3. Make a copy of the msi template folder and rename to "non-pkg"
4. Run `choco template --verbose`
5. Check that the parameters are listed correctly
6. Check that only the msi and zip template folders have the `.parameters` file

## Change Types Made

## Related Issue

Fixes #2500
Depends on #2385 (now merged)

## Change Checklist

### DIFF
--- a/src/chocolatey.tests/infrastructure.app/services/TemplateServiceSpecs.cs
+++ b/src/chocolatey.tests/infrastructure.app/services/TemplateServiceSpecs.cs
@@ -26,6 +26,7 @@ namespace chocolatey.tests.infrastructure.app.services
     using chocolatey.infrastructure.app.services;
     using chocolatey.infrastructure.app.templates;
     using chocolatey.infrastructure.filesystem;
+    using chocolatey.infrastructure.services;
     using Moq;
     using NUnit.Framework;
     using Should;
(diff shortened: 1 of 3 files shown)


gep13 (approved):
LGTM!

gep13:
@TheCakeIsNaOH thanks for pulling this together! I am going to go ahead and get this merged in once the CI builds finish.

Issue #58, opened 2015-02-04
All powershell functions (helpers) rewritten into chocolatey.powershell.dll


ferventcoder:

pauby:
Duplicate of #3477 

Keeping the later issue as it has much more information than this one.

PR #2514, opened 2022-01-06
(#2513) Add support for filtering install locations via softwareName

### LINKED ISSUE #2513: Add support for filtering install locations via softwareName
Currently, `Install-ChocolateyPackage` and `Install-ChocolateyInstallPackage` have the `softwareName` parameter in their examples, but the parameter is not documented (chocolatey/docs#308) and the parameter is not implemented (it just goes to `$ignoredArguments`). The parameter is also in the `chocolateyInstall.ps1` as created by the new command.

I don't think there is an issue yet for implementing this parameter yet. 

It needs to be added to both powershell functions as a parameter, and documented in those functions. Then, it needs to get passed back into the C#, presumably in the same way `ChocolateyPackageInstallLocation` is used. Then finally it can be used to filter the registry snapshot.

TheCakeIsNaOH:
## Description Of Changes

Implement softwareName filtering of registry keys

TODO list:
* [ ] check PS v2 compatibility
* [ ] make sure this actually works
* [ ] check if ChocolateyInstalledSoftwareName is an ok name for the env variable
* [ ] check if behavior (and warnings) for filtering is ok
* [ ] breaking change discussion (not *technically* a breaking change for packages with a correct `softwareName`, some maintainers still may be affected.)
* [ ] document manual tests
* [ ] automated tests?
* [ ] fix commit messages

## Motivation and Context

TODO

## Testing

TODO

## Change Types Made

* [x] Breaking change (fix or feature that could cause existing functionality to change)

## Related Issue

Fixes #2513

## Change Checklist

* [ ] Documentation has been updated (issue open chocolatey/docs#308)

### DIFF
--- a/src/chocolatey.resources/helpers/functions/Install-ChocolateyInstallPackage.ps1
+++ b/src/chocolatey.resources/helpers/functions/Install-ChocolateyInstallPackage.ps1
@@ -121,6 +121,14 @@ only use the ones available with the package. Available in 0.9.10+.
 .PARAMETER IgnoredArguments
 Allows splatting with arguments that do not apply. Do not use directly.
 
+.PARAMETER SoftwareName
+Specifies the display name of the installed software as shown in 
+Programs & Features. Can be an exact string or contain wildcards ("*").
+Available in 0.12.0+
+
+Use this to filter any software installed during the package installation
+script to ensure the correct software install location is found.
+
 .EXAMPLE
 >
 $packageName= 'bob'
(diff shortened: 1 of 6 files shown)


TheCakeIsNaOH:
Currently, the `softwareName` is brought back into the C# codebase as an environment variable string. So if a package installs something that has multiple entries in the registry (e.x. 32 and 64 bit vcredist, or autodesk (grrr)), there is no current way to get the names of both entries back. 

Given that the software name could be any string, there is not really an awesome way of using a delimiter to return multiple strings. 

One possibility for this is [to get the powershell helpers rewritten into a `.dll`](https://github.com/chocolatey/choco/issues/58), which then may be able to return other object types, so an array type thing could be used to return multiple software names.

Given the amount of work required before that could happen, I'm going to close this PR for the moment. I might revisit this PR/issue in the future, but at the moment, there are better things to spend time on.

PR #2517, opened 2022-01-10
(#2481) Warn about unsupported packages on non-Windows

### LINKED ISSUE #2481: Warn about unuspported packages on non-Windows platforms

TheCakeIsNaOH:
## Description Of Changes

This adds a check for non-Windows systems that warns if the package id
does not end in .template on install. 

## Motivation and Context

This check is added to help
provide guidance to users, as templates are the only package type
currently supported on non-Windows systems.

In the future, this check may be expanded to allow other packages
types, as needed.

## Testing

1. Build choco on linux
2. Run `mono choco.exe install curl --allow-unofficial` and confirm that the warning appears
3. Run `mono choco.exe install msi.tempalte --allow-unofficial` and confirm that the warning does not appear

## Change Types Made

## Related Issue

Fixes #2481

## Change Checklist

* [ ] Documentation has been updated chocolatey/docs#280

### DIFF
--- a/src/chocolatey/infrastructure.app/services/NugetService.cs
+++ b/src/chocolatey/infrastructure.app/services/NugetService.cs
@@ -447,6 +447,12 @@ folder.");
                 //todo: get smarter about realizing multiple versions have been installed before and allowing that
                 IPackage installedPackage = packageManager.LocalRepository.FindPackage(packageName);
 
+                if (Platform.get_platform() != PlatformType.Windows && !packageName.EndsWith(".template"))
+                {
+                    string logMessage = "{0} is not a supported package on non-Windows systems.{1}Only template packages are currently supported.".format_with(packageName, Environment.NewLine);
+                    this.Log().Warn(ChocolateyLoggers.Important, logMessage);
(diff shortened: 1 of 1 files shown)


gep13 (approved):
LGTM!

gep13:
@TheCakeIsNaOH thanks again for getting this added!

Issue #2540, opened 2022-01-19
Python alternate source does not handle all keyword


TheCakeIsNaOH:
### What You Are Seeing?

When running `choco upgrade all` with the `python` alternate source, it tries to install a package named `all` from pypi.

### What is Expected?

That the python alternate source either supports the all keyword for upgrade to upgrade all installed python packages, or throws a not implemented exception.

### How Did You Get This To Happen? (Steps to Reproduce)

`choco upgrade all -s python`

### Output Log

<details>
<summary>Full Log Output</summary>

<p>

~~~sh
Chocolatey v0.11.2
Chocolatey is running on Windows v 6.3.9600.0
Attempting to delete file "C:/ProgramData/chocolatey/choco.exe.old".
Attempting to delete file "C:\ProgramData\chocolatey\choco.exe.old".
Command line: "C:\ProgramData\chocolatey\choco.exe" upgrade all -s python --verbose --debug
Received arguments: upgrade all -s python --verbose --debug
RemovePendingPackagesTask is now ready and waiting for PreRunMessage.
Sending message 'PreRunMessage' out if there are subscribers...
[Pending] Removing all pending packages that should not be considered installed...
Performing validation checks.
Global Configuration Validation Checks:
 - Package Exit Code / Exit On Reboot = Checked
System State Validation Checks:
 Reboot Requirement Checks:
 - Pending Computer Rename = Checked
 - Pending Component Based Servicing = Checked
 - Pending Windows Auto Update = Checked
 - Pending File Rename Operations = Checked
 - Pending Windows Package Installer = Checked
 - Pending Windows Package Installer SysWow64 = Checked
The source 'python' evaluated to a 'python' source type

NOTE: Hiding sensitive configuration data! Please double and triple
 check to be sure no sensitive data is shown, especially if copying
 output to a gist for review.
Configuration: CommandName='upgrade'|
CacheLocation='C:\Users\Administrator\AppData\Local\Temp\1\chocolatey'|
ContainsLegacyPackageInstalls='True'|
CommandExecutionTimeoutSeconds='2700'|WebRequestTimeoutSeconds='30'|
Sources='python'|SourceType='python'|Debug='True'|Verbose='True'|
Trace='False'|Force='False'|Noop='False'|HelpRequested='False'|
UnsuccessfulParsing='False'|RegularOutput='True'|QuietOutput='False'|
PromptForConfirmation='False'|AcceptLicense='False'|
[...]

PR #2541, opened 2022-01-20
(#23) Handle all keyword for install

### LINKED ISSUE #23: Handle 'all' keyword for install

TheCakeIsNaOH:
## Description Of Changes

This adds the ability to install all packages from a source via the all
keyword. It runs a list against the current sources, and sets the
package names from that list of packages. 

This functionality is only intended to be used with internal
repositories. Therefore a list of public repositories has been added,
and the current sources are checked against it to ensure only internal
sources are specified.

Renames the `set_package_names_if_all_is_specified function` to
`set_local_package_names_if_all_is_specified`. This is to increase
clarity with the `set_remote_package_names_if_all_is_specified` now being added.

Also this throws a not implemented exception if the all keyword is used with
non-nuget alternate sources. This is because the all keyword only has
handling built in for normal sources, and it is better to throw an
error than to have the source try to install a package called "all"

## Motivation and Context

Adding the all keyword for install will bring C# choco up to feature parity with PowerShell choco. Or, at least all of the feature parity issues will be closed.

## Testing

Run these to validate that choco will throw an error because public repositories are one of the sources:
```
choco install all --allow-unofficial
choco install all --allow-unofficial --source=https://community.chocolatey.org/api/v2/
choco install all --allow-unofficial --source=https://chocolatey.org/api/v2/
choco install all --allow-unofficial --source=https://www.nuget.org/api/v2/
choco install all --allow-unofficial --source=https://licensedpackages.chocolatey.org/api/v2/
choco install all --allow-unofficial --source=https://push.chocolatey.org/api/v2
choco install all --allow-unofficial --source="'https://community.chocolatey.org/api/v2/;"C:\packages"'"
```

Run these to validate that alternate sources fail with a notimplemented exception:
```
choco install all --allow-unofficial --source=WindowsFeatures
choco install all --allow-unofficial --source=ruby
choco install all --allow-unofficial --source=python
choco install all --allow-unofficial --source=webpi
choco install all --allow-unofficial --source=cygwin
```

Download these packages to a local folder:
- `curl` `7.81.0`
- `wget` `1.21.2`
- `iperf2` `2.0.13`
- `iperf2` `2.0.14`
- `iperf2` `2.0.14.1001-alpha20201022`

Run these commands pointing to that local folder to test the behavior of the all keyword:

Ensure that iperf 2.0.14 (i.e. not the prerelease version), wget and curl are installed:
`choco install all --allow-unofficial --source=c:\packages` 

Ensure that the prerelease iperf is installed when prerelease is specified:
`choco install all --allow-unofficial --source=c:\packages --pre` 
 
## Change Types Made

## Related Issue

Fixes #23

## Change Checklist

* [x] Documentation has been updated

### DIFF
--- a/src/chocolatey/infrastructure.app/services/CygwinService.cs
+++ b/src/chocolatey/infrastructure.app/services/CygwinService.cs
@@ -218,6 +218,11 @@ namespace chocolatey.infrastructure.app.services
 
         public ConcurrentDictionary<string, PackageResult> Install(ChocolateyConfiguration config, Action<PackageResult, ChocolateyConfiguration> continueAction, Action<PackageResult, ChocolateyConfiguration> beforeModifyAction)
         {
+            if (config.PackageNames.IsEqualTo(ApplicationParameters.AllPackages))
+            {
+                throw new NotImplementedException("Alternative sources do not allow the use of the 'all' package name/keyword.");
+            }
+
             var args = BuildArgs(config, _installArguments);
             var packageResults = new ConcurrentDictionary<string, PackageResult>(StringComparer.InvariantCultureIgnoreCase);
 
--- a/src/chocolatey.tests.integration/scenarios/InstallScenarios.cs
+++ b/src/chocolatey.tests.integration/scenarios/InstallScenarios.cs
@@ -22,6 +22,7 @@ using System.IO;
 using System.Linq;
 using System.Text;
 using System.Xml.XPath;
+using chocolatey.infrastructure.app;
 using chocolatey.infrastructure.app.commands;
 using chocolatey.infrastructure.app.configuration;
 using chocolatey.infrastructure.app.services;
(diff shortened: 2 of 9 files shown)


TheCakeIsNaOH:
For reference, here is the previous (PowerShell) implementation of handling the all keyword:
https://github.com/chocolatey-archive/chocolatey/blob/master/src/functions/Chocolatey-InstallAll.ps1

TheCakeIsNaOH:
TODOs:
- [ ] Commits reword
- [ ] Fill in PR template
- [ ] More local testing
- [ ] Disable or implement "all" keyword for alternate sources (like #2540 )
- [ ] Check if error message is ok, if more info should be given
- [ ] Check if other specific urls need to be disabled (e.g. other major public nuget repositories, alternate URLs for chocolatey.org/nuget.org?)
- [ ] Documentation in install command help

gep13:
Some quick thoughts on this one...

We would want to limit usage of this to only custom sources, i.e. don't allow this on CCR, or Licensed Feed.

And, add a new property to source element, similar to `selfService` and `adminOnly`.  So we would have `installAll` as an element.

TheCakeIsNaOH:
> We would want to limit usage of this to only custom sources, i.e. don't allow this on CCR, or Licensed Feed.

This should already be the case, see the `PublicNugetSources` in `ApplicationParameters.cs`

> And, add a new property to source element, similar to selfService and adminOnly. So we would have installAll as an element.

Would that be a breaking change for the `choco source` output?

src/chocolatey/infrastructure.app/services/CygwinService.cs
```diff
@@ -212,6 +212,11 @@ public void InstallDryRun(ChocolateyConfiguration config, Action<PackageResult,
 
         public ConcurrentDictionary<string, PackageResult> Install(ChocolateyConfiguration config, Action<PackageResult, ChocolateyConfiguration> continueAction, Action<PackageResult, ChocolateyConfiguration> beforeModifyAction)
         {
+            if (config.PackageNames.is_equal_to(ApplicationParameters.AllPackages))
+            {
+                throw new NotImplementedException("All keyword is not available for alternate sources");
```
gep13, line 217:
```suggestion
                throw new NotImplementedException("Alternative sources do not allow the use of the 'all' package name/keyword.");

```

gep13: This change has been applied.

gep13 (approved):
LGTM!

gep13:
@TheCakeIsNaOH thanks again for getting this functionality added in!

PR #2550, opened 2022-01-22
(#2468) Remove deprecated Chocolatey commands

### LINKED ISSUE #2468: Remove deprecated Chocolatey commands and shims
The following commands have been marked as deprecated and should be removed in v1.0.0:

* `update` (see [docs](https://docs.chocolatey.org/en-us/choco/commands/update))
* `version` (see [docs](https://docs.chocolatey.org/en-us/choco/commands/version))

As well as removing the Chocolatey commands, i.e. `choco update` and `choco version`, we also need to make sure to remove the shim(s) that are installed.

There is a shim `cver` that executes `choco version` and needs to be removed.

An update to the chocolateyInstall.ps1 will be required to ensure that this is removed after successful installation of 1.0.0.

**NOTE:** The `cup` is not being removed.  This correctly points at `choco upgrade` and will be removed in 2.0.0.

gep13:
The following Chocolatey commands have been marked as deprecated for a while now, and with the upcoming release of 1.0.0, it is time to remove them:

* choco version
* choco update

The functionality of these commands is now provided elsewhere, and these commands can be removed.

In addition, the shim for `cver` has also been removed.

Fixes #2468 

https://app.clickup.com/t/20540031/ENGTASKS-723


corbob:
Should there be an upgrade path that removed the `cver.exe` shim?

gep13:
@corbob also not sure what you mean here, can you elaborate?

corbob:
So if I take the nupkg to a fresh system and install it, `cver.exe` is not there. But if I already have Chocolatey on the box and do `choco upgrade chocolatey` `cver.exe` is still on the path.

corbob (approved):
Aside from the previous comment about possibly removing `cver.exe` on upgrade, it all looks good to me. My testing shows it all works as expected both from an upgrade and from a fresh install perspective.

The removal can be discussion and a separate PR if we decide that's something for us to do.

pauby:
That's a good point @corbob.

@gep13 it looks like the shims are not created when Chocolatey is installed for the first time but doesn't remove it on upgrade? Would there have to be changes to the install script?

gep13:
>@pauby said...
it looks like the shims are not created when Chocolatey is installed for the first time but doesn't remove it on upgrade?

Just to make sure that are on the same page, I am reading this as:

> it looks like the shims are created when Chocolatey is installed for the first time but doesn't remove it on upgrade?

If so, then yes, we need to add some steps to the chocolateyInstall.ps1 file to ensure that the `cver` shim file is removed from the bin folder when upgrading.

To illustrate what happens...

With version 0.12.1 of Chocolatey installed, attempting to run `cver` results in the following:

image

When attempting to run it with the code from this PR, you get the following:

image

This is "bad" since the cver shim is still there, which is happening to run `choco version` and that command is no longer registered.

When we remove the cver shim, we then see the following:

image

Which is what we actually want to see.

I will attempt to update this PR at some point to include the work to ensure that the cver shim is removed (unless someone beats me to this).

gep13:
I am converting this to a draft PR until the removal of the shim is completed.

AdmiringWorm (commented):
This overall looks good to me.

The only thing remaining is to remove the shim during upgrade.
The code for this is included in the PR #2551 and the only change needed for this PR would be to add the `cver` executable to the list of shims to remove.

AdmiringWorm (approved):
LGTM 👍

AdmiringWorm:
@gep13 Thanks for getting this added 👍

PR #2551, opened 2022-01-22
(#89) Remove cpack shim for choco pack

### LINKED ISSUE #89: Remove the `cpack` command alias

gep13:
There are known incompatibilities with this shim when other applications
are installed on the machine, for example cmake. Since we are moving
towards usage of a ubiquitous choco command, removing this shim makes
a lot of sense.

Fixes #89


corbob:
Should there be an upgrade path that removed the `cpack.exe` shim?

gep13:
@corbob I am not sure if I follow what you mean here. Can you elaborate?

corbob:
So if I take the nupkg to a fresh system and install it, cpack.exe is not there. But if I already have Chocolatey on the box and do `choco upgrade chocolatey` `cpack.exe` is still on the path.

TheCakeIsNaOH:
> `cpack.exe` is still on the path.

What if a user replaced the `cpack.exe` that points to `choco pack` and instead has a shim that points to the cmake `cpack`? 

Or what if a user manually created a `cpack.exe` that points to `choco pack`, since they wanted the functionality back? If there was code that removed the shim, then that manually created shim might be removed on upgrade.

Although removing the shim still might be the best course of action, there are lots of edge cases.

gep13:
>@TheCakeIsNaOH said...
Although removing the shim still might be the best course of action, there are lots of edge cases.

Agreed.  There is a long standing "rule" that Chocolatey package installations shouldn't directly we adding anything into the Chocolatey bin folder, and the same rule should apply to folks directly adding files in there.  Obviously we can't stop someone doing this if they really want to, but they can't expect that Chocolatey (in future versions) won't do something that results in something that they have done manually stops working.

We have discussed this within the team, and the plan going forward would be to check whether the `cpack` shim exists, if it does, if it is signed with either the Chocolatey Software or Real Dimensions Authenticode signature, we will delete it, as we know that we placed it there, otherwise, we will leave it alone.

gep13:
Converting to to draft until an addition to the chocolateyInstall.ps1 file is made to affect the removal of the `cpack` shim from the bin folder on upgrade to 1.0.0.

AdmiringWorm (approved):
LGTM.

As I added the changes in the PowerShell file I'll need a second pair of eyes to ensure it both works as expected and looks fine.

nuget/chocolatey/tools/chocolateysetup.psm1
```diff
@@ -30,13 +30,54 @@ param (
...
+        if (Test-Path "$filePath.old") {
+          $null = Remove-Item "$filePath.old"
+        }
+      }
+    }
+
+    if ($signature.SignerCertificate.Thumbprint -ne '83AC7D88C66CB8680BCE802E0F0F5C179722764B') {
+      Write-ChocolateyWarning "Shim found in $filePath, but did not match our signature. Ignoring removal..."
+      return
+    }
```
corbob, line 71: Is this part needed if we're removing the file on line 56?

corbob (commented):
Just the one comment from my perspective.

nuget/chocolatey/tools/chocolateysetup.psm1
AdmiringWorm: Good catch, seems I didn't remove this one after refactoring the method to support multiple Subjects.
I'll get that part removed.

nuget/chocolatey/tools/chocolateysetup.psm1
```diff
@@ -30,13 +30,56 @@ param (
   }
 }
 
+function Remove-ShimWithAuthenticodeSignature {
+  param (
+    [string] $filePath
+  )
+  if (!(Test-Path $filePath)) {
+    return
+  }
```
gep13, line 39: @AdmiringWorm the alignment doesn't look quite right here.  Can you push this over to line up with if statement below?

gep13 (commented):
Small change...

nuget/chocolatey/tools/chocolateysetup.psm1
AdmiringWorm: This indentation was actually correct, however, the remaining of the function had the wrong indentation compared to the rest of the file.

I have fixed that up.

gep13:
@AdmiringWorm thanks for getting this over the line!

AdmiringWorm:
@gep13 it was a pretty big line, but finally got there eventually 😆

PR #2553, opened 2022-01-27
(maint) Fixes TeamCity Signing in UppercuT

JPRuskin:
There was an issue with using `signtool.exe` during the TeamCity build

This PR fixes the UppercuT build to work with signing in our environment by:

- Adding a property to be able to pass in the certificate subject name
- Adding a property to track if the build is running in TeamCity or not
- Adding a third potential path to signtool.exe, in the Windows SDK 7.1 folder
- Adding logic to use a certificate stored in the local machine store if running in TeamCity

It also adds better handling for failure during the build.

### DIFF
--- a/.build.custom/codeSign.step
+++ b/.build.custom/codeSign.step
@@ -26,7 +26,7 @@
   </target>
 
   <target name="get_password" if="${property::exists('path.code.cert.password')}">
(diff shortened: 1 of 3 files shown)


gep13 (approved):
LGTM!

gep13:
@JPRuskin let's do this thing!

PR #2559, opened 2022-02-04
(build) Updates TeamCity BuildNumber

JPRuskin:
A very minor update to the UppercuT steps to use the artifact version as the build number, which allows us to have a far more user-friendly experience when finding linked builds / artifacts in our systems.

Please note that though this shows as `level="Warning"` it doesn't appear in the build logs - but needs to be at that level, otherwise TeamCity won't consume it.

This has been shown to work on the dev and local build server(s), and in the Licensed builds.


gep13 (approved):
LGTM!

PR #2595, opened 2022-02-11
(#2545) Adds Clarity to List Command Help

### LINKED ISSUE #2545: Source switch doesn't do anything on choco list --lo
### What You Are Seeing?

**choco list --lo** outputs all installed packages.  Using **--source** switch to specific the repo doesn't affect **choco list --lo**
This means there's no way to get a list of the packages installed on a host from a specific repo the way you can for choco outdated

### What is Expected?
The number of packages returned from **choco list --lo** is the same as **choco list --lo --source** which means source is ignored for **--lo**  queries
Using **choco list --lo --source** for a source that zero packages have been installed from, still returns a full **choco list --lo** output

This means it's only possible to see which packages are installed from a specific source with **choco outdated**, and not **choco list --lo**

**choco list --lo --source reponame**
<p>
</details>

JPRuskin:
## Description of Changes

This PR adds some clarity to the `choco list` command help.

## Motivation and Context

There was confusion around `choco list --local-only` and if it supported `--source` (see issue #2545). This addition to the help line for `--local-only` makes it clearer that searching your local machine won't filter further by source. Given that this is the line after the help output for `--source` it seemed like it didn't need to be clarified there.

## Related Issue

Fixes #2545

## Notes

I worry that it's a little terse, but any further words seemed to add potential confusion. Happy to modify it as folk prefer.

Please note that this won't show up in docs until the docs are regenerated and pushed.

### DIFF
--- a/src/chocolatey/infrastructure.app/commands/ChocolateyListCommand.cs
+++ b/src/chocolatey/infrastructure.app/commands/ChocolateyListCommand.cs
@@ -46,7 +46,8 @@ namespace chocolatey.infrastructure.app.commands
                      "Source - Source location for install. Can use special 'webpi' or 'windowsfeatures' sources. Defaults to sources.",
                      option => configuration.Sources = option.remove_surrounding_quotes())
                 .Add("l|lo|local|localonly|local-only",
-                     "LocalOnly - Only search against local machine items.",
+                     "LocalOnly - Only search against local machine items. Ignores --source if provided.",
+
                      option => configuration.ListCommand.LocalOnly = option != null)
                 .Add("idonly|id-only",
                      "Id Only - Only return Package Ids in the list results. Available in 0.10.6+.",


src/chocolatey/infrastructure.app/commands/ChocolateyListCommand.cs
```diff
@@ -46,7 +46,7 @@ public virtual void configure_argument_parser(OptionSet optionSet, ChocolateyCon
                      "Source - Source location for install. Can use special 'webpi' or 'windowsfeatures' sources. Defaults to sources.",
```
pauby, line 46: Maybe add something here to say when used with `--local-only` it is ignored?

src/chocolatey/infrastructure.app/commands/ChocolateyListCommand.cs
```diff
@@ -46,7 +46,7 @@ public virtual void configure_argument_parser(OptionSet optionSet, ChocolateyCon
                      "Source - Source location for install. Can use special 'webpi' or 'windowsfeatures' sources. Defaults to sources.",
                      option => configuration.Sources = option.remove_surrounding_quotes())
                 .Add("l|lo|local|localonly|local-only",
-                     "LocalOnly - Only search against local machine items.",
+                     "LocalOnly - Only search against local machine items. Ignores --source.",
```
pauby, line 49: Same here. Maybe flesh this out to say when used with `--source` then it's ignored. I think what you have is very short. Not looking for ware and peace, just somewhere in the middle.

pauby (changes requested):

src/chocolatey/infrastructure.app/commands/ChocolateyListCommand.cs
JPRuskin: For what it's worth, as I said in the context above (perhaps it should have been in notes, instead), there seems to be no way to output that line without the modified line directly following it - e.g.
image

Do we need to say it twice in two lines if there's no way to display one without the other?

src/chocolatey/infrastructure.app/commands/ChocolateyListCommand.cs
JPRuskin: Updated to `Ignores --source if provided, as local packages do store installation source.`

src/chocolatey/infrastructure.app/commands/ChocolateyListCommand.cs
```diff
@@ -46,7 +46,7 @@ public virtual void configure_argument_parser(OptionSet optionSet, ChocolateyCon
                      "Source - Source location for install. Can use special 'webpi' or 'windowsfeatures' sources. Defaults to sources.",
                      option => configuration.Sources = option.remove_surrounding_quotes())
                 .Add("l|lo|local|localonly|local-only",
-                     "LocalOnly - Only search against local machine items.",
+                     "LocalOnly - Only search against local machine items. Ignores --source if provided, as local packages do not store installation source.",
```
pauby, line 49:
```suggestion
                     "LocalOnly - Only search against local machine items. Ignores --source if provided.",

```

pauby: Given what you said above let's change this and we're done I think?

pauby: @gep13 Any comments on this?

pauby (changes requested):

JPRuskin:
@gep13 Any comments on this, now things are a little less busy?

src/chocolatey/infrastructure.app/commands/ChocolateyListCommand.cs
gep13: Yes, I think the shorter version is better here.  There is additional work that @AdmiringWorm is doing in this area to mark certain things as deprecated in this area, as they will be removed in future versions, where `choco list` only works for locally installed packages.

gep13 (approved):
LGTM!

gep13:
@JPRuskin thanks for getting this updated!

PR #2620, opened 2022-03-07
(#2538) Store a list of supported features

### LINKED ISSUE #2538: Correctly handle deprecated features
With the release of v0.12.0 we deprecated a feature.  All the code for this feature was removed from the code case, however, after upgrading an installation of Chocolatey, the corresponding entry can still be found in the chocolatey.config file.

This by itself doesn't actually cause a problem, and it needs to be left in the file in order to allow for backwards compatibility, however, when executing the following command on Chocolatey v0.12.0:

```
choco feature enable --name=scriptsCheckLastExitCode
```

It will output that the feature was enabled, when actually, nothing in the current codebase will do anything with this.

In any given Chocolatey version, we need to support only the features that exist in that release, nothing else, even though they exist in the chocolatey.config file.  This is different to the config values, which are intended to be extended through Chocolatey Extensions, etc. However, features are specific to Chocolatey.

This will require an upstream modification for when the Chocolatey Licensed Extension is in play.

AdmiringWorm:
## Description Of Changes

This pull request updates the config settings service to
store a list of supported features. This is done so we
know which features we should handle without reporting
to the user that the feature is not supported, and will have
no effect when running Chocolatey.

## Motivation and Context

To handle features that we support and don't support, and to notify the user if it isn't supported.

## Testing

1. Run build.bat to ensure that the build and unit tests has not been broken
2. run `choco feature disable -n checksumFiles` and make sure there is no mention it is unsupported
3. Run `choco feature enable -n checksumFiles` and make sure there is no mention it is unsupported
4. Ensure that the feature `scriptsCheckLastExitCode` is set in the configuration file, with a disabled value
5. Run `choco feature enable -n scriptsCheckLastExitCode` and make sure the user is notified that the feature is not supported
6. Run `choco feature disable -n scriptsCheckLastExitCode` and make sure the user is notified that the feature is not supported 

## Change Types Made

## Related Issue

Fixes #2538

## Change Checklist

* [x] PowerShell v2 compatibility checked.

### DIFF
--- a/src/chocolatey.tests/chocolatey.tests.csproj
+++ b/src/chocolatey.tests/chocolatey.tests.csproj
@@ -109,6 +109,7 @@
     <Compile Include="infrastructure.app\configuration\ConfigurationOptionsSpec.cs" />
     <Compile Include="infrastructure.app\nuget\NugetCommonSpecs.cs" />
     <Compile Include="infrastructure.app\services\AutomaticUninstallerServiceSpecs.cs" />
+    <Compile Include="infrastructure.app\services\ChocolateyConfigSettingsServiceSpecs.cs" />
     <Compile Include="infrastructure.app\services\ChocolateyPackageServiceSpecs.cs" />
     <Compile Include="infrastructure.app\services\FilesServiceSpecs.cs" />
     <Compile Include="infrastructure.app\services\NugetServiceSpecs.cs" />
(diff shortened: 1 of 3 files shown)


AdmiringWorm:
@gep13 after messaging Paul about this, I have changed the warning reporting to throw an exception instead.

gep13:
@AdmiringWorm thanks for getting this added!

PR #2632, opened 2022-03-08
(#2630) Remove properties from ConfigFileSettings

### LINKED ISSUE #2630: Remove deprecated configurations from Config
The root elements `cacheLocation` and `commandExecutionTimoutSeconds` are deprecated as these are now handled in a different way instead together with other configuration values.
These two are scheduled to be removed in v1 of Chocolatey.

https://github.com/chocolatey/choco/blob/develop/src/chocolatey/infrastructure.app/configuration/ConfigFileSettings.cs#L30-L36

gep13:
## Description Of Changes

Remove usage of Obsolete properties.

There were some usages of these properties, in the Equals and
GetHashCode methods for the class, but their removal isn't believed to
cause any issues.

In addition, we have moved away from using the custom HashCode class, in
favour of the new HashCode class that has been created in the .NET
Framework.  We didn't have access to this class before, since we were
still targetting .NET Framework 4.0.

## Motivation and Context

These properties had previously been marked as Obsolete, and since we
are releasing the 2.0.0 version of Chocolatey, the decision has been
taken to remove these properties completely.

## Testing

N/A - Due to the fact that this was deemed as a simple removal of code that had been marked as Obsolete already.

## Change Types Made

## Related Issue

Fixes #2630

## Change Checklist


AdmiringWorm (changes requested):
The changes themself looks good, but it seems this PR breaks the build on the CI servers.

So it do seem that there is still something relying on these two properties (possibly something in test projects?)

AdmiringWorm:
@gep13 I have moved this PR into draft until we have verified whether it will break other products we provide or not.

AdmiringWorm:
Including this change requires more effort to remove any references to the settings values (mostly falling back to the properties being removed).
Additionally, we can not remove these properties as it will break compatibility with CCM v0.7.0.
Until we no longer support that version, we can not merge these changes.

PR #2644, opened 2022-03-10
(#2614) Update CREDITS.md with relevant contributors, committers and licenses

pauby:
Updated the CREDIS.md file with modern / relevant information that reflects the project today:

* Contributors
* Community Team
* Committers
* 7zip and Shimgen licenses

Note that the Shimgen license _prior_ to this update pointed to a specific commit. Because Shimgen, with the new license hasn't been pulled in yet, I've just pointed to master. This may or may not need to be amended once Shimgen is pulled in.

Finally, I added a whitespace commit as the Markdown linter was shouting at me.

This closes #2614

PR #2645, opened 2022-03-10
(#2615/#2549) Reference CREDITS file for other licenses

### LINKED ISSUE #2549: Choco licensing terms unclarify
Choco is licensed under apache public license - 

https://github.com/chocolatey/choco/blob/develop/LICENSE (1)

I think it's more or less clear - and it gives also freedom to anyone who will make a choco fork - to patch or improve that code anyhow fork owner wants to.

But perhaps one unclear issue is licensing of ShimGen:

https://github.com/chocolatey/choco/blob/develop/src/chocolatey.resources/tools/shimgen.license.txt (2)

By default if you make a fork of choco - you will use also ShimGen as well, which has more restrictive license.

I would propose that (1) should be a list of licenses, where choco license will be in a separate file.

LICENSE => (copy to) => CHOCOLAY_LICENCE
LICENSE - (modify to):
  list of licenses, include into list CHOCOLAY_LICENCE and shimgen license.

It would be good if by default if you make chocolatey fork and compile it - ShimGen functionality would be also disabled completely
- to use more open licensing. And if fork maintainer wants to use shimgen - we would enable it, and get written permission from chocolatey (more functionality).

Ideally however it would be nice if (2) would be equal to (1), as it gives more freedom to chocolatey forks and no need to explicitly disable shimgen.

### LINKED ISSUE #2615: Reference CREDITS.md in project NOTICE file
### What You Are Seeing?

The NOTICE file in the root of the project does not reference the other licenses that the project uses. These are held in the docs/legal/CREDITS.md file.

### What is Expected?

The NOTICE file should be make reference to the other licenses in CREDITS.md file.

N/A.

### Output Log

N/A.

pauby:
In response to [this issue](https://github.com/chocolatey/choco/issues/2549) the CREDITS.md file has been updated with the latest Shimgen license (see #2644) and this PR will reference the updated CREDITS file in the NOTICE file.

Open to changes in the wording.

Closes #2615  #2549

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -12,3 +12,7 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+   Additional legal information and licenses used by Chocolatey CLI can be
+   found in the CREDITS file or at
+   https://github.com/chocolatey/choco/blob/develop/docs/legal/CREDITS.md


AdmiringWorm (approved):
LGTM 👍

PR #2647, opened 2022-03-10
(#2646) Use organizational level templates

### LINKED ISSUE #2646: Use organizational level issues and pull request templates
We have created [organizational level](https://github.com/chocolatey/.github) templates to be used across the repositories, where appropriate. We should look to use them for this repository.

Note that the [organizational templates do not cover all the options](https://github.com/chocolatey/.github/blob/main/.github/ISSUE_TEMPLATE/config.yml) that you are offered when raising an issue here. We can always add more later if needed.

The Security file for this repository should also be replaced with the [organizational level file](https://github.com/chocolatey/.github/blob/main/.github/SECURITY.md)

pauby:
This PR removes all of the repository templates which is required to use the [organizational level templates](https://github.com/chocolatey/.github).

Closes #2646 

Clickup: ENGTASKS-577


AdmiringWorm:
As we spoke about ten days ago, I tested out whether adding a `config.yml` file alone into this repository would override the issue templates or not, and I can confirm a `config.yml` file will override the issue templates.

There is no additional work needed for this PR in this repository. However, we may think about having an extra link item in the `.github` repository that points people to the correct repositories to open issues.

gep13:
@pauby @AdmiringWorm I haven't really been involved in this side of things... are we in a position to merge this, or is additional work required?

AdmiringWorm:
I honestly can't remember if this PR was just put into draft until adding a standalone `config.yml` would override the templates or not, or if there was some additional work that was needed (I don't think there was though, at least in this repository).

Maybe @pauby could confirm what was the case here?

pauby:
@AdmiringWorm I'm fairly sure it was. But I was hoping that because a local `config.yml` file will allow us to override the `'github` one so we could add in 'Package Issue' and maybe some others for the templates we are removing locally?

If we have some suggestions I can add them in or @AdmiringWorm @gep13 feel free to update this PR.

AdmiringWorm:
> @AdmiringWorm I'm fairly sure it was. But I was hoping that because a local `config.yml` file will allow us to override the `'github` one so we could add in 'Package Issue' and maybe some others for the templates we are removing locally?
> 
> If we have some suggestions I can add them in or @AdmiringWorm @gep13 feel free to update this PR.

If we want to override the links already specified in the `.github` repository, we also need to copy over the templates themself as well, otherwise, we won't have any templates at all for this repository.

pauby:
I thought we could override just the `config.yml` and it would pick up the templates? The `config.yml` over on the [`.github`](https://github.com/chocolatey/.github/blob/main/.github/ISSUE_TEMPLATE/config.yml) repository doesn't mention anything about the templates themselves.

AdmiringWorm:
@pauby that is what I spoke about in comment: https://github.com/chocolatey/choco/pull/2647#issuecomment-1073573198 where I had tested whether we could just add a config.yml file, without it overriding the actual templates from the `.github` repository.

Unfortunately, just adding a `config.yml` also overrides those templates

AdmiringWorm:
A potential solution I have been looking at regarding that problem after talking with @gep13 , is to have an action running on the `.github` repository that can sync the necessary templates to those repositories that we have that need to override the templates, add a `config.yml` file, or if one needs additional templates.

Not gotten to something I believe will work appropriately yet though.

gep13 (approved):
LGTM!

gep13:
@pauby thanks for getting this updated!

PR #2648, opened 2022-03-10
(#1479) Add ability to use remembered arguments during uninstall

### LINKED ISSUE #1479: Please persist package installation parameters for uninstall scripts
Imagine you're not relying on an MSI; you're just bin-deploying to a default (overridable) location and adding an entry to %path% for the machine (overridable to 'user' or 'none').

<details>

```ps1
$ErrorActionPreference = 'Stop'

$pp = Get-PackageParameters

$pathType =
    if ('None', 'User', 'Machine' -contains $pp.PathType) { $pp.PathType }
    elseif (!$pp.PathType) { 'Machine' }
    else { throw "Unrecognized PathType '$($pp.PathType)'" }

$installDir = if ($pp.InstallDir) { $pp.InstallDir } else { "$Env:ProgramFiles\Elevate" }
$toolsPath = Split-Path -Parent $MyInvocation.MyCommand.Definition
$platform = if (Get-OSArchitectureWidth 64) { 'x64' } else { 'x86' }

New-Item $installDir -ItemType Directory -Force | Out-Null

Get-ChildItem "$toolsPath/$platform" |
    Move-Item -Destination $installDir -Force

Remove-Item "$toolsPath/*" -Recurse -Exclude *.ps1

if ($pathType -ne 'None') { Install-ChocolateyPath $installDir -PathType $pathType }
```

</details>

(Even if this utility worked when shimmed, there's still the question of the configurable installation location.)

When the user types `choco uninstall thepackage`, how should `chocolateyUninstall.ps1` find the correct installation folder to delete and the correct path variable (machine/user/none) to modify?

I would have expected `Get-PackageParameters` to be persisted by Chocolatey and provided to the uninstall script. It makes sense to allow them to be individually overridden by passing `--params` to `choco uninstall`, but it doesn't make sense for them to start null and thus require the user to duplicate the exact parameters used previously if the user wants a successful uninstall. That's not very user-friendly.

Also, I don't want to have to author an MSI just to have a decently reliable package uninstaller. (I notice that the only other package I can find that [allows a user/machine switch](https://github.com/rgra/choco-packages/blob/51a38523b8fc0212188a55e9bc7bf0342a1a844b/server-jre8/tools/chocolateyInstall.ps1#L70) only [uninstalls the machine](https://github.com/rgra/choco-packages/blob/51a38523b8fc0212188a55e9bc7bf0342a1a844b/server-jre8/tools/chocolateyUninstall.ps1) path.) 

Is there a reason we can't make this just start working out of the box? I'd expect this to apply to uninstalls for sure; probably upgrades as well. Seems like the obvious place to store the `--params` string is the same place that keeps a record that the package is installed.

TheCakeIsNaOH:
## Description Of Changes

WIP

## Motivation and Context

WIP

## Testing

Check that this works:
```
.\choco.exe feature enable -n UseRememberedArgumentsForUninstalls`
.\choco.exe install upgradepackage --version=1.0.0 --source=.\context --package-parameters="/PKGPARAM"
.\choco.exe uninstall upgradepackage --verbose --debug
```
Validate that during the uninstall log, the `chocolateyUninstall.ps1` is called with `/PKGPARAM` included.

Check that arguments are not reused:
WIP

Check that `--ia` are not used:
WIP

Check that the ignore switch works:
WIP

Check that the use switch works:
WIP

## Change Types Made

## Related Issue

Fixes #1479

## Change Checklist


TheCakeIsNaOH:
Need to update the config resetting to use the new version from #2484.

TheCakeIsNaOH:
This is draft until #3003 is merged/closed, as the implementation of this depends on what happens with that PR.

PR #2649, opened 2022-03-11
(#72) Add ability to include sources from config

### LINKED ISSUE #72: When specifying source, also have `--include-configured-sources`
So that when testing things depending on packages, its just easier.

TheCakeIsNaOH:
## Description Of Changes

This adds the --include-configured-sources switch which will append
any sources in the config to manually specified sources (via --source).
It checks if the source already exists before appending, which prevents
duplication when the manually specified sources and the sources from
config are combined.

## Motivation and Context

When testing a package locally, it may have depencies that are not installed, and available on a remote repository, like the Community Repository. This allows installation/upgrading of the package without having to specify that remote repository explicitly in `--source`, assuming it is in the sources configured with the source command. 

## Testing

Validate that the switch is ignored when using an alternate source:
`.\choco.exe install nonexistant-package --source=python --include-configured-sources`
Output should include this message: `Not including sources from config because python is an alternate source`

Validate that config sources are not included when not using switch: 
`.\choco.exe install nonexistant-package --source .`
Output error should specify source as `'.'`, and not include the default community repository.

Validate that config sources are included when using switch:
`.\choco.exe install nonexistant-package --source . --include-configured-sources`
Output error should specify sources as `'.;https://community.chocolatey.org/api/v2/'`

Validate that config sources are not duplicated if included both in `--source` and config:
`.\choco.exe install nonexistant-package --source="https://community.chocolatey.org/api/v2/" --include-configured-sources` 
Output error should specify source as `'https://community.chocolatey.org/api/v2/'`

## Change Types Made

## Related Issue

Fixes #72

## Change Checklist


pauby:
The name of the switch `--include-machine-sources` I don't feel is consistent with other switches. We don't use the word `machine` in any other switches so it's not intuitive.

The issue talks about `--include-default-sources` but given the work so make a source a default push source, that may also be confusing.

TheCakeIsNaOH:
I've updated the switch name to `--include-sources-from-config`

gep13:
What about `--include-configured-sources` for the option name?

TheCakeIsNaOH:
> What about --include-configured-sources for the option name?

Sounds good to me. Should I switch it over to that?

gep13 (approved):
LGTM!

gep13:
@TheCakeIsNaOH thanks, again, for all your work here! Really appreciate it!

PR #2667, opened 2022-03-28
(#2660) Add runtime check for compatibility with installed Chocolatey Licensed Extension

### LINKED ISSUE #2660: Support runtime checks for incompatible Chocolatey CLI and Chocolatey Licensed Extension versions
## Description

We need to provide information, during execution of choco.exe, when it is known that the version being used "could" be incompatible with the version of the Chocolatey Licensed Extension that is installed.  The information should contain links to additional information on how to resolve these incompatibilities, or recommendations on how to proceed if the installed package versions can not be changed, for example, in the case of perpetual licenses.  These warning should appear both at the very start of the operation that is being performed, as well as at the end of the operation.

In addition, there should be a feature flag (and associated options) that allow these warning to be hidden.  It is intended that this feature would be used by those using a perpetual version of Chocolatey Licensed Extension where an upgrade isn't possible, and that they are aware of the incompatibilities and are happy to proceed with them in place.  The warning that is shown should mention how this feature can be enabled.

## Why Is It Needed

When installing a Chocolatey package, Chocolatey can ensure that any dependencies on the package are installed prior to installing the package.  The same is not true in the reference direction.

For example, let's say that Chocolatey Licensed Extension v4.0.0 requires Chocolatey v1.0.0.  Installing the Chocolatey Licensed Extension package will result in v1.0.0 of Chocolatey being installed.  However, there is nothing to stop someone directly installing v1.0.0 of the Chocolatey package, and not first installing the v4.0.0 version of the Chocolatey Licensed Extension package.

As a result of the above, it is possible that an older version of Chocolatey Licensed Extension is being used with a newer version of Chocolatey, or vice versa (in very extreme situations).  We need to provide a warning to the user when this is the case, so that they can take informed action on how to move forward.

## Related Issues and Tickets

* #2659 
* https://gitlab.com/chocolatey/collaborators/choco-licensed/-/issues/606 - addition of the Assembly attribute which will contain information about minimum supported Chocolatey Version information

## Acceptance Tests

1. Install newer version of Chocolatey with older version of Chocolatey Licensed Extension installed
1. Perform any Chocolatey operation
  * See warning at start of operation
  * See warning at end of operation
1. Enable feature to hide warnings
1. Perform any Chocolatey operation
  * No warning at start of operation
  * No warning at end of operation
1.. Repeat above having a newer version of Chocolatey Licensed Extension installed compared to installed version of Chocolatey.  The installed version should be a major version difference compared to the minimum supported version of Chocolatey.  It will likely be necessary to fudge the AssemblyInfo in order to test this out properly. 

## Done Checklist

* [ ] Automated tests have been added to cover this bug.
* [ ] Manual tests have been added to cover this bug.

gep13:
## Description Of Changes

In order to provide as much support to users of Chocolatey as possible,
we have added the concept of a minimum required Chocolatey version
attribute to the Chocolatey Licensed Assembly.  When reading this
assembly (if present), we will extract this attribute, and then use it
to make assertions about whether things are currently known to be
compatible.

When it is found that they are not compatible, a warning is shown
before and after the command execution to try to make it as clear as
possible that things might not work as expected.

A new feature flag has been introduced, disableCompatiblityChecks,
which a user can opt-in to, to prevent the warnings from showing, but
this is disabled by default.  The warning shows how this feature can be
turned on if they are sure that they know what they are doing.

In addition, a command line option has been added to allow by passing
these warnings without having to enable the feature, with the
skip-compatibility-checks option.

In addition, due to the fact that we can, and do, load the Chocolatey
Licensed Assembly from memory, the Location property can be null.  When
this is the case, and it hasn't been possible to extract the version
information, we reach into the custom attributes for the assembly to
pull out the information, as this is needed to assert on the
compatibility of the versions.

## Motivation and Context

This is required to protect people when there are known incompatibilities
with the installed version of the Chocolatey Licensed Extension.  This check
needs to happen before reflecting into the assembly to override the IoC
registrations, which is why this check needs to be done in Chocolatey CLI.

## Testing

1. Get a system that has Chocolatey 0.12.1 and Chocolatey Licensed Extension v3.2.0 installed
1. Run choco list -lo and ensure that it outputs as expected
1. Install the built alpha package from this PR
1. Run choco list -lo and see that there is a new warning before and after the command has executed
1. Run choco list -lo --skip-compatibility-checks and see that the warnings are not shown
1. Run choco feature enable --name="'disableCompatibilityChecks'"
1. Run choco list -lo and see that the warnings are not shown
1. Upgrade to the latest Chocolatey Licensed Extension
1. Run choco list -lo and see that the warnings are not shown
1. Run choco feature enable --name="'disableCompatibilityChecks'"
1. Run choco list -lo and see that the warnings are not shown

## Change Types Made

## Related Issue

Fixes #2660 

## Change Checklist


gep13:
@corbob @vexx32 I have just done some more tests on this one, and it is ready for review.

vexx32 (approved):
This all looks good to me, and locally it appears to be behaving as expected. Thanks for sorting this out! 💖

PR #2670, opened 2022-03-29
Mark Foss as compatible with Foss

corbob:
This marks Foss as compatible when the license type is set.


gep13:
This PR is not required, as we ended up moving forward with this PR #2669.

It wasn't clear which direction was going to be taken, so @corbob created two PR's at the same time, with both options.

PR #2664, opened 2022-03-25
Check licensed assembly version during install

vexx32:
## Description Of Changes

Added a check into the chocolateyInstall.ps1 for Chocolatey to check the metadata on the `chocolatey.extension` assembly if it is currently installed/loaded and emit warnings if there is good reason to suspect the assembly may not be fully compatible with the version of Chocolatey being installed.

/cc @gep13 

NOTE: still in draft as there is still some testing to be done to verify this all works as expected. I hope to have that completed Monday morning.

## Motivation and Context

Helps to avoid issues with incompatible licensed extension versions, and give users some resources to help resolve possible problems if they run into issues.

## Testing

1. Run the build.bat and copy the generated chocolatey nupkg to a c:\packages folder
2. Install v1.0.1 of Chocolatey from the community repository.
3. Install a version 3.x.x of `chocolatey.extension`
4. Install the newly built version of Chocolatey from the package in `c:\packages`.
5. You should see a yellow warning message emitted during the installation stating that the extension version is not known to be compatible with the version of Chocolatey being installed.
6. The installation should succeed.
7. Install version 4.0.0 of the `chocolatey.extension` package.
8. Reinstall the prerelease version of Chocolatey from the package in `c:\packages` with `--force` to force the reinstall.
9. No warnings should be emitted.
10. Test with a prerelease version of the extension that includes a MinimumChocolateyVersion attribute set to higher than 1.0.x and ensure that attempting to install Chocolatey in this scenario correctly also emits a warning (slightly different to the previous warning).
11. Check the installation still works on PowerShell v2.

## Change Types Made

* [ ] Bug fix (non-breaking change)
* [x] Feature / Enhancement (non-breaking change)
* [ ] Breaking change (fix or feature that could cause existing functionality to change)
* [ ] PowerShell code changes.

## Related Issue

Fixes #2659

## Change Checklist

* [ ] Requires a change to the documentation
* [ ] Documentation has been updated
* [ ] Tests to cover my changes, have been added
* [x] All new and existing tests passed.
* [ ] PowerShell v2 compatibility checked.

PR #2672, opened 2022-03-29
(#2659) Handle scenario where install has no buffer

### LINKED ISSUE #2659: Support installation checks for incompatible Licensed Extension when running `choco upgrade chocolatey`
## Description

We need to provide information, at the point of installation, when it is known that the Chocolatey package is being installed onto a system that "could" be incompatible.  The information should contain links to additional information on how to resolve these incompatibilities, or recommendations on how to proceed if the installed package versions can not be changed, for example, in the case of perpetual licenses.

## Why Is It Needed

When installing a Chocolatey package, Chocolatey can ensure that any dependencies on the package are installed prior to installing the package.  The same is not true in the reference direction.

For example, let's say that Chocolatey Licensed Extension v4.0.0 requires Chocolatey v1.0.0.  Installing the Chocolatey Licensed Extension package will result in v1.0.0 of Chocolatey being installed.  However, there is nothing to stop someone directly installing v1.0.0 of the Chocolatey package, and not first installing the v4.0.0 version of the Chocolatey Licensed Extension package.

Handling checking for known incompatibilities, with information on how to remedy these will help folks get the right things installed.

## Related Issues and Tickets

* #2660
* https://gitlab.com/chocolatey/collaborators/choco-licensed/-/issues/606 - addition of the Assembly attribute which will contain information about minimum supported Chocolatey Version information 

## Acceptance Tests

1. When installing Chocolatey Package with an older version (< 4.0.0) installed, warnings should be displayed indicating that there may not be complete compatibility with all functionality, and that some functions may not work as expected
2. When installing Chocolatey with a new version of Chocolatey Licensed Extension (say a higher major version compared to the minimum supported version of Chocolatey, provide a similar warning
3. Installation of the Chocolatey package should never be prevented from happening, but the warnings should make it clear that there may be incompatibilities

## Done Checklist

* [ ] Automated tests have been added to cover this bug.
* [ ] Manual tests have been added to cover this bug.

corbob:
## Description Of Changes

Ensure that we catch the exception if there is no buffer (PSRemoting, Background Agent, etc) and thus no buffer width.

An example of what happens when trying to upgrade (or reinstall) Chocolatey when run as Background agent: 
image

## Motivation and Context

Test Kitchen tests were failing on reinstalling the Chocolatey package but only in Test Kitchen runs, not when run manually.

## Testing

I'm starting with getting a PR up so I can get a build for my local test kitchen, then I will run it through the tests, in particular the ones that have been troublesome. The testing will also mirror #2664 

So my testing (on Windows 7 and Test Kitchen through PSRemoting):

1. Capture package from Team City server (`1.1.0-alpha-20220330-118`)
2. Push this package to internal repository (Test Kitchen needs it here to do it's reinstall)
3. Install v1.0.1 of Chocolatey from the community repository.
4. Install a version 3.x.x of `chocolatey.extension`
5. Install the newly built version of Chocolatey from the internal repository
6. You should see a yellow warning message emitted during the installation stating that the extension version is not known to be compatible with the version of Chocolatey being installed.
7. The installation should succeed.
8. Install version 4.0.0 of the `chocolatey.extension` package.
9. Reinstall the prerelease version of Chocolatey from the internal repository with `--force` to force the reinstall.
10. No warnings should be emitted.
11. Test with a prerelease version of the extension that includes a MinimumChocolateyVersion attribute set to higher than 1.0.x and ensure that attempting to install Chocolatey in this scenario correctly also emits a warning.
12. Check the installation still works on PowerShell v2.

And for completeness sake:

1. Install this build of Chocolatey, chocolatey.extension, and chocolatey-agent.
2. Enable Background agent: `choco feature enable -n useBackgroundService`
3. Force reinstall of Chocolatey: `choco install chocolatey --version 1.1.0-alpha-20220330-118 --force`
4. Verify that it installed successfully.

## Change Types Made

## Related Issue

Fixes #2659

## Change Checklist

* [x] PowerShell v2 compatibility checked.

### DIFF
--- a/nuget/chocolatey/tools/chocolateyInstall.ps1
+++ b/nuget/chocolatey/tools/chocolateyInstall.ps1
@@ -11,12 +11,19 @@ if ($null -ne $licensedAssembly) {
     # The licensed assembly is installed, check its supported Chocolatey versions and/or the assembly
     # version so we can attempt to determine whether it's compatible with this version of Chocolatey.
     $attributeData = $licensedAssembly.GetCustomAttributes($true)
-    
+
     $minVersion = $attributeData |
         Where-Object { $_.TypeId -like '*MinimumChocolateyVersion*' } |
         Select-Object -ExpandProperty Version -First 1
 
-    $messageBorder = '=' * ([System.Console]::BufferWidth - 10)
+    $borderWidth = 70
+    try {
+        $borderWidth = [System.Console]::BufferWidth - 10
+    } catch {
+        # Do nothing. This means we're in a non-interactive environment without a console attached.
+    }
+
+    $messageBorder = '=' * $borderWidth
     $extensionVersionWarning = @"
 $messageBorder
 


nuget/chocolatey/tools/chocolateyInstall.ps1
```diff
@@ -11,12 +11,19 @@ if ($null -ne $licensedAssembly) {
...
     $minVersion = $attributeData |
         Where-Object { $_.TypeId -like '*MinimumChocolateyVersion*' } |
         Select-Object -ExpandProperty Version -First 1
 
-    $messageBorder = '=' * ([System.Console]::BufferWidth - 10)
+    $borderWidth = 70
+    try {
+        $borderWidth = [System.Console]::BufferWidth - 10
+    } catch {
+        # Do nothing. This means we're in a service or remoting or something without a buffer.
```
vexx32, line 23: Probably could just adjust the wording of the comment a bit here, it feels a little unclear.

```suggestion
        # Do nothing. This means we're in a non-interactive environment without a console attached.

```

vexx32 (approved):
Looks good to me, great job catching this one, and thanks for fixing it up! 💖 🌸

gep13 (approved):
LGTM!

gep13:
@corbob thanks for catching this, and also for getting it fixed up!

PR #2713, opened 2022-05-04
Rewrite Report Package Issue using fancy yaml

jsoref:
# Before
image

# After
image


pauby:
This work was completed some time ago and the templates were moved to a central .github repository. Thanks for highlighting this @jsoref, and apologies for the huge delay in getting to this.

PR #2730, opened 2022-06-05
#2206 Add security protocols to config. Add TLS 1.3 protocol.

### LINKED ISSUE #2206: config - Set Minimum Security Protocol
It would be great if there is an option to set what protocol the choco client tool should use when doing an install or update.
Currently the server rejects any TLS1.2 connection but there is a connection that is already sent out from the choco client tool which could lead to some Man in the middle attack. The following line shows the configuration supports TLS1.1 and TLS1.2
https://github.com/chocolatey/choco/blob/master/src/chocolatey/infrastructure/registration/SecurityProtocol.cs#L39
It could help resolve some security issues.
Thank you

BElluu:
## Description Of Changes
The previous configuration allowed the use of SSL3, TLS 1.0, TLS 1.1, TLS 1.2 by default. An entry has been added to the configuration that allows you to specify which protocols are allowed. TLS 1.3 has also been added.
 
 ## Motivation and Context
Ability to decide which protocol should be used when using choco.
 
 ## Testing
`---`
 
 ## Change Types Made
 * [ ]  Bug fix (non-breaking change)
 * [x]  Feature / Enhancement (non-breaking change)
 * [ ]  Breaking change (fix or feature that could cause existing functionality to change)
 
 ## Related Issue
 Fixes #2206  
 
 ## Change Checklist
 * [x]  Requires a change to the documentation
 * [ ]  Documentation has been updated - added to tracking issue here:
 * [ ]  Tests to cover my changes, have been added
 * [x]  All new and existing tests passed.

### DIFF
--- a/src/chocolatey/infrastructure.app/ApplicationParameters.cs
+++ b/src/chocolatey/infrastructure.app/ApplicationParameters.cs
@@ -175,6 +175,7 @@ namespace chocolatey.infrastructure.app
             public static readonly string WebRequestTimeoutSeconds = "webRequestTimeoutSeconds";
             public static readonly string UpgradeAllExceptions = "upgradeAllExceptions";
             public static readonly string DefaultTemplateName = "defaultTemplateName";
+            public static readonly string AllowedSecurityProtocols = "allowedSecurityProtocols";
         }
 
         public static class Features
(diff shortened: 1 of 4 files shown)


BElluu:
Hi. I guess the previous developer doesn't want to do this anymore.
Anything else I should consider when carrying out this task?

gep13:
@BElluu apologies for the length of time that it has taken to get back to you regarding this PR.  We have been very busy preparing for the upcoming release of the Chocolatey CLI v2.0.0, as well as the other Chocolatey products.

Within that release we have changed the way that Chocolatey CLI deals with TLS versioning.  The only reason that we started changing how TLS was handled in Chocolatey CLI was due to the fact that Chocolaty CLI still supported the older .NET Framework 4.0, where newer versions of the framework that Chocolatey CLI could run on had added new TLS versions as well.  The creation of the SecurityProtocol class was an attempt to bridge the gap between running on the older .NET Framework and the new one.

With Chocolatey CLI 2.0.0, we now require the installation of .NET Framework 4.8, therefore the SecurityProtocol class is no longer needed, and the decision has been taken for Chocolatey CLI to use whatever has been set at the Operating System as the Default TLS value to use.  This puts the configuration back into the hands of the System Administrators, rather than at the Chocolatey level.

The changes in this regard were done against this [issue](https://github.com/chocolatey/choco/issues/2902) and in this [PR](https://github.com/chocolatey/choco/pull/3123).

As a result of these changes, it is felt that this PR, and the original one that you created from, are no longer required.  As such, I am going to close the PR.

Let me know if you have any questions.

PR #2747, opened 2022-06-26
chore: terminology corrections

mikeirvingweb:
Changes including spelling of MacOS -> macOS, OSX -> macOS, OS X -> macOS, .Net -> .NET, mono -> Mono in comments, docs and user information.


src/chocolatey/infrastructure/platforms/Platform.cs
```diff
@@ -85,7 +85,7 @@ public static string get_name()
                 case PlatformType.Linux:
                     return "Linux";
                 case PlatformType.Mac:
-                    return "OS X";
+                    return "macOS";
```
gep13, line 88: Initially we were concerned that this change would be a breaking change, however, looking at the various code paths in place, the decision was taken that this won't impact anyone, and the change makes sense in line with the branding of this OS.

gep13 (approved):
LGTM!

gep13:
@mikeirvingweb thank you for taking the time to fix this up, really appreciate it!

mikeirvingweb:
> @mikeirvingweb thank you for taking the time to fix this up, really appreciate it!

Happy to help

PR #2756, opened 2022-07-01
(#1392) Update refreshenv to detect when not in cmd shell

### LINKED ISSUE #1392: Make refreshenv.bat error if powershell.exe is being used
If someone has the powershell profile installed, they would never call this as refreshenv is an alias for Update-SessionEnvironment.

This is quite difficult to determine. I found https://stackoverflow.com/a/16376399/18475, but it won't work if you drop from PowerShell.exe to cmd.exe to run commands, and I haven't found a way to capture the current title and set it back yet.

It also doesn't determine the difference between cmd started up just to run a batch and a long term child process of cmd.exe that you get if you open PowerShell and then drop into cmd.exe (call `cmd.exe` in PowerShell opens a child process). The script found can not tell the difference between the two.

corbob:
## Description Of Changes

Update the `refreshenv.cmd` file to detect if being run from a sub shell and properly report that it doesn't work in that scenario.

## Motivation and Context

By default, if you don't have a PowerShell profile when Chocolatey is first installed, you don't get the `refreshenv` alias from the `chocolateyProfile.psm1` PowerShell module. This means if you run `refreshenv` while in PowerShell it actually runs `refreshenv.cmd` and fails to refresh the environment.

## Testing

1. Build the solution with `./build.bat`
2. Install the built files `choco upgrade chocolatey -pre -y -s ./code_drop/Packages/Chocolatey`
3. Run `refreshenv` in a PowerShell session
4. Verify output regarding not supported there
5. Import module referenced in output
6. Run `refreshenv` again
7. Verify it works as expected
8. Run `refreshenv` from a `cmd.exe` session
9. Verify output of refreshing variables for cmd.exe

## Change Types Made

## Related Issue

Fixes #1392 

## Change Checklist


vexx32 (approved):
Looks good to me! 💖

PR #2823, opened 2022-09-21
TBD

corbob:
This adds tests for pin during install/upgrade, and for hooks.

## Change Types Made

## Related Issue

Fixes #

## Change Checklist


tests/chocolatey-tests/Hooks.Tests.ps1
```diff
@@ -0,0 +1,93 @@
+Import-Module helpers/common-helpers
```
AdmiringWorm, line 1: Since this file explicitly test a specific feature, can you move this file into a `features` directory, since it doesn't follow the same conventions we already use by having the test file be `choco-` command name.

tests/chocolatey-tests/Hooks.Tests.ps1
```diff
@@ -0,0 +1,93 @@
+Import-Module helpers/common-helpers
+
+Describe "hooks tests" -Tag Chocolatey, Hooks {
```
AdmiringWorm, line 3:
```suggestion
Describe "choco hooks tests" -Tag Chocolatey, HooksFeature {

```

tests/chocolatey-tests/Hooks.Tests.ps1
```diff
@@ -0,0 +1,93 @@
...
+                }
+                $null = Invoke-Choco install $Package --version $ver --no-progress
+            }
+
+            $Output = Invoke-Choco $Command $Package $Flag --no-progress
+        }
+
+        # Uninstall/Upgrade exit -1: https://github.com/chocolatey/choco/issues/2822
+        It "Exits with Success (0,-1)" {
+            $Output.ExitCode | Should -BeIn @(0, -1) -Because $Output.String
```
AdmiringWorm, line 51: This will have false positives.

What would happen if the install command fails with exit code `-1`?

Would be better to separate this test into two separate tests where one is skipped on install, and the other only runs on install instead of allowing both `0` and `-1` values for all commands.

AdmiringWorm (changes requested):

tests/chocolatey-tests/Hooks.Tests.ps1
corbob: I've moved this and a few other files around as it seemed to make sense.

tests/chocolatey-tests/Hooks.Tests.ps1
corbob: These make sense 😄 naming things is hard 😅

tests/chocolatey-tests/Hooks.Tests.ps1
corbob: Good points. I've extracted the expected value into the ForEach and so we can be explicit about which one we expect.

tests/chocolatey-tests/commands/choco-upgrade.Tests.ps1
```diff
@@ -120,6 +120,33 @@ Describe "choco upgrade" -Tag Chocolatey, UpgradeCommand {
         }
     }
 
+    Context "Upgrade package with (<Command>) specified" -ForEach @(
+        @{ Command = '--pin' ; Contains = $true }
+        @{ Command = '' ; Contains = $false }
+    ) {
+        BeforeAll {
+            Restore-ChocolateyInstallSnapshot -SetWorkDir
```
AdmiringWorm, line 128: You don't use anything in the working directory, because of this, you don't need to specify the `-SetWorkDir`.

It is only useful if you need the current `CWD` to be in the working directory.
Typically it is never needed.

Not requiring a change, just mentioning it so you are aware of it in the future.

corbob: oooh, I think I copied the context block from somewhere else that used it. I've removed this from both the install and upgrade files.

AdmiringWorm (approved):
LGTM

AdmiringWorm:
@corbob thank you for getting this added and fixed up 👍

PR #2834, opened 2022-09-27
(#1185) Always pass in hook scripts event if empty

### LINKED ISSUE #1185: Allow post/pre hooks to run on upgrade/install/uninstall
The idea for this have been taken from pacman (the main package manager for arch linux).
One usage example on how it is used in pacman would be running reflector after pacman-mirrorlist package is updated: https://wiki.archlinux.org/index.php/Reflector#Pacman_Hook

What I request is similiar: have a directory (either configurable or hard coded to `$env:ChocolateyInstall\hook` that stores a bunch of powershell scripts that are meant to be run after/before a package update/install (or after choco have completed the full update/install of all packages).

The way I'm thinking this could work is through a naming convention. ie: `(pre|post){pacakge-id}-upgrade/install/uninstall.ps1` or `(pre|post)global-upgrade/install/uninstall.ps1`.

The reason I would like to see something like this implemented is because quite a few times I have a need/want to automatically configure some extra configuration for the package(s) that I'm installing (usually user specific configurations) (could range from anything as easy as removing a shortcut, to full configuration of git, virtualbox, smartgit, or even choco. as well as others).

corbob:
## Description Of Changes

Always pass the hook parameters to `chocolateyScriptRunner` even if we're just passing them `$null`

## Motivation and Context

There is a manual test that we have where we try to install a package thusly: `choco install test-params --package-parameters="/NoDesktop 'Not Displayed' /InstallDir:'C:\tools\package destination'"` this was resulting in the following in the log: `2022-09-26 16:08:41,822 7016 [DEBUG] - Running 'ChocolateyScriptRunner' for test-params v0.0.0 with packageScript 'C:\ProgramData\chocolatey\lib\test-params\tools\chocolateyinstall.ps1', packageFolder:'C:\ProgramData\chocolatey\lib\test-params', installArguments: '', packageParameters: '/NoDesktop ', preRunHookScripts: 'Not', postRunHookScripts: 'Displayed /InstallDir:C:\tools\package',` which is clearly not correct.

## Testing

1. I have run this through the debugger with the parameters `install test-params --package-parameters="/NoDesktop 'Not Displayed' /InstallDir:'C:\tools\package destination'"`
2. Confirmed that it installed as I expected it to.

## Change Types Made

## Related Issue

Fixes #1185 

## Change Checklist

### DIFF
--- a/src/chocolatey/infrastructure.app/services/PowershellService.cs
+++ b/src/chocolatey/infrastructure.app/services/PowershellService.cs
@@ -194,14 +194,14 @@ namespace chocolatey.infrastructure.app.services
 
         private string get_script_arguments(string script, IEnumerable<string> hookPreScriptPathList, IEnumerable<string> hookPostScriptPathList, ChocolateyConfiguration config)
         {
-            return "-packageScript '{0}' -installArguments '{1}' -packageParameters '{2}'{3}{4}{5}{6}".format_with(
+            return "-packageScript '{0}' -installArguments '{1}' -packageParameters '{2}'{3}{4} -preRunHookScripts {5} -postRunHookScripts {6}".format_with(
                 script,
                 prepare_powershell_arguments(config.InstallArguments),
                 prepare_powershell_arguments(config.PackageParameters),
                 config.ForceX86 ? " -forceX86" : string.Empty,
                 config.OverrideArguments ? " -overrideArgs" : string.Empty,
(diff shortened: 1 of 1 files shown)


AdmiringWorm (dismissed):
LGTM, and I can confirm that this works as expected. Oddly enough, the problem you noticed does not happen in Debug builds, and only when a release artifact is created.

AdmiringWorm (approved):

PR #2836, opened 2022-09-28
(#886, 2693) Update packages.config install logging and add back resolve or load method for Assembly Resolution

### LINKED ISSUE #886: Packages.config should include all options (including ignorechecksums)

### LINKED ISSUE #2693: Enhance registrations to be able to load officially signed extensions
Currently when/if we want to update existing commands or add new commands in other projects using Chocolatey CLI we need to replace all commands by re-registering the commands and add any new/updated commands to the same list that we are re-registering.

This work item is to make this process more streamlined/easier to do, so we only need to adde/replace the commands we are interested in.

The exact scope of work is not known at this time, but this issue is a placeholder to do this work.

corbob:
## Description Of Changes

Update the debugging output for the install from packages.config file to clearly delineate the start and end of the configuration output.

Add back the original definition of resolve_or_load_assembly to AssemblyResolution.

## Motivation and Context

The output of the Configuration when Licensed Extension is installed is different from when just Open Source is in play. As such, the tests for the new functionality was failing. This updates the output to enable us to better test the command.

Chocolatey GUI fails to load with the update to `resolve_or_load_assembly`, so added back the original definition.

## Testing

1. Downloaded build artifacts from build server and installed on VM
2. Installed Chocolatey GUI
3. Launched Chocolatey GUI
4. Confirmed it launched successfully
5. Installed Licensed Extension
6. Ensured the Pester tests for InstallCommand completed successfully.

## Change Types Made

## Related Issue

Fixes #886
Fixes #2693 

## Change Checklist

### DIFF
--- a/src/chocolatey/infrastructure/registration/AssemblyResolution.cs
+++ b/src/chocolatey/infrastructure/registration/AssemblyResolution.cs
@@ -37,6 +37,18 @@ namespace chocolatey.infrastructure.registration
         /// <param name="assemblySimpleName">Simple Name of the assembly, such as "chocolatey"</param>
         /// <param name="publicKeyToken">The public key token.</param>
         /// <param name="assemblyFileLocation">The assembly file location. Typically the path to the DLL on disk.</param>
+        /// <returns>An assembly</returns>
+        /// <exception cref="Exception">Unable to enter synchronized code to determine assembly loading</exception>
+        public static IAssembly resolve_or_load_assembly(string assemblySimpleName, string publicKeyToken, string assemblyFileLocation) {
+            return resolve_or_load_assembly(assemblySimpleName, publicKeyToken, assemblyFileLocation, false);
+        }
+
+        /// <summary>
+        /// Resolves or loads an assembly. If an assembly is already loaded, no need to reload it.
+        /// </summary>
+        /// <param name="assemblySimpleName">Simple Name of the assembly, such as "chocolatey"</param>
+        /// <param name="publicKeyToken">The public key token.</param>
+        /// <param name="assemblyFileLocation">The assembly file location. Typically the path to the DLL on disk.</param>
         /// <param name="ignoreExisting">Whether any existing library that has previously been loaded should be ignored or not.</param>
         /// <returns>An assembly</returns>
         /// <exception cref="Exception">Unable to enter synchronized code to determine assembly loading</exception>
--- a/tests/chocolatey-tests/commands/choco-install.Tests.ps1
+++ b/tests/chocolatey-tests/commands/choco-install.Tests.ps1
@@ -247,9 +247,11 @@ Describe "choco install" -Tag Chocolatey, InstallCommand {
                 $Output = Invoke-Choco install $env:CHOCOLATEY_TEST_PACKAGES_PATH\alloptions.packages.config --confirm --verbose --debug
 
                 # This is based on two observations: The addition explicitly outputs that it's the Package Configuration.
-                # The configuration output is also about 80 lines.
-                $StartofPackageConfiguration = [array]::IndexOf($Output.Lines, "Package Configuration: CommandName='install'|")
-                $PackageConfigurationOutput = $Output.Lines[$StartofPackageConfiguration..($StartofPackageConfiguration+80)] -join [Environment]::NewLine
+                # The configuration output is about 80 lines.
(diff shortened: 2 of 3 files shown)


src/chocolatey/infrastructure/registration/AssemblyResolution.cs
```diff
@@ -31,6 +31,18 @@ public class AssemblyResolution
...
 
+        /// <summary>
+        /// Resolves or loads an assembly. If an assembly is already loaded, no need to reload it.
+        /// </summary>
+        /// <param name="assemblySimpleName">Simple Name of the assembly, such as "chocolatey"</param>
+        /// <param name="publicKeyToken">The public key token.</param>
+        /// <param name="assemblyFileLocation">The assembly file location. Typically the path to the DLL on disk.</param>
+        /// <returns>An assembly</returns>
+        /// <exception cref="Exception">Unable to enter synchronized code to determine assembly loading</exception>
+        public static IAssembly resolve_or_load_assembly(string assemblySimpleName, string publicKeyToken, string assemblyFileLocation) {
```
gep13, line 42: If the intention is to remove this, should we not mark it as obsolete?

AdmiringWorm: From what Cory told be on Slack, it is being used by Chocolatey GUI, and since it is just an overload of the other method there isn't really a need to deprecate it.

Doesn't hurt to just keep it around.

tests/chocolatey-tests/commands/choco-install.Tests.ps1
```diff
@@ -247,9 +247,11 @@ Describe "choco install" -Tag Chocolatey, InstallCommand {
                 $Output = Invoke-Choco install $env:CHOCOLATEY_TEST_PACKAGES_PATH\alloptions.packages.config --confirm --verbose --debug
 
                 # This is based on two observations: The addition explicitly outputs that it's the Package Configuration.
-                # The configuration output is also about 80 lines.
-                $StartofPackageConfiguration = [array]::IndexOf($Output.Lines, "Package Configuration: CommandName='install'|")
-                $PackageConfigurationOutput = $Output.Lines[$StartofPackageConfiguration..($StartofPackageConfiguration+80)] -join [Environment]::NewLine
+                # The configuration output is about 80 lines.
+                $StartOfPackageConfiguration = [array]::IndexOf($Output.Lines, "Package Configuration Start:")
+                $EndOfPackageConfiguration = [array]::IndexOf($Output.Lines, "Package Configuration End")
```
AdmiringWorm, line 252: Not so sure about just updating the logic here just to adhere to tests.

Not pushing back against it, just mentioning it feels a bit wrong when that is used as the excuse for the change.

corbob: Yeah, I'd tend to agree it's not great. However, the alternatives that I could think of would make the tests quite fragile and mean that any configuration change (either adding configuration options, or changing the default settings) could break the tests.

src/chocolatey/infrastructure/registration/AssemblyResolution.cs
corbob: I'm good with marking this as obsolete if we want. A benefit of not obsoleting it is that anyone currently using Chocolatey.Lib doesn't need to do anything since this just calls the new method with the default value of that method anyways.

AdmiringWorm (approved):
LGTM, but need to leave it up to @gep13 to decide whether we should mark the overload as obsolete or not (as already mentioned, I don't mind either way).

gep13:
If everyone is happy, then I would say let's not add the `Obsolete` attribute, and leave the overload for the method in place.

gep13:

PR #2847, opened 2022-10-04
(#2829) Adds Basic choco license Command

### LINKED ISSUE #2829: Add `choco license` command
### What You Are Seeing?
As a system administrator I would like to be able to view the license information from the command prompt

### What is Expected?
* The registered name of the license
* The expiration date
* The license type
* The license count (total, not usage)
* Anything else that might be useful

-- at some point in the future, i.e. not now, it would be nice to have:

* export to a `json` / `xml` format

I know this is all available in the `chocolatey.license.xml` file, but I think it would be usable to put this into an new `choco license` command, or perhaps extend the `choco support` command.

┆Issue is synchronized with this [Gitlab issue](https://gitlab.com/chocolatey/collaborators/choco-licensed/-/issues/384) by [Unito](https://www.unito.io)

JPRuskin:
## Description Of Changes
This PR adds a delightful new command, `choco license`.

It should offer read _and_ write, but at the moment just offers a basic output of the current license.

It doesn't yet have tests, either.

## Motivation and Context
This has been requested by awesome folk!

## Testing
1. We have tested it locally

## Change Types Made

## Related Issue

Fixes #2829 

## Change Checklist

### DIFF
--- a/src/chocolatey.tests/infrastructure.app/commands/ChocolateyLicenseCommandSpecs.cs
+++ b/src/chocolatey.tests/infrastructure.app/commands/ChocolateyLicenseCommandSpecs.cs
@@ -1,0 +1,265 @@
+﻿// Copyright © 2017 - 2025 Chocolatey Software, Inc
+// Copyright © 2011 - 2017 RealDimensions Software, LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+//
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using chocolatey.infrastructure.app.attributes;
+using chocolatey.infrastructure.app.commands;
+using chocolatey.infrastructure.app.configuration;
+using chocolatey.infrastructure.app.domain;
+using FluentAssertions;
+using Moq;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace chocolatey.tests.infrastructure.app.commands
+{
+    public class ChocolateyLicenseCommandSpecs
+    {
+        [ConcernFor("license")]
+        public abstract class ChocolateyLicenseCommandSpecsBase : TinySpec
+        {
+            protected ChocolateyLicenseCommand Command;
+            protected ChocolateyConfiguration Configuration = new ChocolateyConfiguration();
+
+            public override void Context()
+            {
+                Command = new ChocolateyLicenseCommand();
+            }
+        }
+
+        public class When_Implementing_Command_For : ChocolateyLicenseCommandSpecsBase
+        {
+            private List<CommandForAttribute> _results;
+
+            public override void Because()
+            {
+                _results = Command.GetType().GetCustomAttributes<CommandForAttribute>().ToList();
+            }
+
+            [Fact]
+            public void Should_Have_Expected_Number_Of_Commands()
+            {
+                _results.Should().HaveCount(1);
+            }
+
+            [InlineData("license")]
+            public void Should_Implement_Expected_Command(string name)
+            {
+                _results.Should().ContainSingle(r => r.CommandName == name);
+            }
+
+            [Fact]
+            public void Should_Specify_Expected_Version_For_All_Commands()
+            {
+                _results.Should().AllSatisfy(r => r.Version.Should().Be("2.5.0"));
+            }
+        }
+
+        public class When_parsing_additional_arguments_ : ChocolateyLicenseCommandSpecsBase
+        {
+            private readonly IList<string> _unparsedArgs = new List<string>();
+            private Action _because;
+
+            public override void Because()
+            {
+                _because = () => Command.ParseAdditionalArguments(_unparsedArgs, Configuration);
+            }
+
+            public new void Reset()
+            {
+                _unparsedArgs.Clear();
+            }
+
+            [Fact]
+            public void Should_use_the_first_unparsed_arg_as_the_subcommand()
+            {
+                Reset();
+                _unparsedArgs.Add("info");
+                _because();
+
+                Configuration.LicenseCommand.Command.Should().Be(LicenseCommandType.Info);
+            }
+
+            [Fact]
+            public void Should_throw_when_more_than_one_unparsed_arg_is_passed()
+            {
+                Reset();
+                _unparsedArgs.Add("wtf");
+                _unparsedArgs.Add("bbq");
+                var errored = false;
+                Exception error = null;
+
+                try
+                {
+                    _because();
+                }
+                catch (Exception ex)
+                {
+                    errored = true;
+                    error = ex;
+                }
+
+                errored.Should().BeTrue();
+                error.Should().NotBeNull();
+                error.Should().BeOfType<ApplicationException>();
+                error.Message.Should().Contain("A single license command must be listed");
+            }
+
+            [Fact]
+            public void Should_accept_info_as_the_subcommand()
+            {
+                Reset();
+                _unparsedArgs.Add("info");
+                _because();
+
+                Configuration.LicenseCommand.Command.Should().Be(LicenseCommandType.Info);
+            }
+
+            [Fact]
+            public void Should_accept_uppercase_info_as_the_subcommand()
+            {
+                Reset();
+                _unparsedArgs.Add("INFO");
+                _because();
+
+                Configuration.LicenseCommand.Command.Should().Be(LicenseCommandType.Info);
+            }
+
+            [Fact]
+            public void Should_set_unrecognized_values_to_info_as_the_subcommand()
+            {
+                Reset();
+                _unparsedArgs.Add("wtf");
+                _because();
+
+                Configuration.LicenseCommand.Command.Should().Be(LicenseCommandType.Info);
+            }
+
+            [Fact]
+            public void Should_default_to_list_as_the_subcommand()
+            {
+                Reset();
+                _because();
+
+                Configuration.LicenseCommand.Command.Should().Be(LicenseCommandType.Info);
+            }
+
+            [Fact]
+            public void Should_handle_passing_in_an_empty_string()
+            {
+                Reset();
+                _unparsedArgs.Add(" ");
+                _because();
+
+                Configuration.LicenseCommand.Command.Should().Be(LicenseCommandType.Info);
+            }
+        }
+
+        public class When_Help_Is_Called : ChocolateyLicenseCommandSpecsBase
+        {
+            public override void Because()
+            {
+                Command.HelpMessage(Configuration);
+            }
+
+            [Fact]
+            public void Should_log_a_message()
+            {
+                MockLogger.Verify(l => l.Info(It.IsAny<string>()), Times.AtLeastOnce);
+            }
+
+            [Fact]
+            public void Should_log_the_message_we_expect()
+            {
+                var messages = MockLogger.MessagesFor(LogLevel.Info);
+                messages.Should().HaveCount(19);
+                messages[0].Should().Contain("License Command");
+                messages[2].Should().Contain("Show information about the current Chocolatey CLI license.");
+            }
+        }
+
+        public class When_DryRun_Is_Called : ChocolateyLicenseCommandSpecsBase
+        {
+            public override void Because()
+            {
+                Configuration.LicenseCommand.Command = LicenseCommandType.Info;
+                Command.DryRun(Configuration);
+            }
+
+            [Fact]
+            public void Should_log_a_message()
+            {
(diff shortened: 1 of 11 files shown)


src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
```diff
@@ -0,0 +1,89 @@
...
+    using chocolatey.infrastructure.filesystem;
+    using chocolatey.infrastructure.licensing;
+    using chocolatey.infrastructure.logging;
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text;
+
+    [CommandFor("license", "Retrieve or modify the Chocolatey License")]
+    class ChocolateyLicenseCommand : ICommand
```
AdmiringWorm, line 32:
```suggestion
    public class ChocolateyLicenseCommand : ICommand
```

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        public void configure_argument_parser(OptionSet optionSet, ChocolateyConfiguration configuration)
+        {
+            // We don't currently expect to have any arguments
+            return;
```
AdmiringWorm, line 37: We generally don't specify `return` when it isn't needed.

```suggestion
            // We don't currently expect to have any arguments
```

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        public void handle_additional_argument_parsing(IList<string> unparsedArguments, ChocolateyConfiguration configuration)
+        {
+            // We don't currently expect to have any additional arguments
+            return;
```
AdmiringWorm, line 43: We generally don't specify `return` when it isn't needed.

```suggestion
            // We don't currently expect to have any additional arguments
```

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        public void handle_validation(ChocolateyConfiguration configuration)
+        {
+            // We don't currently accept any arguments, so there is no validation
+            return;
```
AdmiringWorm, line 49: We generally don't specify `return` when it isn't needed.

```suggestion
            // We don't currently accept any arguments, so there is no validation
```

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        public bool may_require_admin_access()
+        {
+            return false;
+        }
+
+        public void noop(ChocolateyConfiguration configuration)
+        {
+            return;
+        }
```
AdmiringWorm, line 68: We generally don't specify `return` when it isn't needed.

```suggestion
        {
        }
```

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            var ourLicense = LicenseValidation.validate();
+            
+            if (config.RegularOutput)
+            {
+                this.Log().Info("We have found a{0}valid license for Chocolatey {1}, which expires {2}".format_with((ourLicense.IsValid ? " " : "n in"), ourLicense.LicenseType, ourLicense.ExpirationDate));
+                this.Log().Info("Registered to:   {0}".format_with(ourLicense.Name));
+                this.Log().Info("Expiration Date: {0} UTC".format_with(ourLicense.ExpirationDate));
+                this.Log().Info("License type:    {0}".format_with(ourLicense.LicenseType));
+                this.Log().Info("Node Count:      {0}".format_with("<COUNT GOES HERE>"));
```
AdmiringWorm, line 80: Instead of blindly outputting this information, it should consider `FOSS` licenses (or missing licenses), and only notify of this instead of the normal output (maybe as a warning?)

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
```diff
@@ -0,0 +1,89 @@
...
+            return;
+        }
+
+        public void run(ChocolateyConfiguration config)
+        {
+            var ourLicense = LicenseValidation.validate();
+            
+            if (config.RegularOutput)
+            {
+                this.Log().Info("We have found a{0}valid license for Chocolatey {1}, which expires {2}".format_with((ourLicense.IsValid ? " " : "n in"), ourLicense.LicenseType, ourLicense.ExpirationDate));
```
AdmiringWorm, line 76: Just something to take into consideration.

If we ever want to translate Chocolatey CLI to other languages, we should not do this conditional inline formatting.

Instead, it should be two separate strings instead.

For instance:

```suggestion
                if (ourLicense.IsValid)
                {
                    this.Log().Info("We have found a valid license for Chocolatey {0}, which expires {1}", ourLicense.LicenseType, ourLicense.ExpirationDate);
                }
                else
                {
                    this.Log().Info("We have found an invalid license for Chocolatey {0}, which expires {1}", ourLicense.LicenseType, ourLicense.ExpirationDate);
                }
```

AdmiringWorm (changes requested):
I know this is a WIP. Just commenting on a few things I noticed.

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
JPRuskin: We've updated this to handle FOSS licenses!

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
```diff
@@ -0,0 +1,138 @@
...
+        {
+        }
+
+        public void run(ChocolateyConfiguration config)
+        {
+            var ourLicense = LicenseValidation.validate();
+            var nodeCount = parse_node_count(ourLicense.Name);
+            var logger = config.RegularOutput ? ChocolateyLoggers.Normal : ChocolateyLoggers.LogFileOnly;
+
+            if (ourLicense.LicenseType == ChocolateyLicenseType.Foss)
```
corbob, line 72: It looks like we can use `is_licensed_version()`: https://github.com/chocolatey/choco/blob/develop/src/chocolatey/infrastructure/licensing/ChocolateyLicense.cs#L36

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
```diff
@@ -0,0 +1,138 @@
...
+                {
+                    return int.Parse(licenseName.Substring(startIndex + 1, endIndex - 1));
+                }
+
+                // If we get this far, start testing at the end of the current range
+                startIndex = startIndex + endIndex;
+            }
+
+            return null;
+
```
JPRuskin, line 135: Gary wouldn't approve this.
```suggestion
```

JPRuskin: He also suggests using the regex used in other parsing parts, for consistency.

```
        private static readonly Regex _licenseCountRegex = new(@"\[.*?(?<licensedMachineCount>\d+).*?\]");
```

gep13:
@JPRuskin @corbob unless there are any objections, I am going to take the liberty of making some changes/updates to this PR.

gep13:
@JPRuskin @corbob I have finished doing an overhaul of this PR.  This changes that I have made include:

- rebased on head of develop, and changed to new way of doing things for Obsolete methods
- added unit tests for LicenseCommand
- added Pester testsed for `choco license` invocation
- added parsing of subcommands, with a default to use `choco license info` - this was after a conversation with Paul, where it was decided that `info` was the appropriate default, rather than say `list` or `get`.
- switched to using Regex for parsing allowed Node Count from license, as this is inkeeping with how it is done elsewhere in products
- added help output from `choco license -h`
- switched `DryRun` to call `Run`, since this is inkeeping with what `choco support` command currently does
- changed string.Format output of DateTime to use universally recognised format strings, also don't include time in the output, as this doesn't really help
- added concept of LicenseCommandType to ChocolateyConfiguration, which allows for future expansion of LicenseCommand with additional subcommands

Let me know if either of you have any questions.

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
gep13: This change, to switch to use a regex, has been implemented.

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
```diff
@@ -0,0 +1,212 @@
+﻿// Copyright © 2017 - 2025 Chocolatey Software, Inc
```
AdmiringWorm, line 1: Not done a full review of this class just yet, but wanted to ask first:

Is there any reason why this class doesn't inherit from the [ChocolateyCommandBase](https://github.com/chocolatey/choco/blob/develop/src/chocolatey/infrastructure.app/commands/ChocolateyCommandBase.cs) instead of re-implementing the entire help document?

src/chocolatey.tests/infrastructure.app/commands/ChocolateyLicenseCommandSpecs.cs
```diff
@@ -0,0 +1,265 @@
...
+                }
+
+                errored.Should().BeTrue();
+                error.Should().NotBeNull();
+                error.Should().BeOfType<ApplicationException>();
+                error.Message.Should().Contain("A single license command must be listed");
+            }
+
+            [Fact]
+            public void Should_accept_add_as_the_subcommand()
```
JPRuskin:
```suggestion
            public void Should_accept_info_as_the_subcommand()

```

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
```diff
@@ -0,0 +1,212 @@
...
+            if (ourLicense.LicenseType == ChocolateyLicenseType.Foss)
+            {
+                this.Log().Warn(logger, "No Commercial License Found, running with Open Source License.");
+                return;
+            }
+
+            if (!ourLicense.IsValid)
+            {
+                this.Log().Warn(logger, "We have found an invalid license for Chocolatey {0}: {1}", ourLicense.LicenseType, ourLicense.InvalidReason);
+                Environment.ExitCode = 1;
```
corbob: Should we be returning here?

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
corbob: @gep13 would it be accurate to say you resolved this as `IsLicensedVersion()` checks that it's a valid license and we're not only checking for valid license below, but if it's not a valid license we're outputting the reason. Assuming that's the case, should we also check for license type Unknown here?

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
corbob: If I were to guess based on the first commit for `ChocolateyCommandBase`, and how long this PR has been here: `ChocolateyCommandBase` didn't exist when this PR was initially created. It would probably benefit updating to derive from that class.

src/chocolatey.tests/infrastructure.app/commands/ChocolateyLicenseCommandSpecs.cs
gep13: Thanks for catching that!  I have fixed this up.

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
gep13: I have fixed this up to now make use of `ChocolateyCommandBase`. Thank you for the reminder about this!

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
gep13: Yes, I believe so.

src/chocolatey/infrastructure.app/commands/ChocolateyLicenseCommand.cs
gep13: I think that this is belt and braces, but there is no harm in adding this additional check.

gep13:
@AdmiringWorm this should be ready for a final review now.  I have fixed up the unit test that was failing as a result of the change that was made to the docs.

AdmiringWorm:
Great work on getting this added and updated @JPRuskin and @gep13. Just waiting on the builds to finish, then we can get this merged.

AdmiringWorm:
We'll ignore the docker build for now, it is exhibiting the same issues that happens from time to time when it comes with SSL connections, and since the ubuntu and windows build succeeded, we have the ones we want at the moment.

PR #2856, opened 2022-10-13
(#2591) Add headers to limit output commands

### LINKED ISSUE #2591: Add a headers row to the output from all appropriate Chocolatey CLI commands when using `--limit-output` option
When you use `--limit-output` or `-r` you can pipe the output to `ConvertFrom-Csv` and provide `-Delimiter '|'` to get a PowerShell object. In order to get a proper one, you need to also pass in `-Header`. It would be beneficial to output a header row for the user so they don't need this.

The output I envision for `choco list -lo -r` would change from:
```
chocolatey|0.12.1
```
to (something like):
```
PackageId|Version
chocolatey|0.12.1`
```

In order to make this change in a way to not break existing scripts/workflows, I would suggest adding a feature `alwaysIncludeHeaders` that defaults to disabled with the possibility of defaulting it to enabled in a future version, as well as introduce a `--include-headers` option for per command control.

corbob:
## Description Of Changes

When a user asks for limited output from Chocolatey, it is not uncommon to pipe that output to `ConvertFrom-String` or `ConvertFrom-Csv` and manually add headers to get back an object. This allows for getting a header row back so that the end user doesn't need to add their own headers and discern what they are.

Adds a new feature that allows the user to always display the headers if desired. This defaults to off so that we don't break any existing scripts.

## Motivation and Context

Reduce the amount of work end users need to do in order to consume limited output.

## Testing

Ran through the tests in the Vagrant test environment. This resulted in two failures that appear unrelated to my changes. I will investigate those further before marking this PR as ready.

**NOTE:** This will require Chocolatey Licensed Extension changes. 

## Change Types Made

## Related Issue

Fixes #2591 

## Change Checklist

### DIFF
--- a/src/chocolatey/StringResources.cs
+++ b/src/chocolatey/StringResources.cs
@@ -655,5 +655,15 @@ namespace chocolatey
             internal const string UnableToDowngrade = "A newer version of {0} (v{1}) is already installed.{2} Use --allow-downgrade or --force to attempt to install older versions.";
             internal const string DependencyFailedToInstall = "Failed to install {0} because a previous dependency failed.";
         }
+
+        public static class OptionDescriptions
+        {
+            public const string IncludeHeaders = "Include header names when --limit-output is used. Requires Chocolatey CLI 2.5.0+";
+        }
+
+        public static class Options
+        {
+            public const string IncludeHeaders = "include-headers";
+        }
     }
 }
--- a/src/chocolatey/infrastructure.app/commands/ChocolateyPinCommand.cs
+++ b/src/chocolatey/infrastructure.app/commands/ChocolateyPinCommand.cs
@@ -178,12 +178,22 @@ If you find other exit codes that we have not yet documented, please
             config.QuietOutput = quiet;
             config.Input = input;
 
+            // This is required, since we only want to show the header if there are any Pinned Packages
+            // to be shown, which will be when we are printing the first Pinned Package
+            var hasHeaderRowBeenOutput = false;
+
             foreach (var pkg in packages.OrEmpty())
             {
                 var pkgInfo = _packageInfoService.Get(pkg.PackageMetadata);
                 if (pkgInfo != null && pkgInfo.IsPinned)
                 {
-                    this.Log().Info(() => "{0}|{1}".FormatWith(pkgInfo.Package.Id, pkgInfo.Package.Version));
+                    if (!config.RegularOutput && config.IncludeHeaders && !hasHeaderRowBeenOutput)
+                    {
+                        OutputHelpers.LimitedOutput("Id", "Version");
+                        hasHeaderRowBeenOutput = true;
+                    }
+
+                    OutputHelpers.LimitedOutput(pkgInfo.Package.Id, pkgInfo.Package.Version.ToNormalizedStringChecked());
                 }
             }
         }
--- a/src/chocolatey/infrastructure.app/services/ChocolateyConfigSettingsService.cs
+++ b/src/chocolatey/infrastructure.app/services/ChocolateyConfigSettingsService.cs
@@ -63,6 +63,11 @@ namespace chocolatey.infrastructure.app.services
         public virtual IEnumerable<ChocolateySource> ListSources(ChocolateyConfiguration configuration)
         {
             var list = new List<ChocolateySource>();
+
+            // This is required, since we only want to show the header if there are any Source
+            // to be shown, which will be when we are printing the first Source
+            var hasHeaderRowBeenOutput = false;
+
             foreach (var source in ConfigFileSettings.Sources.OrEmpty().OrderBy(s => s.Id))
             {
                 if (SkipSource(source, configuration))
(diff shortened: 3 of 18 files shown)


corbob:
The tests added here failed on the pin output with a header due to it being overloaded I think... It might require some further investigation, and perhaps we should do that before merging this in?

For completeness this is the Test Kitchen output:

```
         Context Outputs headers for pin
           [+] Exits success (0) 4ms (2ms|2ms)
           [-] Displays appropriate header 5ms (4ms|1ms)
            Expected 'PackageID|Version', but got $null.
            at $ActualOutput | Should -Be $ExpectedHeaders, C:\Users\vagrant\AppData\Local\Temp\verifier\suites\chocolatey-licensed-tests\chocolatey-tests\features\Headers.Tests.ps1:129
            at <ScriptBlock>, C:\Users\vagrant\AppData\Local\Temp\verifier\suites\chocolatey-licensed-tests\chocolatey-tests\features\Headers.Tests.ps1:129
```

src/chocolatey/StringResources.cs
```diff
@@ -0,0 +1,36 @@
...
+    public static class StringResources
+    {
+        public static class OptionDescriptions
+        {
+            public const string DISPLAY_HEADERS = "Display headers - Display headers when limit-output is used. Requires <INSERT RELEASED VERSION HERE>";
+        }
+
+        public static class Options
+        {
+            public const string DISPLAY_HEADERS = "display-headers|use-headers";
```
pauby, line 33: Lets not use aliases at all if we don't need to. Let's use `--use-headers`.

JPRuskin: Would `add-headers` / `display-headers` be more accurate than `use`? Using the headers is something code further along the theoretical pipeline will do, not Choco (to me, at least).

corbob:
Converting this back to draft as it requires quite a bit of rework and validation.

gep13:
@corbob I may have taken this for a quick spin, and rebased on the head of the develop branch.  Just wanted to give you a heads up, so that you weren't surprised.

src/chocolatey/StringResources.cs
```diff
@@ -65,5 +65,15 @@ public static class ErrorMessages
             internal const string UnableToDowngrade = "A newer version of {0} (v{1}) is already installed.{2} Use --allow-downgrade or --force to attempt to install older versions.";
             internal const string DependencyFailedToInstall = "Failed to install {0} because a previous dependency failed.";
         }
+
+        public static class OptionDescriptions
+        {
+            public const string DISPLAY_HEADERS = "Display headers - Display headers when limit-output is used. Requires 2.3.0";
```
pauby:
```suggestion
            public const string DISPLAY_HEADERS = "Output header - Output header when --limit-output is used. Requires 2.5.0";

```

src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
```diff
@@ -846,6 +854,13 @@ public virtual void Outdated(ChocolateyConfiguration config)
  Output is package name | current version | available version | pinned?
```
gep13:
```suggestion
 Output is package id | current version | available version | pinned?
```

Also, capital letters should perhaps be in place here.

tests/pester-tests/features/Headers.Tests.ps1
```diff
@@ -0,0 +1,133 @@
...
+            $ActualOutput = if ($Output.Lines.Count -gt 1) {
+                $Output.Lines[0]
+            }
+            else {
+                $Output.Lines
+            }
+
+            $ActualOutput | Should -Be $ExpectedHeaders
+        }
+
```
gep13: Can you remove this hideous, unnecessary line, pretty please!

tests/pester-tests/features/Headers.Tests.ps1
```diff
@@ -0,0 +1,133 @@
...
+            $ActualOutput = if ($Output.Lines.Count -gt 1) {
+                $Output.Lines[0]
+            }
+            else {
+                $Output.Lines
+            }
+
+            $ActualOutput | Should -Not -Be $ExpectedHeaders
+        }
+
```
gep13: Doh!

gep13:
@corbob I have taken the liberty of rebasing this PR onto the head of the develop branch.

src/chocolatey/StringResources.cs
gep13: Had a chat with @pauby about this, and we have settled on `--include-headers` which ties up with the existing `--include-version` option which is used with the `choco export` command to include version numbers in the created packages.config file.

src/chocolatey/StringResources.cs
gep13: This is being updated to:

> Include header names when --limit-output is used. Requires Chocolatey CLI 2.5.0+

gep13:
I have a number of local changes on my machine that I will push up here when I get a chance.  This includes the outcomes from a chat with @pauby and a few additional changes after doing some testing of things.

gep13:
During a discussion with @pauby the decision was made to:

* name the common option to `--include-headers`
* name the feature `alwaysIncludeHeaders`
* only show the header row if, and only if, there is something to be output (this is consistent with what the likes of PowerShell does)
* to be consistent with the names of the column headers for the new row, and have them match what the "things" are in for example the chocolatey.config file, or how they are named in the various commands that they are associated with.  The header rows are best "seen" in the Header.Tests.ps1 file, where they are all asserted at the top of the file.

gep13:
@AdmiringWorm I have run the Test-Kitchen tests for this PR, and they have all passed.

gep13:
@AdmiringWorm this is now ready for a review.  Thanks

src/chocolatey/infrastructure.app/commands/ChocolateyPinCommand.cs
```diff
@@ -178,12 +178,21 @@ public virtual void ListPins(ChocolateyConfiguration config)
             config.QuietOutput = quiet;
             config.Input = input;
 
+            // This is required, since we only want to show the header if there are any Pinned Packages
+            // to be shown, which will be when we are printing the first Pinned Package
+            var hasHeaderRowBeenOutput = false;
```
AdmiringWorm, line 183: Can you add a newline added between this argument and the foreach loop.

src/chocolatey/infrastructure.app/services/ChocolateyConfigSettingsService.cs
```diff
@@ -63,6 +63,10 @@ public virtual bool SkipSource(ConfigFileSourceSetting source, ChocolateyConfigu
         public virtual IEnumerable<ChocolateySource> ListSources(ChocolateyConfiguration configuration)
         {
             var list = new List<ChocolateySource>();
+
+            // This is required, since we only want to show the header if there are any Source
+            // to be shown, which will be when we are printing the first Source
+            var hasHeaderRowBeenOutput = false;
```
AdmiringWorm, line 69: Can you please add a newline between the argument and the foreach loop?

src/chocolatey/infrastructure.app/services/ChocolateyConfigSettingsService.cs
```diff
@@ -75,29 +79,35 @@ public virtual IEnumerable<ChocolateySource> ListSources(ChocolateyConfiguration
...
-                        source.Value,
-                        source.Disabled.ToStringSafe(),
-                        source.UserName.QuoteIfContainsPipe(),
-                        source.Certificate,
-                        source.Priority,
-                        source.BypassProxy.ToStringSafe(),
-                        source.AllowSelfService.ToStringSafe(),
-                        source.VisibleToAdminsOnly.ToStringSafe()
-                        ));
+                        if (configuration.IncludeHeaders && !hasHeaderRowBeenOutput)
```
AdmiringWorm, line 95: Any reason why this is a standalone if statement, instead of being combine with the else statement as `else if`?

src/chocolatey/infrastructure.app/services/TemplateService.cs
```diff
@@ -348,7 +351,13 @@ protected void ListCustomTemplateInformation(ChocolateyConfiguration configurati
             }
             else
             {
-                this.Log().Info("{0}|{1}".FormatWith(configuration.TemplateCommand.Name, pkgVersion));
+                if (configuration.IncludeHeaders && !_hasHeaderRowBeenOutput)
```
AdmiringWorm, line 354: Same question as before. Any reason why this hasn't been combined as a `else if` statement?

AdmiringWorm (changes requested):
Some minor nitpicks.

src/chocolatey/infrastructure.app/services/ChocolateyConfigSettingsService.cs
gep13: This needs to be here, since the `else`above is for outputing the actual source, whereas the new `if` statement is for the header row, if the header row is required.

src/chocolatey/infrastructure.app/services/TemplateService.cs
gep13: Same comment as before, this is needed to only output header row when required.

AdmiringWorm (approved):
LGTM

PR #2868, opened 2022-10-25
(#2867) Add --online option to all commands

### LINKED ISSUE #2867: Provide ability to navigate directly to online version of help documentation for a command

I believe that there is a lot of output from the `choco -h` command, and I would like to be able to navigate to the online version of the help documentation, similar to what can be done from PowerShell, for example:

```
Get-Help Get-Command -Online
```

This would make it much easier to get the information I need by reading the online version of the documentation, which is also likely to have the most up to date version of the docs, directly from the command line, without the need to navigate to the docs site.

## Additional Context.

N/A

## Related Issues

N/A

gep13:
## Description Of Changes

This commit adds the --online option to all Chocolatey commands, to allow opening of the online help documentation for the command that is being executed.

For example, running:

choco install -h

will output the help documentation to the console, but running:

choco install -h --online

will not output anything to the command line, and instead open the default browser to the help documentation for the command.

## Motivation and Context

I want to improve the experience of reading the help documentation, by providing a --online option to open in default browser, similar to what is done by the `Get-Help` cmdlet in PowerShell.

## Testing

1. Compile code
2. Set debug options to be `install -h --online`
3. Run choco.exe and see that default browser opens to help page for install command
4. Run `install -h` and see that help is written to output
5. Run `bob -h --online` and see that command is not found

## Change Types Made

## Related Issue

Fixes #2867

## Change Checklist


src/chocolatey/infrastructure.app/configuration/ConfigurationOptions.cs
```diff
@@ -120,6 +120,12 @@ public static OptionSet OptionSet
 
             if (configuration.HelpRequested)
             {
+                if (configuration.ShowOnlineHelp)
+                {
+                    System.Diagnostics.Process.Start("explorer.exe", string.Format("https://docs.chocolatey.org/en-us/choco/commands/{0}", configuration.CommandName));
```
TheCakeIsNaOH, line 125: This will not work on non-Windows systems. It could just be something that is Windows only, and just shows a warning/error, or `xdg-open` could be investigated.

TheCakeIsNaOH (changes requested):

src/chocolatey/infrastructure.app/configuration/ConfigurationOptions.cs
gep13: That is a very good point. This was done on the Hacktoberfest stream as a _very_ quick proof of concept, and why the PR was created in draft format :smile:

gep13: @TheCakeIsNaOH I have fixed this up, and this should now work across OSes. Let me know if you have any questions.

src/chocolatey/infrastructure.app/configuration/ConfigurationOptions.cs
```diff
@@ -75,8 +75,11 @@ public static OptionSet OptionSet
             {
                 _optionSet
                     .Add("?|help|h",
-                         "Prints out the help menu.",
-                         option => configuration.HelpRequested = option != null);
+                        "Prints out the help menu.",
+                        option => configuration.HelpRequested = option != null)
+                    .Add("online",
+                        "Online - Open help for specified command in default browser application.",
```
AdmiringWorm, line 81: Should there be a mention about `--online` can only be used when `--help` or `-h` is used?

src/chocolatey/infrastructure.app/configuration/ConfigurationOptions.cs
```diff
@@ -120,6 +123,12 @@ public static OptionSet OptionSet
 
             if (configuration.HelpRequested)
             {
+                if (configuration.ShowOnlineHelp)
+                {
+                    System.Diagnostics.Process.Start(string.Format("https://docs.chocolatey.org/en-us/choco/commands/{0}", configuration.CommandName));
```
AdmiringWorm, line 128: Just relying on the command name could be a problem (I think), unless we also redirect aliased command names to the correct page on the docs site

TheCakeIsNaOH: It may make sense to create ch0.co links? That way if the docs site changes, the link still can work.

src/chocolatey/infrastructure.app/configuration/ConfigurationOptions.cs
gep13: That is a good suggestion, I will get that added.

src/chocolatey/infrastructure.app/configuration/ConfigurationOptions.cs
gep13: @AdmiringWorm we have pages in place for all the command aliases, you can see them listed here:

* https://github.com/chocolatey/docs/tree/master/input/en-us/choco/commands
* https://github.com/chocolatey/docs/tree/master/input/en-us/create/commands

So I think we are covered there.

@TheCakeIsNaOH I actually had the same thought, but I wasn't sure if it was warranted here or not.  Happy to add them though.  What would be the suggestion here?  Something like:

* https://ch0.co/c/<command name>

src/chocolatey/infrastructure.app/configuration/ConfigurationOptions.cs
corbob: Would it make more sense to have `--online` imply `--help` and just make it clear in the output that it will open the help for the command online?

gep13: No, I don't think so.

This addition, is the first of a few things that I would like to do in this area.  For example, I would like to be able to do:

```
choco install -h --examples
```

and

```
choco install -h --detailed
```

etc.  So I don't think having it imply the help option is a good idea.

src/chocolatey/infrastructure.app/configuration/ConfigurationOptions.cs
TheCakeIsNaOH: @gep13 that formatting/syntax for the short link looks good,

AdmiringWorm, line 128: Not 100% sure, but if I remember correctly, this won't work on all/most Linux distros (possibly not OSX as well).

It may be just for older variants, though. @TheCakeIsNaOH are you able to confirm or deny whether just starting a URL works on the distros that you are aware of?

AdmiringWorm:
Should we warn the user if they are using `--online` but haven't requested help being shown?

src/chocolatey/infrastructure.app/configuration/ConfigurationOptions.cs
TheCakeIsNaOH: I just tested this on a Debian system with a GUI, and it is working for me, but it did not work on a headless system.

TheCakeIsNaOH:
@gep13 from my testing it looks like it does not work on headless/no gui systems. That is why the Ubuntu build is failing.

gep13:
I should also note that I have updated this PR to use the following URL format:

```
https://ch0.co/c/<commandName>
```

All of these short URL's have been created, ready to be used.

src/chocolatey/infrastructure.app/configuration/ConfigurationOptions.cs
```diff
@@ -120,10 +125,33 @@ public static OptionSet OptionSet
...
                 ShowHelp(_optionSet, helpMessage);
             }
             else
             {
+                // Only show this warning once
+                if (!_helpWarningHasBeenShown && configuration.ShowOnlineHelp)
+                {
+                    "chocolatey".Log().Warn("The --online option has been used, without the corresponding -?/--help/-h option.  Command execution will be completed without invoking help.");
+                    _helpWarningHasBeenShown = true;
+                }
```
vexx32, line 153: Something about this check isn't working as expected. It's showing for every invocation, even when I _do_ have the `--help` or `-h` option provided. 

Does this code path get hit multiple times for some reason, and one of the times it doesn't have `-h` set at that point?

vexx32 (changes requested):
Something's not working right with the warning, but other than that this seems to work great :3

src/chocolatey/infrastructure.app/configuration/ConfigurationOptions.cs
gep13: I have taken a stab at fixing this, can you please review this again? Thanks!

src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
```diff
@@ -448,8 +448,11 @@ private static void SetGlobalOptions(IList<string> args, ChocolateyConfiguration
                 {
                     if (!string.IsNullOrWhiteSpace(config.CommandName))
                     {
-                        // save help for next menu
+                        // The setOptions lambda above is called twice, once to set the command name, and then to execute the command with all options.
+                        // To ensure correct option of the help options, we need to reset them to false in the first execution, to then have them parsed correctly in the second
```
vexx32, line 452:
```suggestion
                        // To ensure correct operation of the help options, we need to reset them to false in the first execution, to then have them parsed correctly in the second

```

vexx32 (approved):
This mostly looks good to me, nice work! There's a small typo we should maybe fix before merging, but everything else looks good :3

vexx32:
@gep13 I've updated the comment with a slight adjustment to wording and fixing the typo. Thanks for taking care of this! 💖

PR #2900, opened 2022-11-11
(#2781) Add chocolateyPackageId to env variables

### LINKED ISSUE #2781: `$env:ChocolateyPackageName`: for consistency, it should better be named `$env:ChocolateyPackageId`
Yes, it is a problem because it creates confusion.
The [Docs: Environment Variables](https://docs.chocolatey.org/en-us/create/functions/) explains:
* _ChocolateyPackage**Title**: … equivalent to the <**title** /> field in the nuspec_
* _ChocolateyPackage**Version**: … equivalent to the <**version** /> field in the nuspec_

**This rule is great:** The $Env-Variable is named according to the field in nuspec 😃, therefore we must not ask google for the $env variable name for the nuspec Title / Version field. It is intuitive.

But now, Choco breaks the rule, the Docs explains:
* _ChocolateyPackage**Name**: …  equivalent to the <**id** /> field in the nuspec_

It would be very helpful if in the next version of Chocolatey:
- `$env:ChocolateyPackageName` will be renamed to `$env:ChocolateyPackageId`
- Then, `$env:ChocolateyPackageName` gets an Alias for `$env:ChocolateyPackageId` which will disappear in a later version

I hope that I have not to explain why this is useful 😃 

## Additional Context.
N/A

## Related Issues
N/A

steviecoaster:
## Description Of Changes

Adds a new `$env:ChocolateyPackageId` environment variable to the PowerShell host when installing/upgrading/uninstalling a Chocolatey package.

## Motivation and Context

The nuspec references a package by id, making `$env:ChocolateyPackageName` less idiomatic than `$env:ChocolateyPackageId`. The `$env:ChocolateyPackageName` variable is still available for use, to prevent existing packages from breaking, but moving forward this new env variable is more inline with the nuspec definition of id.

## Testing

1. Create sample package that emits $env:ChocolateyPackageId
2. Verifiy package id is output when installing package

## Change Types Made

## Related Issue

Fixes #2781

## Change Checklist

### DIFF
--- a/src/chocolatey/infrastructure.app/services/PowershellService.cs
+++ b/src/chocolatey/infrastructure.app/services/PowershellService.cs
@@ -435,6 +435,8 @@ namespace chocolatey.infrastructure.app.services
 
             Environment.SetEnvironmentVariable("chocolateyPackageName", package.Id);
             Environment.SetEnvironmentVariable("packageName", package.Id);
+            Environment.SetEnvironmentVariable("chocolateyPackageId", package.Id);
+            Environment.SetEnvironmentVariable("packageId", package.Id);
             Environment.SetEnvironmentVariable("chocolateyPackageTitle", package.Title);
             Environment.SetEnvironmentVariable("packageTitle", package.Title);
             Environment.SetEnvironmentVariable("chocolateyPackageVersion", package.Version.to_string());


src/chocolatey/infrastructure.app/services/PowershellService.cs
```diff
@@ -435,6 +435,8 @@ public void prepare_powershell_environment(IPackage package, ChocolateyConfigura
 
             Environment.SetEnvironmentVariable("chocolateyPackageName", package.Id);
             Environment.SetEnvironmentVariable("packageName", package.Id);
+            Environment.SetEnvironmentVariable("chocolateyPackageId", package.Id);
+            Environment.SetEnvironmentVariable("packageId", package.Id);
```
pauby, line 439: We should deprecate the non-`chocolatey` prefix environment variable names. So I don't think we should add this one in at all.

gep13: Agreed.  The deprecation of the previous environment variables is not something that we are tackling in this PR, but rather than make the problem worse, let's simply not introduce the `packageId` environment variable in this PR.

src/chocolatey/StringResources.cs
```diff
@@ -82,6 +82,12 @@ public static class Package
                 /// <remarks>Business Edition Only Variable</remarks>
                 public const string ChocolateyLicenseType = nameof(ChocolateyLicenseType);
 
+                /// <summary>
+                /// The name or identifier of the package that is being handled.
+                /// </summary>
+                /// <seealso cref="ChocolateyPackageName" />
```
AdmiringWorm, line 94:
```suggestion
                /// <seealso cref="ChocolateyPackageName" />
                /// <remarks>Available in v2.5.0+</remarks>
```

Would be a good idea to denote when it was added, so it flow through to our internal documentation.

AdmiringWorm (changes requested):
Just a single recommendation, otherwise this looks good to me

gep13:
@AdmiringWorm that is a good idea!  I have got that updated just now.

AdmiringWorm (approved):
LGTM

AdmiringWorm:
thanks for adding this @steviecoaster and @gep13. Great work 👍

PR #2907, opened 2022-11-16
(#2738) Add net48 TFM folder in NuGet package 

### LINKED ISSUE #2738: Uplift to .NET Framework 4.8
Since older .Net framework versions are going out of support, and because newer versions of the NuGet libraries require newer .Net versions (#508), Chocolatey CLI should be upgraded to be built on .Net 4.8

This is also a stepping stone to upgrading to dotnet 5+, #2436 

This will also allow upgrading the SimpleInjector version, which likely would be helpful for #2572 and for #2693

- [x] As recently linked, uplifting NUnit to v3.x
- [x] ~~Updating the initial install.ps1 script to check for and install .net 4.8 if required~~ Only adding check to `chocolateySetup.psm1`
- [x] Updating the Chocolatey CLI package install/upgrade process to validate that .net 4.8 is installed. https://github.com/chocolatey/choco/pull/2933
- [ ] Updating documentation on the docs site, and the main site for Chocolatey CLI requirements.
  - https://github.com/chocolatey/docs/issues/638
  - https://github.com/chocolatey/chocolatey.org/issues/228
- [ ] Validating that all currently supported operating systems work with .net 4.8. I think they all do, but I'm not sure if there is a windows server version or something that is not compatible. 
  - https://github.com/chocolatey/choco/issues/2992 
  - It bumps the requirement to 2008 R2 https://github.com/chocolatey/choco/issues/2903
- [x] Add a check into chocolatey.console that .net 4.8 is installed. https://github.com/chocolatey/choco/pull/2935
- [ ] Investigate if there are any other dependency libraries that can/should be uplifted or migrated to an alternative. I think this is something that does not need to happen before .net 4.8 version of choco comes out. https://github.com/chocolatey/choco/issues/2896
- [ ] Update TLS handling https://github.com/chocolatey/choco/issues/2902 https://github.com/chocolatey/choco/issues/2585

gep13:
## Description Of Changes

Now that we are building specifically for .NET 4.8, we should add the
generated chocolatey.dll file to the correct TFM folder within the
NuGet package, so that it is correctly located, and extracted when
consuming the NuGet package.

The Credits.txt file has been left at the root level of the lib folder,
rather than replicating into each TFM folder.

## Motivation and Context

This makes it clear what framework the assembly is targetting.

## Testing

Ran build.bat locally, and then verified that the net48 folder existed in the generated nupkg file. This was done using NuGet Package Explorer.

## Change Types Made

## Related Issue

Relates to #2738 

## Change Checklist

### DIFF
--- a/recipe.cake
+++ b/recipe.cake
@@ -289,12 +289,14 @@ Task("Prepare-NuGet-Packages")
     .IsDependeeOf("Sign-Assemblies")
     .Does(() =>
 {
-    // Copy legal documents
     CleanDirectory(BuildParameters.Paths.Directories.NuGetNuspecDirectory + "/lib");
+    EnsureDirectoryExists(BuildParameters.Paths.Directories.NuGetNuspecDirectory + "/lib/net48");
+
+    // Copy legal documents
     CopyFile(BuildParameters.RootDirectoryPath + "/docs/legal/CREDITS.md", BuildParameters.Paths.Directories.NuGetNuspecDirectory + "/lib/CREDITS.txt");
 
-    CopyFiles(BuildParameters.Paths.Directories.PublishedLibraries + "/chocolatey_merged/*", BuildParameters.Paths.Directories.NuGetNuspecDirectory + "/lib");
-    CopyFile(BuildParameters.Paths.Directories.PublishedLibraries + "/chocolatey/chocolatey.xml", BuildParameters.Paths.Directories.NuGetNuspecDirectory + "/lib/chocolatey.xml");
+    CopyFiles(BuildParameters.Paths.Directories.PublishedLibraries + "/chocolatey_merged/*", BuildParameters.Paths.Directories.NuGetNuspecDirectory + "/lib/net48");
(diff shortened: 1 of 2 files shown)


recipe.cake
```diff
@@ -291,11 +291,11 @@ Task("Prepare-NuGet-Packages")
     .Does(() =>
 {
     // Copy legal documents
-    CleanDirectory(BuildParameters.Paths.Directories.NuGetNuspecDirectory + "/lib");
+    CleanDirectory(BuildParameters.Paths.Directories.NuGetNuspecDirectory + "/lib/net48");
```
AdmiringWorm, line 294: Any reason why you changed this to only clean the .NET 4.8 directory?

Shouldn't this clean out everything?

gep13: There is no real reason, no. The CleanDirectory here is/was really being used to ensure that the directory existed, since the nuspec directory is cleaned out elsewhere during the build.

I have no issue with switching to the CleanDIrectories alias, and doing both lib and lib/net48.

Assuming that this is what you are suggesting.

gep13: I have went with a combination of the two approaches here. Let's ensure that the `lib` directory is cleaned, and then ensure that the `net48` folder exists.

PR #2919, opened 2022-11-28
(#2865) Avoid re-loading chocolatey.dll

### LINKED ISSUE #2865: Getting Cannot Combine Path Error
### What You Are Seeing?
When attempting to create a package running a `choco new --file "C:\temp\WIP\vlc.exe" --build-package --outputdirectory "C:\temp\WIP"` I am receiving the following error message. Cannot combine a path with ':' attempted to combine 'C:\temp\WIP\vlc.exe' with 'C:\temp\WIP'

### What is Expected?
The new package should be created an placed on C:\temp\WIP

1. Have Chocolatey V 1.2.0 and chocolatey.extension V 5.0.0 installed 
1. Download the VLC exe and rename it vlc.exe
1. Place that vlc.exe in C:\temp\WIP
1. run the following command: `choco new --file "C:\temp\WIP\vlc.exe" --build-package --outputdirectory "C:\temp\WIP"`

### System Details
* OS Build: Windows Server 2022
* Chocolatey version: 1.2.0

Attempted running this with chocolatey V 1.1.0 and chocolatey.extension V 4.2.0 with no error.

### Output Log
<details>
<summary>Relevant chocolatey log output</summary>

<p>

~~~sh
2022-10-21 12:58:19,830 7200 [ERROR] - Cannot combine a path with ':' attempted to combine 'C:\temp\WIP\vlc.exe' with 'C:\temp\WIP'
2022-10-21 12:58:19,834 7200 [ERROR] - More Details: System.ApplicationException: Cannot combine a path with ':' attempted to combine 'C:\temp\WIP\vlc.exe' with 'C:\temp\WIP'
   at chocolatey.infrastructure.filesystem.DotNetFileSystem.combine_paths(String leftItem, String[] rightItems)
   at chocolatey.infrastructure.app.services.TemplateService.generate(ChocolateyConfiguration configuration)
   at chocolatey.infrastructure.app.runners.GenericRunner.run(ChocolateyConfiguration config, Container container, Boolean isConsole, Action`1 parseArgs)
   at chocolatey.infrastructure.app.runners.ConsoleApplication.run(String[] args, ChocolateyConfiguration config, Container container)
   at chocolatey.console.Program.Main(String[] args)
2022-10-21 12:58:19,835 7200 [DEBUG] - Exiting with 1
~~~

</p>

</details>

### Related Info.
- [ZenDesk Ticket](https://chocolatey.zendesk.com/agent/tickets/15670)

vexx32:
## Description Of Changes

- Rewrote assembly resolver to prevent loading `chocolatey.dll` in extension contexts and ensure it's only returned for ILMerged assemblies.
- Little bit of a refactor to the assembly resolver so we don't need to check the same condition like 3 or 4 times.

## Motivation and Context

Previously there was an issue where in certain cases, chocolatey.dll could be re-loaded as an extension after the chocolatey licensed extension commands are loaded, overwriting the commands already loaded.

This could be triggered by having other dlls in the extensions folder that are not stamped with our public key, such as
chocolateygui-extension. This fix ensures that we are both not loading specifically chocolatey.dll via our assembly resolver, and also that extensions not stamped with our public key don't cause the main Chocolatey assembly to be loaded as an extension dll.

ILMerged assemblies are still loaded by returning a reference to the chocolatey assembly.

## Testing

1. Build the project.
2. Install both chocolatey.extension.dll and chocolateygui.extension.dll

### Operating Systems Testing

- Windows 11

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

Fixes #2865


vexx32:
Adding Kim as an additional reviewer here as I had to rewrite a bunch of this to make sense of it 😁

src/chocolatey.console/Program.cs
```diff
@@ -238,20 +238,45 @@ private static void add_assembly_resolver()
 #else
                 var chocolateyPublicKey = ApplicationParameters.UnofficialChocolateyPublicKey;
 #endif
+                if (!requestedAssembly.get_public_key_token().is_equal_to(chocolateyPublicKey))
+                {
+                    // This resolver is only loading official extensions for Chocolatey and ILMerged assemblies.
+                    // Everything else is not handled by this resolver.
+                    return null;
```
corbob, line 245: It's not this line in particular, but the return is confusing when looking at the diff (see screenshot below). As the diff shows we're inside of void method `add_assembly_resolver()` and yet we are returning things. I had to expand the view to see that we are in a delegate definition. Would it be an idea to move this delegate definition into its own method and pass the method reference to the delegate handler?



I'm not sure if such a refactor necessarily belongs in a bugfix release though, so perhaps if we decide we want to refactor like that we can open an issue and address it in a future release.

corbob (commented):
Just a question about the rather large delegate and if we should refactor it. I'll take it for a spin in my VM, and we can discuss the potential refactor tomorrow.

src/chocolatey.console/Program.cs
vexx32: Yeah I'm all for cleaning this up some more. Unsure if it's something we want to do at this point, but happy to discuss. It's a relatively simple change to make, just moving some code around a little more.

/cc @admiringworm

corbob:
I haven't dug into the code, but there is now a mildly helpful error (although not really helpful since the only way to resolve it is to remove the dll).

image

vexx32:
Hmm. Well it's not _wrong_ necessarily but it likely shouldn't be erroring there. Will have a look, didn't see that when debugging.

src/chocolatey.console/Program.cs
AdmiringWorm: With the current codebase, I am fine with short-circuiting if the requested assembly isn't officially signed.
But do note that this will require more changes needed when/if we ever decide in the future to open up (or partially open) creating extensions by non-officially signed binaries.

src/chocolatey.console/Program.cs
```diff
@@ -238,20 +238,45 @@ private static void add_assembly_resolver()
...
+                            ApplicationParameters.LicensedChocolateyAssemblySimpleName,
+                            ApplicationParameters.LicensedAssemblyLocation,
+                            requestedAssembly.get_public_key_token()).UnderlyingType;
                     }
+                    catch (Exception ex)
+                    {
+                        "chocolatey".Log().Warn("Unable to load chocolatey.licensed assembly. {0}".format_with(ex.Message));
+                        return null;
+                    }
+                }
```
AdmiringWorm, line 272: This is a breaking change compared to the dynamic loading introduced in v1.2.0.

You are re-introducing a hard-coded path for the licensed chocolatey assembly, which was one of the introductions of the dynamic loading tried to prevent the need.

This must happen after loading extensions or continuing the check for the assembly if none is found in the hard-coded path.

src/chocolatey.console/Program.cs
```diff
@@ -238,20 +238,45 @@ private static void add_assembly_resolver()
...
 
-                    if (Directory.Exists(ApplicationParameters.ExtensionsLocation))
+                if (Directory.Exists(ApplicationParameters.ExtensionsLocation))
+                {
+                    var extensions = Directory.EnumerateFiles(ApplicationParameters.ExtensionsLocation, requestedAssembly.Name + ".dll", SearchOption.AllDirectories).ToList();
```
AdmiringWorm, line 276: Don't use `.ToList()` here. We rely on lazy enumeration in case there are/will be multiple dll files in the future.

Doing `ToList` adds all assemblies into memory (the string at least), while not doing it will load them one by one.

AdmiringWorm (changes requested):
A few things to note, additionally there was a reason a refactor wasn't made but rather additions to the existing resolver.

To refactor it, we need to prove out the existing logic compared to what it is being refactored into, as I couldn't find an easy way to prove this out reliably that was skipped.

So bottom line, we need to prove out how everything is resolved before refactoring it, to prevent a breaking change so that should probably not happen in a bug fix release.

Also, you need to update the assembly resolver addition in `GetChocolatey` as well: https://github.com/chocolatey/choco/blob/develop/src/chocolatey/GetChocolatey.cs#L74

src/chocolatey.console/Program.cs
vexx32: Gotcha. The reason I kept this was I noticed that the LicensedAssemblyLocation actually checks two different locations, one of which matches the usual path in /extensions, the other just looks in the same dir as the chocolatey executable.

I'll remove the return here so it can continue into the "normal" dynamic loading.

src/chocolatey.console/Program.cs
vexx32: The ToList() is needed so we can check the Count before enumerating. This is done so that if we know we found something here that looks like it should be an extension but it can't be loaded, we ensure that we don't continue to the fallback which is to return the `chocolatey` assembly itself. 

That behaviour _was_ the bug this is fixing. This is necessary to fix the bug.

vexx32:
In terms of proving out its behaviour, it's relatively straightforward to add breakpoints into the resolver and follow what path each resolution makes as Chocolatey runs. With the code here, it's pretty clear when debugging that:

1. The licensed extension loads successfully
2. Other chocolatey extensions _can_ be loaded if (and only if) they have the correct strong name signing key, otherwise the resolution attempt returns null and will not load them (and, crucially, will not accidentally load the `chocolatey` assembly itself as an extension, which is what caused #2865)
3. ILMerged assemblies can be resolved to the `chocolatey` assembly.

src/chocolatey.console/Program.cs
vexx32: I opted to move this to after the dynamic check instead, retaining the return. That way in most cases the dynamic check should take care of it, and it can be a safe fallback that can return null if it also can't load the licensed extension.

src/chocolatey.console/Program.cs
AdmiringWorm: I am aware, which is why had suggested using a boolean value to determine whether we had gone into the foreach loop, instead of calling `ToList` and checking its Count value.

Using a boolean value to determine the need has a lower overhead than having to call `ToList` every time.

Bottom line, if going down this route, change the `EnumerateFiles` to `GetFiles` instead, which returns a pre-allocated array that can be used (I would suggest going with a for loop instead of a foreach loop in that case though for less overhead).

AdmiringWorm:
> In terms of proving out its behaviour, it's relatively straightforward to add breakpoints into the resolver and follow what path each resolution makes as Chocolatey runs. With the code here, it's pretty clear when debugging that:
> 
> 1. The licensed extension loads successfully
> 2. Other chocolatey extensions _can_ be loaded if (and only if) they have the correct strong name signing key, otherwise the resolution attempt returns null and will not load them (and, crucially, will not accidentally load the `chocolatey` assembly itself as an extension, which is what caused [Getting Cannot Combine Path Error #2865](https://github.com/chocolatey/choco/issues/2865))
> 3. ILMerged assemblies can be resolved to the `chocolatey` assembly.

I should have been clearer, when saying proving it out I am talking about proving it out with tests, not as part of a debugging session.
Either unit/integration tests, or automated/manual e2e tests.

Integration tests would be preferable but would require a bit of refactoring to happen first.

src/chocolatey.console/Program.cs
AdmiringWorm: > Gotcha. The reason I kept this was I noticed that the LicensedAssemblyLocation actually checks two different locations, one of which matches the usual path in /extensions, the other just looks in the same dir as the chocolatey executable.

Not sure why it checks those two places. If the assembly is located next to the `choco` executable (or in the same directory at least), we wouldn't actually have any control of its loading as it will be loaded automatically by the default assembly loader in almost all cases (not overridable).

vexx32:
I don't think unit tests are likely to be particularly useful here. Integration tests where we load the chocolatey licensed extension already exist, and if this broke anything _all_ of those would very quickly break.

Perhaps we should add a test verifying that installing the chocolatey GUI licensed extension no longer breaks things, but that's the only test I could see worth adding. Do you have something else in mind?

src/chocolatey.console/Program.cs
vexx32: That's a fair point actually. So this truly seems like it's redundant and not doing anything that the other path isn't already doing, then.

src/chocolatey.console/Program.cs
vexx32: Yeah, using a bool is a good idea, thanks. I'll go with that I think.

vexx32:
@AdmiringWorm I've added a manual test to our internal testing documentation for now to verify this works properly. As we discussed, it doesn't seem feasible right this minute to test without the official licensed extension installed, so it's not something we can readily test with just OSS components.

src/chocolatey.console/Program.cs
vexx32: @AdmiringWorm it sounds like we're in agreement here, but just to confirm -- I think we should remove this alternate code path for the licensed extension and let it load with the more generic code, yeah?

AdmiringWorm: Yeah, I believe we are and that it can be removed.

Unless I have missed something, I believe we will never get into this alternative code path because of the dynamic loading.

AdmiringWorm (approved):
LGTM, I have ran through a few tests with the code for this. Both with current licensed extension, old license extension and without license extension and with a stray DLL, and without one.

I'll leave this open in case you wanted to get that alternative path removed, and to let @corbob have another look over.

AdmiringWorm (changes requested):
Disregard my previous comment, I hadn't installed chocolateygui.extension.

I get the following non-terminating error:

image

We may want to do special handling for chocolateygui.licensed so this isn't outputted.
this would need to go into https://github.com/chocolatey/choco/blob/develop/src/chocolatey/FileSystemExtensions.cs#L31

THere is also my previous comment about the logic in Program.cs needs to be duplicated into GetChocolatey.cs as well.

vexx32:
I've updated the commit to do the following:

1. Remove the agreed redundant logic for loading specifically the licensed extension and let it just load via the normal path
2. Move the assembly resolver to a dedicated method
3. Reference the resolver from both chocolatey.console.Program and Lets.GetChocolatey

Perhaps more of a refactor than was strictly necessary, but given we intend to duplicate the changes to Lets.GetChocolatey anyway I felt it was warranted to just de-dupe the unnecessarily duplicated code, since the intention is for both of them to behave the same.

vexx32:
@gep13 @AdmiringWorm OK so... the only reason this _didn't_ error previously was because the unloadable DLL was considered valid and loaded by Chocolatey by returning a reference to the main `chocolatey` assembly.

What is the desired behaviour here going forward? The code path seems to be designed to _warn_ if unloadable DLLs are present in the extensions directory, but given this location is now also used for the ChocolateyGUI licensed extension which Chocolatey CLI should not be loading itself, we may need to rethink that stance.

I _think_ the best solution is:

1. If the DLL is rejected because it does not have the requested public key, we ignore it entirely. No warning, nothing. Just ignored. (Maybe we add a "warn if the extension name matches <list of expected official extension names>" as well? unsure)
2. If the DLL has the correct public key and still doesn't load for some reason, we error that out as is the current behaviour.

Because currently if I just catch the FileLoadException that occurs when we try to load an assembly with the wrong public key, we get a _warning_ that the extension can't be loaded. This was the original intention, but due to this same bug this warning was _never able to be shown_.

As a result I think the "expected behaviour" has become for choco to completely ignore those unloadable extension dlls instead of warning. Perhaps we demote the 'warning' to a debug log message instead, so they're there if we need them but users don't see them in console output by default.

vexx32:
@corbob @AdmiringWorm @gep13 I chatted to Gary and we agreed that the existing behaviour where Chocolatey effectively _ignores_ DLLs in the /extensions directory should be maintained. To that end, we are catching most errors that come from Assembly.Load() when loading extensions now that the resolver no longer resolves them to the `chocolatey` assembly.

This then leads to the other code path which previously would have warned (though this warning path I don't think could ever have been hit) and we agreed to relegate the warning to a `Debug` log instead. This way, users won't see the load failures for arbitrary DLLs, but if extensions fail to load for some reason and no errors seem to be present, they can run choco with `--debug` or check the logs to see what's going wrong there.

I have added a commit to this effect, so I think we're good for a final review here.

Gary and I also sanity-checked the other extension resolvers being used across several other codebases (CCM, licensed extension, chocolateyGUI) and I will likely also need to update the one being used in the Chocolatey Licensed test setup to match these ones, but we don't think the others need to be touched.

src/chocolatey/infrastructure/registration/AssemblyResolution.cs
```diff
@@ -279,13 +283,93 @@ public static IAssembly load_extension(string assemblySimpleName)
...
+        public static System.Reflection.Assembly resolve_extension_or_merged_assembly(object sender, ResolveEventArgs args)
+        {
+            var requestedAssembly = new AssemblyName(args.Name);
+
+#if FORCE_CHOCOLATEY_OFFICIAL_KEY
+                var chocolateyPublicKey = ApplicationParameters.OfficialChocolateyPublicKey;
+#else
+            var chocolateyPublicKey = ApplicationParameters.UnofficialChocolateyPublicKey;
+#endif
+            if (!requestedAssembly.get_public_key_token().is_equal_to(chocolateyPublicKey))
```
corbob, line 322:
```suggestion
            var requestedAssembly = new AssemblyName(args.Name);

#if FORCE_CHOCOLATEY_OFFICIAL_KEY
            var chocolateyPublicKey = ApplicationParameters.OfficialChocolateyPublicKey;
#else
            var chocolateyPublicKey = ApplicationParameters.UnofficialChocolateyPublicKey;
#endif
            if (!requestedAssembly.get_public_key_token().is_equal_to(chocolateyPublicKey))

```

corbob (changes requested):
Just a minor formatting issue. Still need to take it for a test in a VM...

src/chocolatey/infrastructure/registration/AssemblyResolution.cs
```diff
@@ -279,13 +283,93 @@ public static IAssembly load_extension(string assemblySimpleName)
...
+                var assembly = System.Reflection.Assembly.Load(fullName);
+
+                if (assembly != null)
+                {
+                    return Assembly.set_assembly(assembly);
+                }
+            }
+            // Ignore load failures, so we return null and let the caller handle the failure to load the extension.
+            // A failure here will pretty much always be a public key mismatch, an invalid DLL, or a failure to find
+            // the licensed extension, both of which should be handled by the caller.
```
corbob, line 299:
```suggestion
            // the licensed extension, all of which should be handled by the caller.

```

corbob (changes requested):

corbob (approved):
LGTM!

AdmiringWorm (approved):
LGTM

PR #2926, opened 2022-12-06
(#2865) Don't resolve chocolatey for licensed ext

vexx32:
## Description Of Changes

- Fix case in assembly resolver for when it's blindly loading the licensed extension due to finding a valid license.

## Motivation and Context

Previously, the change to the assembly resolver had one edge case; if there is a chocolatey.license.xml that is found, choco will attempt to load the licensed extension specifically. It does not check before that whether the extension is installed or the file is anywhere it should be expected to be, all that is offloaded to the resolver.

This change ensures that if we're looking for that extension due to loading a valid license, we cannot return the base Chocolatey assembly which would be completely incorrect and confuse a lot of choco's built in license assertions and assembly version logic for that extension.

## Testing

1. Install a valid and current `chocolatey.license.xml`, but DO NOT install any extensions
1. Ensure there is _no_ `extensions` folder in Chocolatey's install directory at all
1. Run `choco list -lo` 
1. You should get just the normal error about the licensed assembly not being found (see below)

image

### Operating Systems Testing

Windows 11

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

Fixes the fix for #2865

### DIFF
--- a/src/chocolatey/infrastructure/registration/AssemblyResolution.cs
+++ b/src/chocolatey/infrastructure/registration/AssemblyResolution.cs
@@ -369,7 +369,8 @@ namespace chocolatey.infrastructure.registration
 
             // There are things that are ILMerged into Chocolatey. Anything with
             // the right public key except extensions should use the choco/chocolatey assembly
-            if (!requestedAssembly.Name.EndsWith(".resources", StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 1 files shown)


corbob:
LGTM, thanks for getting this fixed up @vexx32

PR #2933, opened 2022-12-13
(#2738) Move setup module to install .Net 4.8

TheCakeIsNaOH:
## Description Of Changes

This moves the setup module used during install and upgrade to
check for and install .Net 4.8 instead of .Net 4.0

The timing of the check/install is moved to before the Chocolatey
installation directory is created. This allows for re-running the
install.ps1 script if a reboot is required without needing a manual
cleanup of the partial installation.

This also adds a check for the other reboot required exit code of
1641, and throws the install/upgrade if the installer requires
a reboot.

## Motivation and Context

As Chocolatey CLI is being moved to .Net 4.8, it requires insuring that the required framework is installed.

## Testing

Testing on Server 2012 R2:
- Upgraded from Chocolatey CLI 1.2.1 to a built nupkg from this PR. It installed .net 4.8 and threw as a reboot was required. After a reboot, a second attempt at an upgrade worked.
- Installed .net 4.8, rebooted, and upgraded from Chocolatey CLI 1.2.1 to a built nupkg from this PR. It upgraded without issues.
- Installed .net 4.8, did not reboot, and upgraded from Chocolatey CLI 1.2.1 to a built nupkg from this PR. It installed .net 4.8 (since the installer does not detect a previous install until after a reboot) and threw as a reboot was required (as per the exit code). After a reboot, a second attempt at an upgrade worked.

Testing on clean Windows 7 SP1:
- Installed the built nupkg from this PR via a local copy of the `install.ps1` script. It installed .net 4.8, did not require a reboot, and installed without issues. Local path to the nupkg specified by setting `$env:ChocolateyDownloadUrl`
- Installed Chocolatey 1.2.1, rebooted, and Upgraded from Chocolatey CLI 1.2.1 to a built nupkg from this PR. It installed .net 4.8 and threw as a reboot was required. After a reboot, a second attempt at an upgrade worked.

### Operating Systems Testing
- Windows Server 2012 R2
- Windows 7 SP1

## Change Types Made

* [ ] Bug fix (non-breaking change).
* [x] Feature / Enhancement (non-breaking change).
* [x] Breaking change (fix or feature that could cause existing functionality to change).
* [ ] Documentation changes.
* [ ] PowerShell code changes.

## Change Checklist

* [x] Requires a change to the documentation.
* [ ] Documentation has been updated.
* [ ] Tests to cover my changes, have been added.
* [x] All new and existing tests passed?
* [x] PowerShell code changes: PowerShell v2 compatibility checked?

## Related Issue

[...]

PR #2935, opened 2022-12-14
(#2738) Check installed .NET version installed is at least 4.8

TheCakeIsNaOH:
## Description Of Changes

This adds a registry check to `chocolatey.console` to validate that .Net 4.8 or later is installed. This check is added because Chocolatey CLI can start to execute but runs into errors when the system has an older version of .Net framework installed. The check is very early on in the execution so as to print out a helpful error before any errors due to incompatibilities with the older .Net version show up.

## Motivation and Context

It is better to have a check and helpful message than various unhelpful error messages.

## Testing

1. Build PR 
1. On Server 2012 R2 (with the default .Net 4.5.2 installed), upgrade from Chocolatey CLI v1.2.1 via the nupkg built in step 1
1. Validate that `choco` commands cause an error (which happens too early in execution for a check to be added)
1. Update .Net to 4.6.0 and reboot
1. Validate that `choco` commands print out the error messaged added in this PR
1. Update .Net to 4.8.0 and reboot
1. Validate that `choco` commands do not trigger the error messaged added in this PR

Additionally, test on Windows 10 with .Net 4.8 installed that `choco` commands do not trigger the error messaged added in this PR

### Operating Systems Testing
- Windows Server 2012R2
- Windows 10 22H2

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

Part of  #2738
https://app.clickup.com/t/20540031/PROJ-368

### DIFF
--- a/src/chocolatey.console/Program.cs
+++ b/src/chocolatey.console/Program.cs
@@ -21,6 +21,7 @@ namespace chocolatey.console
     using System.IO;
     using System.Linq;
     using System.Reflection;
+    using Microsoft.Win32;
     using chocolatey.infrastructure.information;
     using infrastructure.app;
     using infrastructure.app.builders;
@@ -30,6 +31,7 @@ namespace chocolatey.console
     using infrastructure.extractors;
     using infrastructure.licensing;
     using infrastructure.logging;
+    using infrastructure.platforms;
     using infrastructure.registration;
     using infrastructure.tolerance;
     using SimpleInjector;
@@ -123,6 +125,8 @@ namespace chocolatey.console
                         "chocolatey".Log().Info(ChocolateyLoggers.Important, () => "Please run 'choco -?' or 'choco <command> -?' for help menu.");
                     }
                 }
+
+                check_installed_dotnetfx_version();
 
                 if (warnings.Count != 0 && config.RegularOutput)
                 {
@@ -296,6 +300,28 @@ choco feature enable --name=""'disableCompatibilityChecks'""
 
 Or by passing the --skip-compatibility-checks option when executing a
 command.");
+        }
+
+        private static void check_installed_dotnetfx_version()
+        {
+            if (Platform.get_platform() == PlatformType.Windows)
+            {
+                // https://learn.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#minimum-version
+                const int NET48RELEASEBUILD = 528040;
+                const string REGKEY = @"SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full\";
+
+                using (var ndpKey = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry32).OpenSubKey(REGKEY))
+                {
+                    if (ndpKey == null || ndpKey.GetValue("Release") == null || (int)ndpKey.GetValue("Release") < NET48RELEASEBUILD)
+                    {
+                        throw new ApplicationException(
+                            @".NET 4.8 is not installed or may need a reboot to complete installation.
+Please install .NET Framework 4.8 manually and reboot the system.
+Download at 'https://download.visualstudio.microsoft.com/download/pr/2d6bb6b2-226a-4baa-bdec-798822606ff1/8494001c276a4b96804cde7829c04d7f/ndp48-x86-x64-allos-enu.exe'"
+                                .format_with(Environment.NewLine));
+                    }
+                }
+            }
         }
     }
 }


src/chocolatey.console/Program.cs
```diff
@@ -44,6 +44,8 @@ namespace chocolatey.console
     using Console = System.Console;
     using Environment = System.Environment;
     using IFileSystem = infrastructure.filesystem.IFileSystem;
+    using Microsoft.Win32;
+    using chocolatey.infrastructure.platforms;
```
gep13, line 48: Can we order these so that there are in alphabetical order?

src/chocolatey.console/Program.cs
```diff
@@ -370,5 +374,24 @@ private static void write_warning_for_incompatible_versions()
...
+            {
+                const string REGKEY = @"SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full\";
+                const int NET48RELEASEBUILD = 528040;
+
+                using (var ndpKey = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry32).OpenSubKey(REGKEY))
+                {
+                    if (ndpKey == null || ndpKey.GetValue("Release") == null || (int)ndpKey.GetValue("Release") < NET48RELEASEBUILD)
+                    {
+                        throw new ApplicationException(
+                            ".NET 4.8 is not installed or may need a reboot to complete installation{0}Install .NET 4.8 and reboot the system"
```
gep13, line 390: Can this use similar wording to what is done here: https://github.com/chocolatey/choco/pull/2933/files#diff-21bdb1742369a87a1e2f23b0c59b77df037fe5d3ded2862a14cef9be2bfd13e6R788 including the link to the download location?

src/chocolatey.console/Program.cs
```diff
@@ -370,5 +374,24 @@ private static void write_warning_for_incompatible_versions()
 Or by passing the --skip-compatibility-checks option when executing a
 command.");
         }
+
+        private static void check_installed_dotnetfx_version()
+        {
+            if (Platform.get_platform() == PlatformType.Windows)
```
gep13, line 380: Do we need to worry about execution on an earlier version of Mono?

src/chocolatey.console/Program.cs
```diff
@@ -370,5 +374,24 @@ private static void write_warning_for_incompatible_versions()
 Or by passing the --skip-compatibility-checks option when executing a
 command.");
         }
+
+        private static void check_installed_dotnetfx_version()
+        {
+            if (Platform.get_platform() == PlatformType.Windows)
+            {
+                const string REGKEY = @"SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full\";
+                const int NET48RELEASEBUILD = 528040;
```
gep13, line 383: Where did this number come from? If from documentation on Microsoft site, can you add a link in a comment?

src/chocolatey.console/Program.cs
```diff
@@ -124,6 +126,8 @@ private static void Main(string[] args)
                     }
                 }
 
+                check_installed_dotnetfx_version();
```
gep13, line 129: Is there a reason not to do this as the very first thing that is done in this method?

gep13 (changes requested):
This looks great to me! I have left some comments/questions.

gep13:
@TheCakeIsNaOH am I right in saying that even with this change in place, there is still an error when running on .NET Framework 4.5.1? i.e. we don't see this actual error that we want to throw?

gep13:
@TheCakeIsNaOH @AdmiringWorm are there any tests that we can do around this change? Or do we not think that they are required?

src/chocolatey.console/Program.cs
TheCakeIsNaOH: If I do it at the beginning of the method, logging it not set up, so it does not output anything to the screen or to the logs.

TheCakeIsNaOH:
> am I right in saying that even with this change in place, there is still an error when running on .NET Framework 4.5.1? i.e. we don't see this actual error that we want to throw?

Correct, the error is about the `System.Array.Empty()` method not being found.

src/chocolatey.console/Program.cs
TheCakeIsNaOH: Maybe. 

Mono 6.6 is the first version with the .NET 4.8 assemblies, so anything earlier than that may not work.

However, the only recommended current way to run choco on non-Windows is in the docker container, and that has a new enough version of Mono. 

It is possible to get the display version of mono via reflection, but I don't see a great way to parse it that would work for versions from custom/git builds of mono and all of the various possible OS names. 
https://stackoverflow.com/questions/4178129/how-to-determine-the-revision-from-which-current-mono-runtime-was-built-and-inst

src/chocolatey.console/Program.cs
TheCakeIsNaOH: Link added.

src/chocolatey.console/Program.cs
TheCakeIsNaOH: I've updated the wording, hopefully it looks better?

src/chocolatey.console/Program.cs
TheCakeIsNaOH: Reordered

AdmiringWorm:
> @TheCakeIsNaOH @AdmiringWorm are there any tests that we can do around this change? Or do we not think that they are required?

Could perhaps be possible to do this as part of the E2E tests, but that would require manipulating the registry as part of the initialization (and restoring it once the test has run).

I think @corbob may be in a better position to answer that though since I think he has experience with manipulating the necessary part of the registry.

src/chocolatey.console/Program.cs
TheCakeIsNaOH: I've updated it to include the download URL.

corbob:
The `NDP` keys in the registry are owned by everyone's favorite account: `TrustedInstaller`. That means in order for us to change them, we need to change the owner away from `TrustedInstaller`, and then grant modify permissions. The last time I looked at automating something like that I think ended up finding `regini` as an option, but the documentation about the correct incantation is a little sparse, and it requires the Windows Server 2003 resource kit to actually get `regini`. We could possibly set the permissions through PowerShell, but I don't know of any resources showing how you might do that.

All of that to say: If we can get permissions to the registry keys needed, there's no reason we shouldn't have an E2E test to cover it, but I'm not aware of automated ways to set the permissions.

In case anyone decides to look at regini: https://learn.microsoft.com/en-US/troubleshoot/windows-client/application-management/change-registry-values-permissions

TheCakeIsNaOH:
> We could possibly set the permissions through PowerShell, but I don't know of any resources showing how you might do that.

Yes, it is possible. I have some internal packages that take ownership from trusted installer to set registry keys.

The trick is adjust token privileges to add the `SeTakeOwnershipPrivilege`, and then try to take ownership. 
The blog post to let powershell add token privileges is here:
https://www.leeholmes.com/adjusting-token-privileges-in-powershell/

And there is an example here of how to take ownership:
https://social.technet.microsoft.com/Forums/en-US/e718a560-2908-4b91-ad42-d392e7f8f1ad/take-ownership-of-a-registry-key-and-change-permissions?forum=winserverpowershell

This code has been bouncing around for a while, so I would expect there to be lots of other examples.

src/chocolatey.console/Program.cs
gep13: Ah, gotcha. Makes sense.

src/chocolatey.console/Program.cs
gep13: This also makes sense. Nothing else needs to be done here.

gep13 (approved):
LGTM!

gep13:
@TheCakeIsNaOH thanks for getting this added!

PR #2940, opened 2022-12-16
(#2887) Tab completion: only complete license-appropriate things

### LINKED ISSUE #2887: PowerShell tab completion add in licensed commands/options only if license exists.

It would be good if the PowerShell tab completion had the licensed commands and options available if and only if a license (and/or the licensed extension) is available. If they are added always, then open source users have a worse experience (as there are switches added that they can't use), and if they are not added, then licensed users are missing some switches.

Therefore, the tab completion should add in the licensed commands and options available if and only if a license (and/or the licensed extension) is available.

## Additional Context.

As per todo comment in code here:
https://github.com/chocolatey/choco/blob/499aa20223a66b013bd2a56dd539701ab3cb85eb/src/chocolatey.resources/helpers/ChocolateyTabExpansion.ps1#L74-L75

This can be moved to the licensed issues repository if the code ends up in the licensed extension instead of in open source.

## Related Issues

Todo comment noticed here: https://github.com/chocolatey/choco/pull/2799#discussion_r984749907

vexx32:
## Description Of Changes

- Only add licensed / Business-only commands and options to tab completion if the user has the appropriate Chocolatey license.

Note that this check is deliberately naive; we don't look at the license too hard, just try to determine if there _is_ one and if it is a Business/Trial license. Since this only affects tab completion and Choco does its own license checks separately, we don't need to be too stringent about exactly what is needed for the licensed options to be returned.

In future, we should simply query available commands and options from choco directly, but this is a "good enough" patch in the meantime until that day comes.

## Motivation and Context

Previously all commands and options would be returned from tab completion regardless of whether they could actually be used.

With this change, tab completion (superficially) checks the user's Chocolatey license without checking too hard and populates the available commands and options appropriately.

## Testing

1. Ensure you have a Business license installed for Chocolatey (at `$env:ChocolateyInstall\license\chocolatey.license.xml`).
1. Import the `chocolateyProfile.psm1` module in a PowerShell session
1. Type `choco[SPACE]` and tab through the options.
1. You should have the usual FOSS commands, as well as `sync`, `optimize`, and `download`.
1. Add the `new` command to the current input and tab-complete through the options from there.
1. You should have the usual FOSS options, plus licensed options like `--build-package`, and also the business-only options like `--file`.
1. Try a few other commands and ensure that all expected options are available.

- Repeat the above with a `Professional` license installed instead, and you should have similar results to the above, though the specific business-only command `sync` should be missing, and `--file` and other Business-only options should be missing.
- Repeat once more with no license installed at all, and you should see all the licensed commands and options are no longer present.

### Operating Systems Testing

I tested on Windows 11, we should test on a few more platforms before merging.

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

* [ ] PowerShell code changes: PowerShell v2 compatibility checked? (TBD)

## Related Issue

Fixes #2887

### DIFF
--- a/src/chocolatey.resources/helpers/ChocolateyTabExpansion.ps1
+++ b/src/chocolatey.resources/helpers/ChocolateyTabExpansion.ps1
@@ -26,7 +26,7 @@ $script:choco = "$env:ChocolateyInstall\choco.exe"
 
 function script:chocoCmdOperations($commands, $command, $filter, $currentArguments) {
     $currentOptions = @('zzzz')
-    if ($currentArguments -ne $null -and $currentArguments.Trim() -ne '') {
+    if (-not [string]::IsNullOrWhiteSpace($currentArguments)) {
         $currentOptions = $currentArguments.Trim() -split ' '
     }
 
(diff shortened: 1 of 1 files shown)


src/chocolatey.resources/helpers/ChocolateyTabExpansion.ps1
```diff
@@ -35,56 +35,93 @@ function script:chocoCmdOperations($commands, $command, $filter, $currentArgumen
...
+        $commandOptions.install += $proInstallUpgradeOptions
+        $commandOptions.upgrade += $proInstallUpgradeOptions + " --exclude-chocolatey-packages-during-upgrade-all --include-chocolatey-packages-during-upgrade-all"
+        $commandOptions.new += " --build-package --use-original-location --keep-remote --url='' --url64='' --checksum='' --checksum64='' --checksumtype='' --pause-on-error"
+        $commandOptions.pin += " --note=''"
+
+        # Add Business-only commands and options if the license is a Business or Trial license
+        [xml]$xml = Get-Content -Path $licenseFile.FullName
+        $licenseType = $xml.license.type
+
+        if ($licenseType -in 'Business', 'BusinessTrial') {
```
corbob, line 92:
```suggestion
        if ('Business', 'BusinessTrial' -contains $licenseType) {

```
Need to use `-contains` because `-in` wasn't introduced until [PowerShell 3](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comparison_operators?view=powershell-7.3#-in-and--notin)

corbob (changes requested):
Just one minor PowerShell 2 compatibility change. Will take this for a spin on a VM shortly.

src/chocolatey.resources/helpers/ChocolateyTabExpansion.ps1
vexx32: Good catch, thanks lol

src/chocolatey.resources/helpers/ChocolateyTabExpansion.ps1
```diff
@@ -35,56 +35,94 @@ function script:chocoCmdOperations($commands, $command, $filter, $currentArgumen
...
+            }
+        }
+    }
 }
 catch {
 }
 
+foreach ($key in $commandOptions.Keys) {
+    $commandOptions.$key += $allcommands
+}
```
corbob, line 117: This doesn't seem to work. And I haven't been able to find a way that does work... 
image

vexx32: Ah yep, classic footgun there, love those. I've amended the commit to collect the keys in a separate array before going and changing the hashtable.

corbob (changes requested):
I have to reject this due to behaviour that Rain and I have been discussing on Slack. 

The behaviour I'm seeing in the testing: if you try to tab-complete a path that contains choco or chocolatey as a distinct part of the path it locks up the shell to the point you have to close the terminal. Tab completing `cd \chocoR<tab>` works for the test vagrant that has `C:\chocoRoot`, but `cd \choco<tab>` locks up. As well `ipmo \programdata\chocolatey\hel<tab>` locks up (hence the belief that it's choco or chocolatey as distinct path parts).

I don't know what in the code is causing this, but the rejection will at least prevent us from accidentally merging this.

corbob (approved):
LGTM!

corbob:
Thanks for getting this updated @vexx32 👍

PR #2952, opened 2022-12-28
(#2854) Add helper to read config values

### LINKED ISSUE #2854: Create PowerShell helper function to read config values

When running some types of hook packages, it may be ideal or required to retrieve config values from some location. After discussion with @gep13, the `config` section of the `chocolatey.config` it would most likely be a good location for users to set config values for hooks to use. However, there is no easy way for hooks to retrieve these values. 

So it would be good to add a helper function to retrieve these values, so they can be used in a hook script.

## Additional Context.

N/A

## Related Issues

- Issue #1185

TheCakeIsNaOH:
## Description Of Changes

This adds a PowerShell helper to get config values from the Chocolatey configuration file. This helper is intended to be used to allow hook scripts to access configuration values set in the config.

## Motivation and Context

See #2854 and https://github.com/chocolatey-community/chocolatey-hooks/issues/6#issuecomment-1334857165

## Testing

TODO

### Operating Systems Testing
- Windows 10 22H2

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

* [x] PowerShell code changes: PowerShell v2 compatibility checked?

## Related Issue

Fixes #2854


src/chocolatey.resources/helpers/functions/Get-ChocolateyConfigValue.ps1
```diff
@@ -0,0 +1,65 @@
...
+        [Object[]] $ignoredArguments
+    )
+
+    try {
+        $installLocation = Get-ChocolateyPath -pathType 'InstallPath'
+        $configPath = Join-Path $installLocation "config\chocolatey.config"
+        [xml]$configContents = Get-Content -Path $configPath
+        return $configContents.chocolatey.config.add | `
+                Where-Object { $_.key -eq $configKey } | `
+                Select-Object -ExpandProperty value
```
corbob, line 58: General PowerShell practices recommend against line continuation, and there is prior art within the Chocolatey codebase for it. However, I cannot find any instances of using it in places where a line break is handled by PowerShell without it (ie: after a pipe character as used here). I do however see a number of instances where we do just that, such as here: https://github.com/chocolatey/choco/blob/06acb28162b6ee3035acafe91561d8969244f400/src/chocolatey.resources/helpers/functions/Install-ChocolateyShortcut.ps1#L218-L219, and here: https://github.com/chocolatey/choco/blob/06acb28162b6ee3035acafe91561d8969244f400/src/chocolatey.resources/helpers/functions/Install-ChocolateyInstallPackage.ps1#L317-L319

As such, could we remove them here.

```suggestion
        return $configContents.chocolatey.config.add |
                Where-Object { $_.key -eq $configKey } |
                Select-Object -ExpandProperty value
```

src/chocolatey.resources/helpers/functions/Get-ChocolateyConfigValue.ps1
```diff
@@ -0,0 +1,65 @@
...
+    }
+    catch {
+        Write-Warning "Unable to read config value '$configKey' with error"
+        Write-Warning $_
+        return ""
```
corbob, line 63: The Comment Based Help on Lines 31-33 says that we will output nothing if we can't read the key. As such, we should probably just not have a return here.

```suggestion
```

This would also keep with PowerShell expectations where a Get that can't find what it's looking for either throws an error or returns nothing.

corbob (changes requested):
A couple of suggestions. As well, we would do well to either add some Pester tests that exercises these, or open an issue to track adding pester tests.

Some tests that I see above and beyond the testing listed in the PR description would be to change the config values, and ensure that the changed values are read. As well as adding non-standard values and ensure that they too can be read.

src/chocolatey.resources/helpers/functions/Get-ChocolateyConfigValue.ps1
corbob: In fact... Do we know what scenario we might expect an error here? It would be good to understand how to force this code path, but I'm not entirely sure short of a corrupted config file which Chocolatey may already correct or error beforehand (unless maybe non-admin 🤔 )

TheCakeIsNaOH: I think it would take either corrupted xml syntax or a missing/no read permissions config file to trigger this path.

src/chocolatey.resources/helpers/functions/Get-ChocolateyConfigValue.ps1
TheCakeIsNaOH: Fixed

src/chocolatey.resources/helpers/functions/Get-ChocolateyConfigValue.ps1
TheCakeIsNaOH: Removed the return statement

TheCakeIsNaOH:
> As well, we would do well to either add some Pester tests that exercises these, or open an issue to track adding pester tests.

Agreed. I don't really have a good setup to iterate on the Pester tests, so a separate issue, or waiting until the behavior is nailed down is the way to go.

pauby:
> we would do well to either add some Pester tests that exercises these, or open an issue to track adding pester tests.

Let's get some tests added as part of _this_ PR before we merge it. Is were adding code line this, we need to be adding tests too.

corbob:
> Let's get some tests added as part of this PR before we merge it. If we're adding code line this, we need to be adding tests too.

@TheCakeIsNaOH as you're not in a position to test the pester tests here, I'll get these tests added to the PR.

src/chocolatey.resources/helpers/functions/Get-ChocolateyConfigValue.ps1
corbob: I've discussed with the team this bit. I'm going to update this to Write-Error with the exception object attached so that the package author can handle it in a PowerShell way instead of just swallowing the exception. I have been able to cause this catch to be triggered, but only in wholly unsupported ways.

corbob (approved):

gep13:
@corbob thanks for getting this over the line, and thanks to @TheCakeIsNaOH for the initial implementation!

TheCakeIsNaOH:
Thanks @corbob for creating the tests.

Issue #508, opened 2015-12-16
Upgrade to use version 6.4.0 of NuGet.Client assemblies


ferventcoder:
https://github.com/NuGet/Home/issues/1839 

Upgrade NuGet.Core to a version that better handles conflict resolution

May fix #507
May fix #227
May fix #116
May fix #1158
Relates to #2592 - support for embedded icons

### References
- https://github.com/NuGet/NuGetGallery/issues/3004#issuecomment-223145697
- http://stackoverflow.com/a/34959690
- https://daveaglick.com/posts/exploring-the-nuget-v3-libraries-part-1
- https://daveaglick.com/posts/exploring-the-nuget-v3-libraries-part-2
- https://daveaglick.com/posts/exploring-the-nuget-v3-libraries-part-3
- https://martinbjorkstrom.com/posts/2018-09-19-revisiting-nuget-client-libraries
- https://devblogs.microsoft.com/nuget/Play-with-packages/

ferventcoder:
https://github.com/NuGet/Home/issues/1870

ferventcoder:
This will require .NET 4.5 for the upgrade to a v3 version of NuGet so this will need to move to 0.9.11 where we are eclipsing support for .NET 4 (and Windows 2003).

timbrown5:
Hi ferventcoder,
I know it's a bit early (as 0.9.10 has only just been released) but I was wondering if you have an estimate for when this work might be started. It would be good to test if this fixes the issues I was having (even if it's a development build).

ferventcoder:
It is definitely a bit early. No estimate yet on this.

timbrown5:
Just wondered if there is any update on when this might be fixed?

ferventcoder:
Looking at late 2017 depending on where it falls in prioritization. Right now it is a few back on the list. We have a pretty aggressive schedule so it could be done in 2017 but it may fall back to 2018.

timbrown5:
Hi @ferventcoder,
Is there any update on this?
Thanks,
Tim

ferventcoder:
@timbrown5 Note the Milestone - that gives you the best idea on where things are. It's definitely scheduled, but priorities shift based on customer needs, etc.

eriviere:
What is the current version of NuGet used by Chocolatey?
I am unable to install a NuGet Package because it believes I am using version 2.11 of Nuget. I am actually using choco version 0.10.11

The error is this one:
[...]

PR #2998, opened 2023-01-25
(#508) Fix NuGet pushes to correctly result in errors

corbob:
## Description Of Changes

* Update Push command to not ignore duplicate packages reported from repositories
* Update tests to work with the new wording of NuGet push failures
* Add some wording to the NuGet push tests to indicate which version it was attempting to push.

## Motivation and Context

* The initial implementation wasn't throwing when `409` was returning from the server.
* Tests were disabled as they weren't correctly throwing

## Testing

* Ran tests in Team City against the PR build targeting this branch for the pester tests. This resulted in 11 failed tests (22 total as we test two OSes). The 4 related to NuGet feed pushes are known to fail in Team City, and do not relate to the push failing, more that we couldn't verify the push was successful as Team City run Test Kitchen isn't getting package listings at the moment. The other 7 tests also failed on Kim's PR 3001, and I'm not sure as they're not related to any code that I touched at all.

image

* Ran tests in local Test Kitchen targeting my CloudSmith repository, confirmed that packages were indeed pushed to both a v2 and v3 CloudSmith endpoint
* Ran tests in local Test Kitchen targeting a MyGet repository. Confirmed that packages pushed successfully to the v2 endpoint. Unclear about if there even is a v3 endpoint (there is a note about pushing symbols to the v3 endpoint, but choco failed to push to that endpoint saying the credentials were invalid)

### Operating Systems Testing

Windows Server 2016, 2019 (Test Kitchen environments)

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

* Relates to #508
* PROJ-466

### DIFF
--- a/src/chocolatey/infrastructure.app/nuget/NugetPush.cs
+++ b/src/chocolatey/infrastructure.app/nuget/NugetPush.cs
@@ -27,6 +27,7 @@ namespace chocolatey.infrastructure.app.nuget
     using NuGet.Configuration;
     using NuGet.Protocol;
     using NuGet.Protocol.Core.Types;
+    using System.Net.Http;
 
     public class NugetPush
     {
@@ -39,8 +40,10 @@ namespace chocolatey.infrastructure.app.nuget
             }
             const bool disableBuffering = false;
 
-            // Controls adding /api/v2/packages to the end of the source url
-            const bool noServiceEndpoint = true;
+            // Controls adding /api/v2/packages to the end of the source url, IF that source url is missing an endpoint (is just a host with no path like https://push.chocolatey.org/)
+            const bool noServiceEndpoint = false;
+            // Controls if NuGet throws on 409 HTTP response
+            const bool skipDuplicate = false;
 
             //OK to use FirstOrDefault in this case as the command validates that there is only one source
             SourceRepository sourceRepository = NugetCommon.GetRemoteRepositories(config, nugetLogger, filesystem).FirstOrDefault();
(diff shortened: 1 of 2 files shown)


src/chocolatey/infrastructure.app/nuget/NugetPush.cs
```diff
@@ -40,7 +41,9 @@ public static void push_package(ChocolateyConfiguration config, string nupkgFile
             const bool disableBuffering = false;
 
             // Controls adding /api/v2/packages to the end of the source url
-            const bool noServiceEndpoint = true;
+            const bool noServiceEndpoint = false;
```
TheCakeIsNaOH, line 44: We may need to do some further parsing/checking here, assuming this change fixes pushing to CCR, since I think it break would break pushing to Nexus (IIRC).

corbob: That's what I'm not entirely sure on. Right now I'm not able to push to a v3 feed from Visual Studio, it prompts me for username/password no matter what (even reverting to the head of develop). But from my tests against v2 it had no problem. I'm running this on Test Kitchen right now to see how it behaves with pushing to the v3 feeds there.

Assuming that works, great, if not then I'm thinking we just wrap this:

```csharp
            // Controls adding /api/v2/packages to the end of the source url
            bool noServiceEndpoint = true;

            if (config.Sources == ApplicationParameters.ChocolateyCommunityFeedPushSource
                || config.Sources == ApplicationParameters.ChocolateyCommunityFeedPushSourceOld)
            {
                noServiceEndpoint = false;
            }
```

corbob: Actually, I can explain why it being set to false is sufficient. When it's building the URI to push to, it looks at the absolute path, and if the absolute path after removing the starting `/` characters is empty, and `noServiceEndpoint` is `false`, then it sets the path to `/api/v2/package/` plus whatever it was asked to append (in our case nothing). So when you already have it going to say `https://nexus.local.testing/repository/test/` then the absolute path is `/repository/test/`. So it is only expected to add `/api/v2/package/` if you're going to a direct hostname (`https://push.chocolatey.org`, `https://nuget.org`, etc)

The relevant line: https://github.com/chocolatey/NuGet.Client/blob/40e0be43ff5afc6962fac29156942c4e9650a979/src/NuGet.Core/NuGet.Protocol/Resources/PackageUpdateResource.cs#L808

gep13: Based on this comment, and additional conversations around this topic, going to move forward with getting this merged.

gep13 (approved):
LGTM!

gep13:

PR #2999, opened 2023-01-26
(#1935) Add tests to ensure upstream NuGet version is used.

### LINKED ISSUE #1935: When specific nuget version is needed by a package it is the chocolatey version that is used
I have a problem similar of the one in ticket 194.
I have received this message during installation of a software with choco install:

    The 'Microsoft.NETCore.Platforms 3.0.0' package requires NuGet client version '2.12' or above, but the current NuGet version is '2.11.0.0'.

I have looked at ticket 194 but I need more help to solve it again.

So please tell me how to solve it.
Thanks,
Mario

TheCakeIsNaOH:
## Description Of Changes

The added integration tests ensure that a minClientVersion attribute that is higher
then the Chocolatey and NuGet assembly versions does not prevent installation
or packing. 

And a unit test ensures that the user agent is set, when NuGet sources are configured
and that it uses the upstream version of NuGet that the NuGet assembles are
based on. This same version is used in other areas as well, like minClientVersion,
so it applies to multiple areas.

## Motivation and Context

See #1935 

With the uplifted NuGet assemblies, the `minClientVersion` is no longer checked at all, bypasses the fixes the issue of the Chocolatey assembly version being used instead of the NuGet assembly version being used.

## Testing

Ran tests in Visual Studio and with `build.ps1`
Built and installed package with `minClientVersion` set to `100.0.0` to manually to ensure that it packs and installs ok.

### Operating Systems Testing
- Windows 10 22H2

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

Fixes #1935

### DIFF
--- a/src/chocolatey.tests.integration/context/unsupportedelements/1.0.0/unsupportedelements.nuspec
+++ b/src/chocolatey.tests.integration/context/unsupportedelements/1.0.0/unsupportedelements.nuspec
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
-  <metadata>
+  <metadata minClientVersion="100.0.0">
     <id>unsupportedelements</id>
     <version>1.0.0</version>
     <title>unsupportedelements</title>
--- a/src/chocolatey.tests.integration/context/unsupportedelements/1.1.0/unsupportedelements.nuspec
+++ b/src/chocolatey.tests.integration/context/unsupportedelements/1.1.0/unsupportedelements.nuspec
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
-  <metadata>
+  <metadata minClientVersion="100.0.0">
     <id>unsupportedelements</id>
     <version>1.1.0</version>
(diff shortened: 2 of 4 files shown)


gep13:
@TheCakeIsNaOH the failing tests here would indicate that something thinks it is using 6.4.1 and not 6.4.0:

> Actual:   Chocolatey Command Line/ via NuGet Client/6.4.1 (Unix 21.6.0.0)

I don't think that there is a 6.4.1 release yet, so I am not clear on where this value is coming from.  Is this something you would have some time to dig into? If not, I will add this to my list of things to look at.

TheCakeIsNaOH:
> I don't think that there is a 6.4.1 release yet, so I am not clear on where this value is coming from.

The version value used in the user agent (and for checking against the `minClientVersion`, if used) is from here:
https://github.com/NuGet/NuGet.Client/commit/b9acf15930e419d373e9a62fa5aef61d0da22fab

I hard coded it explicitly to `6.4.1.zlocal.1` when testing, since that is the upstream (aka unreleased) version I was testing against. It looks like that value was not updated to `6.4.0` when the changes were rebased onto 6.4.x

The reason it is hard coded is that otherwise it would check the assembly version, which means it would have the Chocolatey assembly version when ilmerged.

pauby:
CU-PROJ-568

gep13:
@TheCakeIsNaOH thank you for getting this added!

PR #3011, opened 2023-02-02
(#508) Update methods to work with ordering/count

gep13:
## Description Of Changes

There have been some upstream changes in the NuGet.Client assemblies to provide the ability to both order by DownloadCount of packages, as well as to provide information about the Count of packages for a given search.

With these added, we need to taken advantage of these methods.

The commit makes use of the new SearchCountAsync method, which is then later used in things like Chocolatey GUI.

In addition, it changes the default to use the searchResource instead of the listResource.  When a given source doesn't have a searchResource available, it will now fallback to the listResource, rather than the other way around.

Finally, the paging of a search is now assumed to be provided via the searchResource, rather than directly in Chocolatey CLI.  This change was made given the complexities of expected results when multiple sources are used, i.e. what constitutes a single page of results. Additional work may be required in this area once we do more testing.

## Motivation and Context

Changes are needed here so that upstream applications like Chocolatey GUI can work as expected.

## Testing

1. Clone and checkout the branch associated with this PR: https://github.com/chocolatey/NuGet.Client/pull/25
2. Clone and checkout the branch associated with this PR - i.e. the one you are readin
3. Build NuGet.Client and pull in the complied assemblies to the chocolatey/choco repository
4. Build Chocolatey/chco and pull in the compiled assemblies to Chocolatey GUI
5. Run Chocolatey GUI from within Visual Studio - ensure that you have CCR as a source
6. Click on the `chocolatey` source and ensure that AdobeReader and Chocolatey are the first packages in the list
7. Ensure that there are something like 192 pages of packages
8. Click on the last page and see the results appear (there likely won't be a full page of packages)

### Operating Systems Testing

- Windows 10

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

Relates to #508

### DIFF
--- a/src/chocolatey/infrastructure.app/nuget/NugetList.cs
+++ b/src/chocolatey/infrastructure.app/nuget/NugetList.cs
@@ -47,7 +47,26 @@ namespace chocolatey.infrastructure.app.nuget
 
         public static int GetCount(ChocolateyConfiguration configuration, ILogger nugetLogger, IFileSystem filesystem)
         {
-            return execute_package_search(configuration, nugetLogger, filesystem).GetAwaiter().GetResult().Count();
+            var packageRepositories = NugetCommon.GetRemoteRepositories(configuration, nugetLogger, filesystem);
+            var packageRepositoriesResources = NugetCommon.GetRepositoryResources(packageRepositories);
+            string searchTermLower = configuration.Input.to_lower();
+
+            SearchFilter searchFilter = new SearchFilter(configuration.Prerelease);
+            searchFilter.IncludeDelisted = configuration.ListCommand.LocalOnly;
+
+
+            int totalCount = 0;
+            foreach (var repositoryResources in packageRepositoriesResources)
+            {
+                if (repositoryResources.searchResource == null)
+                {
+                    continue;
+                }
+
+                totalCount += repositoryResources.searchResource.SearchCountAsync(searchTermLower, searchFilter, nugetLogger, CancellationToken.None).GetAwaiter().GetResult();
+            }
+
+            return totalCount;
         }
 
         private async static Task<IQueryable<IPackageSearchMetadata>> execute_package_search(ChocolateyConfiguration configuration, ILogger nugetLogger, IFileSystem filesystem)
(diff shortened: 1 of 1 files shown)


src/chocolatey/infrastructure.app/nuget/NugetList.cs
```diff
@@ -47,57 +47,81 @@ public static IEnumerable<IPackageSearchMetadata> GetPackages(ChocolateyConfigur
...
             string searchTermLower = configuration.Input.to_lower();
+
             SearchFilter searchFilter = new SearchFilter(configuration.Prerelease);
             searchFilter.IncludeDelisted = configuration.ListCommand.LocalOnly;
-            var cacheContext = new ChocolateySourceCacheContext(configuration);
+            searchFilter.OrderBy = SearchOrderBy.Id;
 
+            if (configuration.ListCommand.OrderByPopularity)
+            {
+                searchFilter.OrderBy = SearchOrderBy.DownloadCount;
```
TheCakeIsNaOH, line 84: Unless order by ID is specified on the command line, this should default to download count, unless I'm missing something. If order by popularity is specified on the command line, then at the end, a sort of the packages from all the repositories by download count should be done.

src/chocolatey/infrastructure.app/nuget/NugetList.cs
```diff
@@ -47,57 +47,81 @@ public static IEnumerable<IPackageSearchMetadata> GetPackages(ChocolateyConfigur
...
+
+            int totalCount = 0;
+            foreach (var repositoryResources in packageRepositoriesResources)
+            {
+                if (repositoryResources.searchResource == null)
+                {
+                    continue;
+                }
+
+                totalCount += repositoryResources.searchResource.SearchCountAsync(searchTermLower, searchFilter, nugetLogger, CancellationToken.None).GetAwaiter().GetResult();
```
TheCakeIsNaOH, line 66: This need to be updated to account for a couple of specific cases:
- If `--exact` is passed in, in which case the count is one or zero.
- If `--exact --all-versions` is passed in.
- If `--all-versions` is passed in.

TheCakeIsNaOH (changes requested):

src/chocolatey/infrastructure.app/nuget/NugetList.cs
AdmiringWorm: > Unless order by ID is specified on the command line, this should default to download count, unless I'm missing something. 

Ordering by the identifier is the default in Chocolatey CLI, ordering by the popularity/download count is an opt-in feature when searching for packages through the CLI, so this code is currently correct.

> If order by popularity is specified on the command line, then at the end, a sort of the packages from all the repositories by download count should be done.

That is a good point, and I agree that should probably be done, both for the ID and the download count before we return all the results.

src/chocolatey/infrastructure.app/nuget/NugetList.cs
```diff
@@ -47,57 +47,81 @@ public static IEnumerable<IPackageSearchMetadata> GetPackages(ChocolateyConfigur
...
-                        {
-                            results.Add(enumerator.Current);
-                            if (results.Count > totalToGet)
-                            {
-                                break;
-                            }
-                        }
-                    }
-                    else
+                    if (repositoryResources.searchResource != null)
```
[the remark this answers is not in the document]
TheCakeIsNaOH: Actually, after a second look at this, this is actually ok, my previous comment is invalid.

src/chocolatey/infrastructure.app/nuget/NugetList.cs
TheCakeIsNaOH: Sounds good.

AdmiringWorm (dismissed):
LGTM, some additional changes are needed. But this will be part of a different pull request.

src/chocolatey/infrastructure.app/nuget/NugetList.cs
gep13: @TheCakeIsNaOH as we discussed, while these are needed, this initial scope of work was to get the immediate issues with Chocolatey GUI resolved. We can take these as known issues into the first alpha of Chocolatey GUI, and work on correcting this once some more testing has been done. This will tie into the work around verifying that all of the emitted queries from Chocolatey CLI match what was emitted before.

gep13:
@AdmiringWorm I am going to assume that the failing Mac build is a glitch in the matrix, but we can revisit this if it becomes a problem.  Main build on Windows both here, and internally, is working as expected.

PR #3024, opened 2023-02-13
(#508) Ensure PackagesLocation exists.

corbob:
## Description Of Changes

Ensure the PackagesLocation exists as the new NuGet library doesn't handle it missing very well.

## Motivation and Context

If we get into a scenario where anything is done within NuGet.Client and the `lib` directory doesn't exist, then it will throw an exception instead of handling the missing directory. This ensures that we have a directory there.

## Testing

1. I have tested with previous develop build and deleting the `lib` directory. This encountered an error.
2. Tested with this build, and there is no error.
3. Running tests on Test Kitchen

### Operating Systems Testing

* Windows 10 22H2
* Windows Server 2016 (Test Kitchen)
* Windows Server 2019 (Test Kitchen)

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

Related to #508

### DIFF
--- a/tests/chocolatey-tests/chocolatey.Tests.ps1
+++ b/tests/chocolatey-tests/chocolatey.Tests.ps1
@@ -395,7 +395,7 @@ Describe "Ensuring Chocolatey is correctly installed" -Tag Environment, Chocolat
         BeforeAll {
             New-ChocolateyInstallSnapshot
             Remove-Item -Path $env:ChocolateyInstall/lib/ -Recurse -Force
-            $Output = Invoke-Choco
+            $Output = Invoke-Choco list --local-only
         }
 
         AfterAll {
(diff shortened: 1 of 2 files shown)


tests/chocolatey-tests/chocolatey.Tests.ps1
```diff
@@ -390,4 +390,20 @@ Describe "Ensuring Chocolatey is correctly installed" -Tag Environment, Chocolat
             $Output.Lines | Should -Contain '.NET 4.8 is not installed or may need a reboot to complete installation.'
         }
     }
+
+    Context 'Chocolatey lib directory missing' {
+        BeforeAll {
+            New-ChocolateyInstallSnapshot
+            Remove-Item -Path $env:ChocolateyInstall/lib/ -Recurse -Force
+            $Output = Invoke-Choco
```
AdmiringWorm, line 398: From memory, this is an incomplete test. The reason why an empty non-existing lib directory fails is because of NuGet itself when it tries to list packages that are installed.

Just running `choco` in this case would not trigger this happening (or at least not that I have seen).

The test should be a call to `choco list -lo` or something similar to that.

corbob: Good catch. Apparently `choco` always exits 1, so this test wouldn't actually work. I've enhanced the test, but as per Slack discussions this PR will stay draft while we work through Licensed Extension behaviour.

gep13:
@corbob just going through some PR's, and wondered what the state of this one was.  Something that we still want/need?

corbob:
I think maybe I kept it open with the intention of the  tests being brought in when we determined when/where this might need to be fixed. That being said, I don't see any reason for us to keep it open just because we might take parts from it so I'll close it.

PR #3025, opened 2023-02-14
(#508) Fix dependency resolution

gep13:
## Description Of Changes

There are times when the dependency resolution, and the attempt to find
a valid solution for package installation is too aggressive.  For
example, if you install a package and omit the installation of the
dependencies, then any follow up attempt to install a package will
fail, since it is checking for all dependencies to be met for all
installed packages.  This is not always valid, since it should only
worry about the package that is being installed/upgraded.

This commit adds a check to see if the package being installed has any
dependencies, and if it doesn't, don't check for dependencies for the
locally installed packages.  On top of this, check to also see if the
package being installed is an existing dependency on an existing
installed package.  If this is the case, then go through normal
dependency resolution.

In addition, this PR improves the regex used to extract the dependency name, by
ensuring that `-` and `.` are allowed to be included in the dependency
name.  This fixes an issue where the error output includes something
like the following:

> Unable to resolve dependency ()

i.e. where the dependency name was not output.

## Motivation and Context

During testing of Chocolatey Licensed extension, an edge case was found where a package
fails to install due to a resolution dependency problem.  Investigation was performed, and
it was found that the error came back to Chocolatey OSS, after the uplift to NuGet.Client
assemblies.

## Testing

1. Tested locally out of Visual Studio as well as ran full suite of Unit/Integration Tests (there were some failing tests during the refactoring of this issue), and with these changes in place, all tests pass
1. Full Test-Kitchen pass has been performed on this PR, and all tests pass
1. The steps of a failing Pester test from Chocolatey Licensed code base has been performed and known to pass

### Operating Systems Testing

N/A

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

Relates to #508

### DIFF
--- a/src/chocolatey/infrastructure.app/services/NugetService.cs
+++ b/src/chocolatey/infrastructure.app/services/NugetService.cs
@@ -700,7 +700,7 @@ Please see https://docs.chocolatey.org/en-us/troubleshooting for more
 
                         if (!invalidDependencyMatch.Groups["packageId"].Success)
                         {
-                            string resolvePattern = @"Unable to resolve dependency \'(?<packageId>\w+)\'";
+                            string resolvePattern = @"Unable to resolve dependency \'(?<packageId>[\w\-\.]+)\'";
                             var resolveDependencyMatch = Regex.Match(ex.Message, resolvePattern, RegexOptions.IgnoreCase);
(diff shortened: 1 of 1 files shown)


AdmiringWorm (approved):
LGTM, am a bit worried about the failing tests on test kitchen, but these doesn't seem to be because of the changes in this PR.

PR #3031, opened 2023-02-15
(maint) Update Gitreleasemanager configuration

vexx32:
## Description Of Changes

- Added the Deprecate label to GRM configuration.
- Added config to set headers in the release notes specifically for Deprecate and Bug so that their headers match our release notes format.
**NOTE:** We should cherry-pick this commit into support/1.x branch after it's merged as well.

## Motivation and Context

We're deprecating quite a few things, we should probably use a proper label for it.

Also, the auto-gen'd release notes don't quite match our release notes format, so updated the config to make that work as expected.

## Testing

Ran `./build.ps1 -Target releasenotes` from the support/1.x branch with this configuration change and verified the generated release notes match up to what we want.

### Operating Systems Testing

N/A

## Change Types Made

* [ ] Documentation changes.
* [x] Build enhancement

## Change Checklist

## Related Issue

N/A, build adjustment only

### DIFF
--- a/gitreleasemanager.yaml
+++ b/gitreleasemanager.yaml
@@ -1,5 +1,6 @@
 issue-labels-include:
 - Breaking Change
+- Deprecate
 - Feature
 - C4B Feature
 - Bug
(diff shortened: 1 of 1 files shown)


gitreleasemanager.yaml
```diff
@@ -5,6 +5,7 @@ issue-labels-include:
 - Bug
 - Enhancement
 - Documentation
+- Deprecate
```
gep13, line 8: @vexx32 the order of the entries in this part of the configuration file also controls the order of the entries on the release notes.  I think we would want this higher up in the order, no?  Perhaps beside `Breaking Change`?  Thoughts?

gep13: I have taken the liberty of moving this slightly further up in the order of things.

gep13:
@vexx32 thanks for getting this updated!

PR #3046, opened 2023-02-24
(#508) Fix dependency resolution error messages

vexx32:
## Description Of Changes

- Pull duplicated code into a shared method on NugetService
- Refactor duplicate code into a loop to make it simpler to add additional cases if we find any more error message patterns
- Relegate the warning about us developers missing an error case to a debug message; users don't care about that and they can't do anything about that
- Avoid double-logging the same error message in the case of a straightforward "Unable to resolve dependency 'thing'" from NuGet
- Avoid potentially logging the same error message potentially multiple times for no apparent reason

## Motivation and Context

There was still at least one case where the dependency resolution failure message came back without a dependency name, when there was a circular dependency chain.

Added a case for that error type and also refactored this code to reduce duplication, placing it into a protected method so that subclasses of the NugetService can reuse the error handling code a bit if needed.

One of the cases resulted in a complete duplication of the error message and some of the others were partially redundant, so also moved the warning into the main error message text since the whole thing is really an error and not a warning here.

## Testing

For testing, I used our known bad test packages. For both `choco install` and `choco upgrade`:

1. Attempt to install/upgrade `missingdependency`
2. Attempt to install/upgrade `circulardependency1`

In both cases, you should get an error of the format `Unable to resolve dependency '<DEPENDENCY NAME>'...`

3. install `missingdependency` with `--ignore-dependencies`
4. Attempt to install/upgrade `circulardependency1`

You should get the warning about an unresolved dependency constraint with the final error in the same format.

### Operating Systems Testing

Win11

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

I'm seeing some unit test failures, but they appear to be the same ones as currently are failing on `develop` already, so I don't believe I've introduced any. 🙂 

## Related Issue

Related to #508

### DIFF
--- a/src/chocolatey/infrastructure.app/services/NugetService.cs
+++ b/src/chocolatey/infrastructure.app/services/NugetService.cs
@@ -709,28 +709,11 @@ Please see https://docs.chocolatey.org/en-us/troubleshooting for more
                     }
                     catch (NuGetResolverConstraintException ex)
                     {
-                        this.Log().Warn(ex.Message);
-
-                        string constraintPattern = @"constraint: (?<packageId>\w+)\s\(";
-                        var invalidDependencyMatch = Regex.Match(ex.Message, constraintPattern, RegexOptions.IgnoreCase);
-                        var invalidDependencyName = invalidDependencyMatch.Groups["packageId"].Success ? invalidDependencyMatch.Groups["packageId"].Value : string.Empty;
-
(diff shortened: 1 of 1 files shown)


gep13 (approved):
LGTM!

gep13:
@vexx32 thanks for getting this fixed up!

PR #3050, opened 2023-02-28
(#508) Add Repository optimizations back into Chocolatey execution

gep13:
## Description Of Changes

This commit changes the find_package method to make use of a new
PackageAsync method which has been added to the ListResource within the
NuGet.Clint assemblies, which provides this optimization.

In addition, this commit makes a slight refactoring to the code where
find_package is called, and removes unnecessary calls to the
GetAllVersionsAsync method.  This work is exactly the same as the work
done in the GetMetaDataAsync method of the find_package method, and
therefore not required.

## Motivation and Context

In order for the new Chocolatey CLI to function in the same way as the
previous 1.x versions, we need to take into account the
UsePackageRepositoryOptimizations feature.  WHen this is enabled, as
few queries as necessary should be used.  For example, when querying
for an exact package id, or checking to see if a new package version is
available, we can make a single query to a source via the Packages()
API, rather than doing multiple calls to the FindPackagesById endpoint.
 This greatly reduces the number of queries that are required overall.

## Testing

1. Testing steps are detailed in this PR - https://github.com/chocolatey/NuGet.Client/pull/29

### Operating Systems Testing

- Windows 10

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

- Relates to #508
- https://app.clickup.com/t/20540031/PROJ-506

### DIFF
--- a/src/chocolatey/infrastructure.app/nuget/NugetList.cs
+++ b/src/chocolatey/infrastructure.app/nuget/NugetList.cs
@@ -239,20 +239,7 @@ namespace chocolatey.infrastructure.app.nuget
                 }
                 else
                 {
-                    if (version == null)
-                    {
-                        var versions = new SortedSet<NuGetVersion>(VersionComparer.Default);
-                        foreach (var repositoryResources in packageRepositoriesResources)
-                        {
-                            //We want all versions available across all repositories
-                            versions.AddRange((await repositoryResources.findPackageByIdResource.GetAllVersionsAsync(searchTermLower, cacheContext, nugetLogger, CancellationToken.None))
-                                .Where(a => configuration.Prerelease || !a.IsPrerelease));
-                        }
-                        version = versions.Max();
-                        if (version == null) return new List<IPackageSearchMetadata>().AsQueryable();
-                    }
-
-                    var exactPackage = find_package(searchTermLower, configuration, nugetLogger, cacheContext, packageRepositoriesResources.Select(x => x.packageMetadataResource), version);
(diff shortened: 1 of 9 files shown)


src/chocolatey/infrastructure.app/nuget/NugetList.cs
```diff
@@ -334,119 +321,62 @@ public static ISet<IPackageSearchMetadata> find_all_package_versions(string pack
...
 
-            foreach (var resource in resources)
-            {
-                var metadata = resource.GetMetadataAsync(new PackageIdentity(packageName, version), cacheContext, nugetLogger, CancellationToken.None).GetAwaiter().GetResult();
-                if (metadata != null)
-                {
-                    return metadata;
+                    foreach (var listResource in listResources)
+                    {
+                        metadataList.Add(listResource.PackageAsync(packageName, config.Prerelease, nugetLogger, CancellationToken.None).GetAwaiter().GetResult());
```
AdmiringWorm, line 347: Since we don't validate whether the returned results is `null` or not, this causes a null reference exception to happen later if the package couldn't be found.

example commands
```
choco info chocolate --source chocolatey
```
(also affects outdated if one or more packages couldn't be found).

```suggestion
                        var package = listResource.PackageAsync(packageName, config.Prerelease, nugetLogger, CancellationToken.None).GetAwaiter().GetResult();
                        if (package != null)
                        {
                            metadataList.Add(package);
                        }
```

AdmiringWorm (changes requested):
Just one thing, otherwise this looks good to me

src/chocolatey/infrastructure.app/nuget/NugetList.cs
gep13: Great catch! Will get this added!

AdmiringWorm (approved):
LGTM

AdmiringWorm:
@gep13 thanks for getting this added, great work 👍

Issue #2828, opened 2022-09-22
Replace nuget sources with complied nuget package that are used in tests


AdmiringWorm:
## Is Your Feature Request Related To A Problem? Please describe.

Currently we have the sources of the packages that we use during the run of integration and E2E tests located in the repository, as well as other places that we use as the source of truth for these source files.

## Describe The Solution. Why is it needed?

These should be replaced by already compiled binaries of the nuget package instead of having to built the packages themself when running these tests. This allows us to reduce the duplication and only have one place we can use as the source of truth for these packages.

This will require changes to the vagrant image available in the repository to set a local source to the location of the packages.

## Additional Context.

N/A

## Related Issues

N/A

AdmiringWorm:
Additional notes that should be considered when working on this issue.

Currently we are using a few packages from the community repository as well, we should at the same time update the `Invoke-Tests` powershell helper to download these packages to the same directory that we will store the other `nupkg` files in.
The tests that make use of this are currently marked as internal and will not run during open source execution (since we disable the use of the community repository during tests).

corbob:
We should also look to update the description and summary for the packages as mentioned [here](https://github.com/chocolatey/choco/pull/3051#discussion_r1132990059)

PR #3051, opened 2023-02-28
(#1092) Ensure chocolateyBeforeModify.ps1 is run for dependencies

### LINKED ISSUE #1092: chocolateyBeforeModify.ps1 is not get called for dependent packages
1. I have two packages A and B.
2. Package B is a dependency of package A.
3. Currently both packages are installed with version 1.0
4. Both packages are at version 1.1 in package repository(both packages are newer than installed ones)
5. After calling "choco upgrade A" both packages gets updated to version 1.1(which is good)
6. But chocolateyBeforeModify.ps1 gets called only for package A(I expected that chocolateyBeforeModify.ps1 should be called for package B also)

vexx32:
## Description Of Changes

- Remove manual backup of package files and run of beforeModify script for the target package
- Inject backup and beforeModify actions into the IPackageManager's Installing and Uninstalling actions to ensure that these happen for every package being changed, even for dependencies which are changing during a `choco install` operation.

## Motivation and Context

This improves the overall soundness of our package operations, and fixes an issue where updating Chocolatey licensed packages can trip over the chocolatey-agent package if it ends up being updated during processing of dependencies (for instance, by updating one of _its_ dependencies outside of its currently permitted version range, while a newer agent package is available that will work with the newer dependency version.

## Testing

1. Run the unit tests
2. Test out the failing test case with chocolatey-agent upgrade from choco 1.x (this PR build) and current stable agent, licensed extension, GUI, and gui extension to choco v2.x alpha and alpha releases of licensed components and GUI by upgrading the `chocolateygui.extension` package. With this change, the upgrade should correctly run the chocolatey-agent beforeModify script and succeed smoothly.

There are a couple unit tests I'd still like to add, but the currently-added ones cover the main priority here at the moment.

### Operating Systems Testing

Win10, Win11

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

Fixes #1092

### DIFF
--- a/src/chocolatey.tests.integration/context/dependenciesbeforemodify/hasdependencywithbeforemodify/1.0.0/hasdependencywithbeforemodify.nuspec
+++ b/src/chocolatey.tests.integration/context/dependenciesbeforemodify/hasdependencywithbeforemodify/1.0.0/hasdependencywithbeforemodify.nuspec
@@ -1,0 +1,22 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+    <metadata>
+        <id>hasdependencywithbeforemodify</id>
+        <version>1.0.0</version>
+        <title>hasdependencywithbeforemodify</title>
+        <authors>__REPLACE_AUTHORS_OF_SOFTWARE__</authors>
+        <owners>__REPLACE_YOUR_NAME__</owners>
+        <requireLicenseAcceptance>false</requireLicenseAcceptance>
+        <description>__REPLACE__</description>
+        <summary>__REPLACE__</summary>
+        <releaseNotes />
+        <copyright />
+        <tags>hasdependencywithbeforemodify admin</tags>
+        <dependencies>
+            <dependency id="isdependencywithbeforemodify" version="[1.0.0, 2.0.0.0)" />
+        </dependencies>
+    </metadata>
+  <files>
+    <file src="tools\**" target="tools" />
+  </files>
+</package>
--- a/src/chocolatey.tests.integration/context/dependenciesbeforemodify/hasdependencywithbeforemodify/2.0.0/hasdependencywithbeforemodify.nuspec
+++ b/src/chocolatey.tests.integration/context/dependenciesbeforemodify/hasdependencywithbeforemodify/2.0.0/hasdependencywithbeforemodify.nuspec
@@ -1,0 +1,22 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+    <metadata>
+        <id>hasdependencywithbeforemodify</id>
+        <version>2.0.0</version>
+        <title>hasdependencywithbeforemodify</title>
+        <authors>__REPLACE_AUTHORS_OF_SOFTWARE__</authors>
+        <owners>__REPLACE_YOUR_NAME__</owners>
+        <requireLicenseAcceptance>false</requireLicenseAcceptance>
+        <description>__REPLACE__</description>
+        <summary>__REPLACE__</summary>
+        <releaseNotes />
+        <copyright />
+        <tags>hasdependencywithbeforemodify admin</tags>
+        <dependencies>
+            <dependency id="isdependencywithbeforemodify" version="[2.0.0, 3.0.0.0)" />
+        </dependencies>
+    </metadata>
+  <files>
+    <file src="tools\**" target="tools" />
+  </files>
+</package>
--- a/src/chocolatey.tests.integration/scenarios/UpgradeScenarios.cs
+++ b/src/chocolatey.tests.integration/scenarios/UpgradeScenarios.cs
@@ -3525,5 +3525,175 @@ namespace chocolatey.tests.integration.scenarios
                 MockLogger.contains_message("pre-beforemodify-all.ps1 hook ran for upgradepackage 1.1.0", LogLevel.Info).ShouldBeFalse();
             }
         }
+
+        public class when_upgrading_a_package_with_beforeModify_script_with_dependencies_with_beforeModify_scripts_and_hooks : ScenariosBase
+        {
+            private const string TargetPackageName = "hasdependencywithbeforemodify";
+            private const string DependencyName = "isdependencywithbeforemodify";
+
+            public override void Context()
+            {
+                base.Context();
+
+                Scenario.add_packages_to_source_location(Configuration, "{0}.*".format_with(TargetPackageName) + Constants.PackageExtension);
+                Scenario.add_packages_to_source_location(Configuration, "{0}.*".format_with(DependencyName) + Constants.PackageExtension);
+                Scenario.add_packages_to_source_location(Configuration, "scriptpackage.hook" + "*" + Constants.PackageExtension);
+                Scenario.install_package(Configuration, DependencyName, "1.0.0");
+                Scenario.install_package(Configuration, TargetPackageName, "1.0.0");
+                Scenario.install_package(Configuration, "scriptpackage.hook", "1.0.0");
+
+                Configuration.PackageNames = Configuration.Input = TargetPackageName;
+            }
+
+            public override void Because()
+            {
+                Results = Service.upgrade_run(Configuration);
+            }
+
+            [Fact]
+            public void should_upgrade_the_package()
+            {
+                var packageFile = Path.Combine(Scenario.get_top_level(), "lib", TargetPackageName, "{0}.nupkg".format_with(TargetPackageName));
+                var package = new OptimizedZipPackage(packageFile);
+                package.Version.Version.to_string().ShouldEqual("2.0.0.0");
+            }
+
+            [Fact]
+            public void should_upgrade_the_minimum_version_dependency()
+            {
+                var packageFile = Path.Combine(Scenario.get_top_level(), "lib", DependencyName, "{0}.nupkg".format_with(DependencyName));
+                var package = new OptimizedZipPackage(packageFile);
+                package.Version.Version.to_string().ShouldEqual("2.0.0.0");
+            }
+
+            [Fact]
+            public void should_contain_a_message_that_everything_upgraded_successfully()
+            {
+                MockLogger.contains_message("upgraded 2/2", LogLevel.Warn).ShouldBeTrue();
+            }
+
+            [Fact]
+            [WindowsOnly]
+            [Platform(Exclude = "Mono")]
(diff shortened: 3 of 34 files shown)


src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -632,6 +651,21 @@ public virtual void remove_rollback_directory_if_exists(string packageName)
                 uninstallSuccessAction: null,
                 addUninstallHandler: false);
 
+            EventHandler<PackageOperationEventArgs> beforeModifyHandler = (sender, eventArgs) =>
+            {
+                // Soundness: any code called here must avoid calling packageManager methods for
+                // InstallPackage() and UninstallPackage(), or in some other way handle the
```
corbob, line 657: There's a few places that we have this comment, is there anything we can do to check this doesn't happen?

src/chocolatey.tests.integration/scenarios/UpgradeScenarios.cs
```diff
@@ -3525,5 +3525,105 @@ public void should_not_have_executed_beforemodify_hook_script_for_upgrade_versio
...
+            private const string DependencyName = "isdependencywithbeforemodify";
+
+            public override void Context()
+            {
+                base.Context();
+
+                Scenario.add_packages_to_source_location(Configuration, "{0}.*".format_with(TargetPackageName) + Constants.PackageExtension);
+                Scenario.add_packages_to_source_location(Configuration, "{0}.*".format_with(DependencyName) + Constants.PackageExtension);
+                Scenario.install_package(Configuration, DependencyName, "1.0.0");
+                Scenario.install_package(Configuration, TargetPackageName, "1.0.0");
```
corbob, line 3541: Would it be an idea to extract these version numbers into a variable of the class in case they change?

(This comment would apply equally to the rest of the version numbers in the tests here)

src/chocolatey.tests.integration/context/dependenciesbeforemodify/hasdependencywithbeforemodify/1.0.0/hasdependencywithbeforemodify.nuspec
```diff
@@ -0,0 +1,22 @@
...
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+    <metadata>
+        <id>hasdependencywithbeforemodify</id>
+        <version>1.0.0</version>
+        <title>hasdependencywithbeforemodify</title>
+        <authors>__REPLACE_AUTHORS_OF_SOFTWARE__</authors>
+        <owners>__REPLACE_YOUR_NAME__</owners>
+        <requireLicenseAcceptance>false</requireLicenseAcceptance>
+        <description>__REPLACE__</description>
+        <summary>__REPLACE__</summary>
```
corbob, line 11: Should we add the contents of the `purpose.txt` document to the description, summary, or both? I realize a lot of our test packages don't have this, but would be helpful when looking at the packages in a repository to understand the purpose.

corbob (commented):
Just a few questions, nothing that necessarily requires changes though.

src/chocolatey/infrastructure.app/services/NugetService.cs
vexx32: I'm open to ideas but I'm not sure there's a great way to ensure this. Also, in at least one case (though in code I haven't touched here) I think we actively and knowingly do call UninstallPackage from the PackageUninstalled(/ing?) handler, though at least there we do guard against the recursion by checking how many times it's been called.

So... like yeah it'd be good to have, but I don't know that there's a good way to go about it.

src/chocolatey.tests.integration/context/dependenciesbeforemodify/hasdependencywithbeforemodify/1.0.0/hasdependencywithbeforemodify.nuspec
vexx32: We could do? TBH if we wanna do this I might suggest we just ditch purpose.txt files entirely and put it in the package summary / description instead.

If we wanna do that I'd suggest we do them all in one PR though, get 'em all done in one sweep.

src/chocolatey.tests.integration/scenarios/UpgradeScenarios.cs
vexx32: In general, unless the test structure is changing, the version numbers shouldn't need to change. The only reason the package IDs got put in properties is because I had to reference them like 10 times each lol

src/chocolatey.tests.integration/context/dependenciesbeforemodify/hasdependencywithbeforemodify/1.0.0/hasdependencywithbeforemodify.nuspec
corbob: That makes sense. I'll mark this conversation resolved and add a note to #2828 since it feels like that'd be where we do it.

src/chocolatey/infrastructure.app/services/NugetService.cs
corbob: The only way I could think to maybe do it would be looking at the call stack or something like that, and not sure it'd be worth it in most instances due to performance? Maybe some unit tests that does it?

I'll leave this conversation open for @gep13 to weigh in on if he has any opinions or ideas about it.

vexx32: I'm not sure if you can get _that_ kind of funky in C# land even with Reflection, but if you can I don't think it's a good idea to do it here 😂

src/chocolatey.tests.integration/context/dependenciesbeforemodify/hasdependencywithbeforemodify/1.0.0/hasdependencywithbeforemodify.nuspec
```diff
@@ -0,0 +1,22 @@
...
+        <releaseNotes />
+        <copyright />
+        <tags>hasdependencywithbeforemodify admin</tags>
+        <dependencies>
+            <dependency id="isdependencywithbeforemodify" version="[1.0.0, 2.0.0.0)" />
```
gep13, line 16: Was the four part version number here on purpose?

src/chocolatey.tests.integration/context/dependenciesbeforemodify/hasdependencywithbeforemodify/2.0.0/hasdependencywithbeforemodify.nuspec
```diff
@@ -0,0 +1,22 @@
...
+        <authors>__REPLACE_AUTHORS_OF_SOFTWARE__</authors>
+        <owners>__REPLACE_YOUR_NAME__</owners>
+        <requireLicenseAcceptance>false</requireLicenseAcceptance>
+        <description>__REPLACE__</description>
+        <summary>__REPLACE__</summary>
+        <releaseNotes />
+        <copyright />
+        <tags>hasdependencywithbeforemodify admin</tags>
+        <dependencies>
+            <dependency id="isdependencywithbeforemodify" version="[2.0.0, 3.0.0.0)" />
```
gep13, line 16: Similar question to before... why the four part version number?

src/chocolatey.tests.integration/scenarios/UpgradeScenarios.cs
```diff
@@ -3525,5 +3525,105 @@ public void should_not_have_executed_beforemodify_hook_script_for_upgrade_versio
...
+            {
+                Results = Service.upgrade_run(Configuration);
+            }
+
+            [Fact]
+            public void should_upgrade_the_package()
+            {
+                var packageFile = Path.Combine(Scenario.get_top_level(), "lib", TargetPackageName, "{0}.nupkg".format_with(TargetPackageName));
+                var package = new OptimizedZipPackage(packageFile);
+                package.Version.Version.to_string().ShouldEqual("2.0.0.0");
```
gep13, line 3556: Why is this a four part version number?

src/chocolatey.tests.integration/scenarios/UpgradeScenarios.cs
```diff
@@ -3525,5 +3525,105 @@ public void should_not_have_executed_beforemodify_hook_script_for_upgrade_versio
...
+            }
+
+            [Fact]
+            public void should_upgrade_the_minimum_version_dependency()
+            {
+                var packageFile = Path.Combine(Scenario.get_top_level(), "lib", DependencyName, "{0}.nupkg".format_with(DependencyName));
+                var package = new OptimizedZipPackage(packageFile);
+                package.Version.Version.to_string().ShouldEqual("2.0.0.0");
```
gep13, line 3564: Why is this a four part version number?

src/chocolatey/infrastructure.app/services/NugetService.cs
gep13: To prevent usage of these methods, we could add them to the BannedAPI's list, and provide an exemption for the current uses of it.  That way, any new usage of it will be flagged, and will need to be overridden.  The only slight issue here is that I believe we have only introduced the BannedAPI's on the develop branch, so may not be able to use it here.

gep13 (changes requested):

gep13:
@vexx32 for the tests that you are going to add to this, do we need to add any additional tests for hook packages to ensure that they continue to work as expected, or do the current tests cover the changes that you are making in this PR?

src/chocolatey.tests.integration/context/dependenciesbeforemodify/hasdependencywithbeforemodify/1.0.0/hasdependencywithbeforemodify.nuspec
vexx32: I was copying existing packages, so yes, but I don't know why they were originally written this way. (see: `hasdependency`, for example)

src/chocolatey.tests.integration/scenarios/UpgradeScenarios.cs
vexx32: `Version.Version` is a `System.Version` and will give a four-part version. I'm unsure exactly why this route was chosen, I simply copied the existing patterns here to avoid reinventing the wheel ^^.

vexx32:
> do we need to add any additional tests for hook packages to ensure that they continue to work as expected, or do the current tests cover the changes that you are making in this PR?

I'll double check the existing tests for that, but I suspect we'll want to add an extra test for that to make sure they work as expected.

gep13:
The failing Ubuntu build here is an issue with NuGet Restore, and nothing to do with the changes in this PR.

gep13:
@vexx32 thank you very much for getting this done!

PR #3052, opened 2023-03-01
(maint) Use specific version for dependencies

gep13:
## Description Of Changes

This commit adds a specific version number in each dependency in each of the nuspec files.  This is necessary, since moving into new NuGet.Client assemblies, the version number is a required field in the dependency element, and we have found issues where repository managers don't correctly show/find the dependencies on a V3 feed, when a version isn't available in this element.  There may be other work that needs to be done in Chocolatey to account for this new requirement, but updating all our internal testing packages to have a version number for each dependency will ensure that some automated tests work as expected.

## Motivation and Context

Some end to end tests are failing when using the V3 feed, since not all dependencies are known, and therefore fail to be installed.

## Testing

Once pushed, these new packages will be part of the Test-Kitchen execution, and will be tested there.

### Operating Systems Testing

N/A

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

N/A

### DIFF
--- a/src/chocolatey.tests.integration/context/dependencies/hasdependency/2.0.1/hasdependency.nuspec
+++ b/src/chocolatey.tests.integration/context/dependencies/hasdependency/2.0.1/hasdependency.nuspec
@@ -13,7 +13,7 @@
     <copyright />
     <tags>hasdependency admin</tags>
     <dependencies>
-      <dependency id="isdependency" />
+      <dependency id="isdependency" version="1.0.0" />
     </dependencies>
   </metadata>
   <files>
--- a/tests/chocolatey-tests/commands/testpackages/chocolatey-dummy-package/1.0.0/chocolatey-dummy-package.nuspec
+++ b/tests/chocolatey-tests/commands/testpackages/chocolatey-dummy-package/1.0.0/chocolatey-dummy-package.nuspec
@@ -12,7 +12,7 @@
     <summary>This is a purposely empty dummy package for testing choco push functionality</summary>
     <description>This is a purposely empty dummy package for testing choco push functionality</description>
     <dependencies>
-      <dependency id="chocolatey" />
+      <dependency id="chocolatey" version="0.10.15" />
     </dependencies>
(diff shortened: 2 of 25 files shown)


src/chocolatey.tests.integration/context/dependencies/hasdependency/2.0.1/hasdependency.nuspec
```diff
@@ -13,7 +13,7 @@
     <copyright />
     <tags>hasdependency admin</tags>
     <dependencies>
-      <dependency id="isdependency" />
+      <dependency id="isdependency" version="2.1.0" />
```
AdmiringWorm, line 16: This doesn't look right, it should depend on the earliest version to make it equivalent to the current dependency range. I believe the earliest version is `1.0.0` of this package.

tests/chocolatey-tests/commands/testpackages/chocolatey-dummy-package/1.0.0/chocolatey-dummy-package.nuspec
```diff
@@ -12,7 +12,7 @@
     <summary>This is a purposely empty dummy package for testing choco push functionality</summary>
     <description>This is a purposely empty dummy package for testing choco push functionality</description>
     <dependencies>
-      <dependency id="chocolatey" />
+      <dependency id="chocolatey" version="1.0.1" />
```
AdmiringWorm, line 15: Again we should probably set the dependency range to be the minimum version of `chocolatey` that is available, not sure which version this would be but it definitely would be in a pre-v1 range at least.

AdmiringWorm (changes requested):
A few things to fix up with the dependency ranges specified. The comment on the Chocolatey dependency is relevant for all instances of that dependency.

AdmiringWorm (approved):
LGTM

AdmiringWorm:
@gep13 thanks for getting those missing minimum versions added 👍

PR #3065, opened 2023-03-14
(#1092) Incorporate beforemodify changes into develop branch

vexx32:
## Description Of Changes

- Cherry-pick ISourceRunner changes and amend to match develop-branch changes to ISourceRunner
- Cherry-pick tests added in #3051 with minor amendments to work in the develop branch
- Manually reapply beforeModify changes from #3051 and ensure the new methods are being used in the NugetService with the newer NuGet code patterns.

I may have snuck in a few extra bits where I obsoleted some method overloads that probably should not be `public` and/or had arguments they weren't using and replaced them with a call to a new overload that is `protected` where possible and/or doesn't ask for arguments it doesn't need. Sue me, those bothered me a lot. I'm not sure we strictly _need_ to obsolete them first given how much else we're also breaking, but it wouldn't majorly hurt at this point to do so.

## Motivation and Context

This is being pulled in so that we can ensure we keep feature parity with the v1.x branch as much as possible.

## Testing

1. Ran the added integration tests, no failures. (Feel free to rip out the commit making the code changes if you'd like to see how the tests do prior to this code change; there are a few failures to compare.)
2. Ran the entire unit/integration tests, no different failures than there currently seem to be on develop. (Locally, I get 101 tests failing both on this branch and develop. Cory thinks it's because I don't give VS admin permissions, but _who does that?_ 😁)

### Operating Systems Testing

Tested on win11

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

- #1092 
- #3051

### DIFF
--- a/src/chocolatey.tests.integration/context/dependenciesbeforemodify/hasdependencywithbeforemodify/1.0.0/hasdependencywithbeforemodify.nuspec
+++ b/src/chocolatey.tests.integration/context/dependenciesbeforemodify/hasdependencywithbeforemodify/1.0.0/hasdependencywithbeforemodify.nuspec
@@ -1,0 +1,22 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+    <metadata>
+        <id>hasdependencywithbeforemodify</id>
+        <version>1.0.0</version>
+        <title>hasdependencywithbeforemodify</title>
+        <authors>__REPLACE_AUTHORS_OF_SOFTWARE__</authors>
+        <owners>__REPLACE_YOUR_NAME__</owners>
+        <requireLicenseAcceptance>false</requireLicenseAcceptance>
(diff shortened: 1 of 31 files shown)


TheCakeIsNaOH (commented):
Overall it looks good. 

For the tests, would it be worthwhile to check that the beforemodify scripts are not run more than once? 

And to have a test that the beforemodify script of a new dependency is not run during upgrade (maybe this is already covered). So if `packageA` v1.0.0 has no dependencies, but v2.0.0 has a dependency on `packageB` (which is not installed), and `packageB` has a beforemodify script.

vexx32:
I believe I have tests verifying that the beforeModify scripts are only run for the currently-installed version and not the version that is installed. If we want to explicitly add a test that makes sure we aren't running them multiple times we can do that as well.

The latter suggestion is definitely worth adding, though, good idea. 🙂

gep13 (approved):
LGTM!

gep13:
@vexx32 thanks for getting this brought over from the support branch!

PR #3073, opened 2023-03-17
(#2788) Remove side-by-side installs

### LINKED ISSUE #2788: Remove side-by-side installs

Side-by-side installs [were deprecated](#2787 ) so we should now remove the functionality from the collection for version 2.0 or later of Chocolatey CLI 

We should look to Puppet, Ansible and DSC providers for this to ensure they are aware of the deprecation.

* Puppet: [Does not support side-by-side installations](https://forge.puppet.com/modules/puppetlabs/chocolatey#known-issues).
* Ansible: Issue raised in the [Chocolatey Ansible Collection](https://github.com/chocolatey/chocolatey-ansible/issues/95)
* cChoco: [Doesn't support side-by-side installations](https://github.com/chocolatey/cChoco/issues/65).
* Chocolatey DSC Module: [Issue raised](https://github.com/chocolatey-community/Chocolatey/issues/61) in the repository.
* Chef: Unable to determine whether the recipe support side-by-side installs or not so raised an issue [here](https://github.com/chocolatey-community/chocolatey-cookbook/issues/153) and [here](https://github.com/chef/chef/issues/13368).

N/A

## Additional Context.

N/A

## Related Issues

- Issue #2787

vexx32:
## Description Of Changes

- Remove side by side installs
- Remove all the other stuff that was using / referencing it
- Add a warning for the removed param so folks don't expect it to work and find it doesn't at all

## Motivation and Context

Removing deprecated features.

## Testing

1. Run unit tests
2. Run Invoke-Tests.ps1
3. Do a `install $something --allow-multiple` run and note:
   - A warning is emitted saying the param isn't supported
   - The installed package is _not_ placed in a versioned folder

### Operating Systems Testing

Win11

## Change Types Made

* [x] Breaking change (fix or feature that could cause existing functionality to change).
* [ ] Documentation changes.

## Change Checklist

## Related Issue

Fixes #2788


src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -1818,30 +1784,6 @@ protected virtual void ensure_package_files_have_compatible_attributes(Chocolate
             _filesService.ensure_compatible_file_attributes(installDirectory, config);
         }
 
-        [Obsolete("This overload is obsolete and will be removed in a future version.")]
-        public virtual void rename_legacy_package_version(ChocolateyConfiguration config, IPackageMetadata installedPackage, ChocolateyPackageInformation pkgInfo)
-            => normalize_package_legacy_folder_name(config, installedPackage, pkgInfo);
-
-        protected virtual void normalize_package_legacy_folder_name(ChocolateyConfiguration config, IPackageMetadata installedPackage, ChocolateyPackageInformation pkgInfo)
-        {
```
TheCakeIsNaOH, line 1826: Don't we still need this? For people upgrading directly from very old versions.

src/chocolatey/infrastructure.app/configuration/PackagesConfigFilePackageSetting.cs
```diff
@@ -50,9 +50,6 @@ public sealed class PackagesConfigFilePackageSetting
         [XmlAttribute(AttributeName = "forceX86")]
         public bool ForceX86 { get; set; }
 
-        [XmlAttribute(AttributeName = "allowMultipleVersions")]
```
TheCakeIsNaOH, line 53: Is it worth keeping this around, but emit a warning that the attribute is no longer supported?

corbob (changes requested):

src/chocolatey/infrastructure.app/services/NugetService.cs
vexx32: IMO I don't think it's worth the hassle to keep this. I did initially try to keep this in play, but there were a lot of weird edge case issues, and it's not something we can reliably test from what I was able to tell when messing with this. Our existing tests for this actually relied on being able to use the SideBySide install functionality in the first place to "pretend" the legacy package install.

Given that, I don't think it's sensible to keep this given we're breaking side-by-side functionality completely and deliberately. 

@gep13?

src/chocolatey/infrastructure.app/commands/ChocolateyCommandBase.cs
```diff
@@ -0,0 +1,47 @@
...
+            if (!unparsedOptions.or_empty_list_if_null().Any() || !removedOptions.or_empty_list_if_null().Any())
+            {
+                return;
+            }
+
+            foreach (var removed in removedOptions)
+            {
+                if (unparsedOptions.Contains(removed))
+                {
+                    this.Log().Warn("The {0} option is no longer supported.", removed);
```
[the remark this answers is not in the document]
vexx32: Hmm, that's a good point.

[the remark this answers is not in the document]
vexx32: This class itself won't have the config to look at, but we can check before calling it

src/chocolatey/infrastructure.app/configuration/PackagesConfigFilePackageSetting.cs
TheCakeIsNaOH: To not break the schema of Chocolatey `packages.config` files. The deprecation warnings only mentioned the actual side by side installs, so I think a user could have an expectation of having existing `packages.config` files still work, while ignoring/warning on that attribute.

src/chocolatey/infrastructure.app/commands/ChocolateyCommandBase.cs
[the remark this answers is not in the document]
corbob: As discussed on Slack, I've passed in a bool (`RegularOutput`) that is just the property from the config object, so that we can output Warn if regular output and Debug if not.

src/chocolatey/infrastructure.app/configuration/PackagesConfigFilePackageSetting.cs
corbob: Assuming that the reading of the `packages.config` happens the same way that the config file and some Intune files happens, then the presence of the attribute in the file will not cause any issues as long as it's a valid xml document. This of course is something that we should test out of course (once I've gotten a build off of this PR I'll give it a try and can report back.)

corbob: Yeah, as long as the xml is valid, it loads just fine, even if it has attributes and elements that aren't part of the class.



As with the other comment, I'd leave it up to @gep13 to determine if we should keep it or not.

pauby: Unless there is a technical reason this should be kept, we need to remove it as it's no longer needed.

src/chocolatey/infrastructure.app/services/NugetService.cs
TheCakeIsNaOH: > Our existing tests for this actually relied on being able to use the SideBySide install functionality in the first place to "pretend" the legacy package install.

Can't this be switched to using the same method used for testing uninstalls of multiple side by side versions? Unless I misremembering the code.

src/chocolatey/infrastructure.app/configuration/PackagesConfigFilePackageSetting.cs
TheCakeIsNaOH: I would agree that it can be removed completely since then the attribute would be ignored.

src/chocolatey/infrastructure.app/commands/ChocolateyCommandBase.cs
gep13, line 1: What happened with the line-endings on this file?

src/chocolatey/infrastructure.app/commands/ChocolateyCommandBase.cs
```diff
@@ -0,0 +1,56 @@
...
+                return;
+            }
+
+            foreach (var removed in removedOptions)
+            {
+                if (unparsedOptions.Contains(removed))
+                {
+                    var message = "The {0} option is no longer supported.".format_with(removed);
+
+                    if (RegularOutput)
```
gep13, line 44:
```suggestion
                    if (regularOutput)

```

src/chocolatey/infrastructure.app/commands/ChocolateyInstallCommand.cs
```diff
@@ -208,7 +208,7 @@ public virtual void handle_additional_argument_parsing(IList<string> unparsedArg
 
             if (configuration.RegularOutput)
             {
-                warn_for_removed_options(unparsedArguments.Where(arg => arg.StartsWith("-")), _removedOptions);
+                warn_for_removed_options(unparsedArguments.Where(arg => arg.StartsWith("-")), _removedOptions, configuration.RegularOutput);
```
gep13, line 211: Why are we passing in this parameter to the `warn_for_removed_options`? Why not simply not call this method at all?

src/chocolatey.tests.integration/Scenario.cs
```diff
@@ -209,6 +209,11 @@ public static void create_directory(string directoryPath)
             _fileSystem.create_directory(directoryPath);
         }
 
+        public static void move_directory(string path, string newPath)
+        {
+            _fileSystem.move_directory(path, newPath);
+        }
```
gep13, line 215: What was the reason for adding this? It doesn't seem to get called from anywhere.

tests/chocolatey-tests/commands/choco-removed.Tests.ps1
```diff
@@ -134,4 +134,41 @@ exit $command.Count
             $Output.Lines | Should -Contain "Unknown command --remove. Setting to list."
         }
     }
+
+    Context 'Ensure --allow-multiple removed from Chocolatey'  -Tag InstallCommand, UpgradeCommand, UninstallCommand, AllowMultiple -Foreach @(
```
gep13, line 138:
```suggestion
    Context 'Ensure --allow-multiple removed from Chocolatey' -Tag InstallCommand, UpgradeCommand, UninstallCommand, AllowMultiple -Foreach @(

```

src/chocolatey/infrastructure.app/services/NugetService.cs
gep13: I am inclined to say that we continue with removing this functionality.  We have done everything that we can to let folks know about the upcoming deprecation of side-by-side, and providing an upgrade path from a very old version of Chocolatey is not something that we support.  The only thing that we would support is the previous major version, which in this case would be 1.x, where this method was already in place, and any required normalization has already been done.

src/chocolatey/infrastructure.app/configuration/PackagesConfigFilePackageSetting.cs
gep13: I am ok with this being removed.

gep13 (changes requested):

gep13 (approved):
LGTM!

src/chocolatey/infrastructure.app/commands/ChocolateyCommandBase.cs
gep13: @corbob  I have moved forward with the removal of this change, happy to chat about it once you are in.

gep13:
@vexx32 thank you for getting this fixed up!

src/chocolatey.tests.integration/Scenario.cs
vexx32: Was added as I was trying to make some tests to verify the "legacy" behaviour would still work, no longer needed if we're not keeping that.

PR #3084, opened 2023-03-21
(#2902)(#2585) Update TLS protocol handling

### LINKED ISSUE #2585: Improve handling of old TLS version warning
https://github.com/chocolatey/choco/blob/d8ca7bad2ad64538a0c1026a9317700dbb7f03db/src/chocolatey/infrastructure/registration/SecurityProtocol.cs#L44

### LINKED ISSUE #2902: Update handling of HTTPS security protocols

Now that Chocolatey CLI is moving to .Net 4.8, there is a redundant fallback and warning message in `SecurityProtocol`. Since .Net 4.8 always has TLS v1.2 available, there is no need to continue to have this fallback and warning.

https://github.com/chocolatey/choco/blob/edde97906053fbb3768d18300c2ed3f3716c4a98/src/chocolatey/infrastructure/registration/SecurityProtocol.cs#L43-L56

At minimum, this specific warning/fallback should be removed.

But it would be better to have a re-evaluation of the status of SSL/TLS in Chocolatey CLI, and best practices for .Net 4.8 programs. 

This document recommends that the TLS version should not be explicitly set by programs, and use the OS defaults.
https://learn.microsoft.com/en-us/dotnet/framework/network-programming/tls

However, from what I understand, this may mean that on Windows 7, depending on updates/registry settings then .Net apps may not have TLS 1.2 enabled by default. As per:
https://learn.microsoft.com/en-us/windows/win32/secauthn/protocols-in-tls-ssl--schannel-ssp-#tls-protocol-version-support

So it may make sense to enable TLS 1.1/1.2 if the Windows build/version corresponds to Windows 7/Server 2008 R2, and use the OS default if the OS build is newer?

## Additional Context.

N/A

## Related Issues

- Issue #2738

TheCakeIsNaOH:
## Description Of Changes

This updates the TLS protocols that Chocolatey sets now that Chocolatey is built with .NET 4.8.

First, this removes the fallback and warning about being unable to set TLS 1.2, as .NET 4.8 will always have TLS 1.2 available. Second, it uses the TLS 1.2 definition directly available in the system libraries. Third, it removes the provideWarning bool in the method definition, as the warning is no longer applicable.

Finally, an OS version check is added, so the system default will be used on Windows 8/Server 2012 and newer, as per Microsoft guidance, but explicitly enabling TLS 1.2 on older operating systems. See code comments for links to documentation.

## Motivation and Context

Keeping current with TLS protocols available is important for security reasons, for example this allows TLS v1.3 on Windows 11 / Server 2022

## Testing

TODO

### Operating Systems Testing
- Windows 10 22H2

## Change Types Made

* [x] Breaking change (fix or feature that could cause existing functionality to change).
* [ ] Documentation changes.

## Change Checklist

## Related Issue

Fixes #2902
Fixes #2585 by removing the need for the warning

### DIFF
--- a/src/chocolatey/infrastructure.app/runners/GenericRunner.cs
+++ b/src/chocolatey/infrastructure.app/runners/GenericRunner.cs
@@ -146,7 +146,7 @@ Chocolatey is not an official build (bypassed with --allow-unofficial).
             }
 
             FailOnMissingOrInvalidLicenseIfFeatureSet(config);
-            SecurityProtocol.SetProtocol(provideWarning: true);
+            SecurityProtocol.SetProtocol(config);
             EventManager.Publish(new PreRunMessage(config));
 
             try
@@ -230,7 +230,7 @@ Chocolatey is not an official build (bypassed with --allow-unofficial).
             }
 
             FailOnMissingOrInvalidLicenseIfFeatureSet(config);
-            SecurityProtocol.SetProtocol(provideWarning: true);
+            SecurityProtocol.SetProtocol(config);
             EventManager.Publish(new PreRunMessage(config));
 
             try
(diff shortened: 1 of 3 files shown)


TheCakeIsNaOH:
I'm open to discussion on the new behavior.

vexx32:
@TheCakeIsNaOH thanks for your work on this!

After reviewing the documentation the .NET team provides and talking with the team, we've decided we should go a bit of a different route here. I've opened #3123 to make this change instead, so I'll close this one off for now.

PR #3126, opened 2023-04-20
(tests) Skip some Pester tests if the repository endpoint is a v3 endpoint

corbob:
## Description Of Changes

Mark some pester tests to skip if they're being tested against a v3 endpoint and the Licensed Extension is in play.

## Motivation and Context

For some reason the results coming back from our v3 endpoint are not consistent, and we're seeing failures in these tests when licensed extension is tested against our v3 endpoint.

## Testing

1. Ran Test Kitchen in Team City against both Chocolatey CLI and Chocolatey Licensed Extension
2. Noted these tests failing only on Licensed Extension with output indicating some of the packages not coming back from the repository server.
3. Added the skip conditions
4. Ran Test Kitchen in Team City against both Chocolatey CLI and Chocolatey Licensed Extension

### Operating Systems Testing

* Windows Server 2016
* Windows Server 2019

## Change Types Made

* [ ] Documentation changes.
* [x] Pester test changes.

## Change Checklist

## Related Issue

* PROJ-518

### DIFF
--- a/tests/chocolatey-tests/commands/choco-search.Tests.ps1
+++ b/tests/chocolatey-tests/commands/choco-search.Tests.ps1
@@ -1,4 +1,4 @@
-﻿<#
+<#
     .Synopsis
         Tests for `choco search` and aliases
 
@@ -18,6 +18,7 @@ Import-Module helpers/common-helpers
 Describe "choco <_>" -ForEach $Command -Tag Chocolatey, SearchCommand, FindCommand {
     BeforeDiscovery {
         $licensedProxyFixed = Test-PackageIsEqualOrHigher 'chocolatey.extension' 2.2.0-beta -AllowMissingPackage
+        $hasEnabledV3Feed = Test-HasNuGetV3Source
     }
 
     BeforeAll {
@@ -35,7 +36,8 @@ Describe "choco <_>" -ForEach $Command -Tag Chocolatey, SearchCommand, FindComma
         Remove-ChocolateyTestInstall
     }
 
-    Context "Searching packages with no filter (Happy Path)" {
+    # Skip when searching against a v3 source as our current source is not returning consistent results.
+    Context "Searching packages with no filter (Happy Path)" -Skip:$hasEnabledV3Feed {
         BeforeAll {
             $Output = Invoke-Choco $_
         }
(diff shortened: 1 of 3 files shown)


tests/chocolatey-tests/commands/choco-search.Tests.ps1
```diff
@@ -35,7 +36,8 @@ Describe "choco <_>" -ForEach $Command -Tag Chocolatey, SearchCommand, FindComma
         Remove-ChocolateyTestInstall
     }
 
-    Context "Searching packages with no filter (Happy Path)" {
+    # Skip on Licensed Extension when searching against a v3 source as our source is not returning consistent results.
+    Context "Searching packages with no filter (Happy Path)" -Skip:$hasV3Feed {
```
vexx32, line 40: If I'm understanding correctly, this means _all_ of these tests will be skipped completely if any one configured source has a v3 feed url?

Can we disable the v3 feed url and use other urls? Or is that not feasible in our current configuration?

tests/chocolatey-tests/commands/choco-search.Tests.ps1
```diff
@@ -35,7 +36,8 @@ Describe "choco <_>" -ForEach $Command -Tag Chocolatey, SearchCommand, FindComma
         Remove-ChocolateyTestInstall
     }
 
-    Context "Searching packages with no filter (Happy Path)" {
+    # Skip on Licensed Extension when searching against a v3 source as our source is not returning consistent results.
```
vexx32, line 39: I know we already spoke about this, but these comments need updated as we're doing this in CLI and not just CLE :)

tests/chocolatey-tests/commands/choco-search.Tests.ps1
corbob: In the Test Kitchen environment as it exists today, we don't really have more than one test feed enabled during the discovery phase, and we also currently only test against either the v2 or the v3 feed, but not against both in the same run.

Just to be a little more explicit about what we're skipping, I've updated the variable name to be `$hasEnabledV3Feed` since that's what we're checking against. Theoretically the setup feed that we pull Chocolatey and other such packages from could be v3 while the test is against v2.

tests/chocolatey-tests/commands/choco-search.Tests.ps1
corbob: I've updated these comments and rebased everything onto the head of the release branch 😄

vexx32 (approved):

PR #3128, opened 2023-04-20
(#1720) Parse multi-protocol proxy configurations

### LINKED ISSUE #1720: Chocolatey does not override NuGet proxy with its own
### What You Are Seeing?

* System proxy: configured but disabled

image

* NuGet proxy: configured in `%appdata%\NuGet\NuGet.config`

image

Scenario:

My employer recently decommissioned their proxy server (it no longer accepts HTTP requests, but its DNS record still exists). Due to various issues with tools and proxies in the past, I have hardcoded the proxy for various tools, one of these being NuGet.

The end result is that when I run any Cholatey command that invokes NuGet to fetch package sources, it fails with `<package-name> was not found with the source(s) listed`:

```sh
> choco upgrade chocolatey
Chocolatey v0.10.11
Upgrading the following packages:
chocolatey
By upgrading you accept licenses for the packages.
[NuGet] An error occurred while loading packages from 'https://chocolatey.org/api/v2/': Unable to connect to the remote server

chocolatey was not found with the source(s) listed.
 If you specified a particular version and are receiving this message, it is possible that the package name exists but the version does not.
 Version: ""; Source(s): "https://chocolatey.org/api/v2/"

Chocolatey upgraded 0/1 packages.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).

Warnings:
 - chocolatey - chocolatey was not found with the source(s) listed.
```

If I run choco with the `-dv` option, the only line about proxy is `Proxy.BypassOnLocal='True'`. Due to this, I assumed that choco (and hence NuGet) was using no proxy, and hence looked elsewhere. Additionally, if I run `choco upgrade chocolatey --proxy=""` (i.e. force no proxy), the same issue occurs.

Removing the proxy in NuGet.config makes everything work, but it took me a while to figure this out, and as a result I suggest the following:

1. Always force NuGet to use the proxy that Chocolatey is using. This is problematic because apparently NuGet is stupid and doesn't allow the proxy to be specified per invocation (i.e. via the command-line), only via system-wide `HTTP_PROXY` etc. environment variables. Possible solution is to save the current `HTTP_PROXY` before invoking NuGet, set it to the choco value, then restore it after the NuGet invocation completes.
2. Output more information about the proxy NuGet (and not choco) is using in the diagnostic log. (This would be unnecessary if they were always using the same proxy as suggested in the previous point.)

corbob:
## Description Of Changes

Update Proxy configuration to parse the setting if it's specifying multiple protocols.

## Motivation and Context

Windows has the concept of protocol specific proxies, as such it can store the proxy information in the registry as `http=server:port;https=server:port;ftp=server:port;socks=server:port`. We need to parse this instead of just blindly taking it.

## Testing

1. Ran build locally
2. Ran build in Team City.
3. Ran Test Kitchen on Team City.

### Operating Systems Testing

- Windows Server 2016
- Windows Server 2019

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

* Fixes #1720 
* PROJ-583

### DIFF
--- a/src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
+++ b/src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
@@ -476,28 +476,35 @@ namespace chocolatey.infrastructure.app.builders
             var registryService = container.GetInstance<IRegistryService>();
             var internetSettingsRegKey = registryService.GetKey(RegistryHive.CurrentUser, "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Internet Settings");
 
-            if (internetSettingsRegKey.GetValue("ProxyEnable").ToStringSafe().IsEqualTo("1"))
+            if (!internetSettingsRegKey.GetValue("ProxyEnable").ToStringSafe().IsEqualTo("1"))
             {
-                var proxySetting = internetSettingsRegKey.GetValue("ProxyServer").ToStringSafe();
+                return;
+            }
+
+            var proxySetting = internetSettingsRegKey.GetValue("ProxyServer").ToStringSafe();
+
+            if (string.IsNullOrWhiteSpace(proxySetting))
+            {
+                return;
+            }
+
+            if (proxySetting.IndexOf(';') != -1)
+            {
+                var allProxies = proxySetting.Split(';');
+                proxySetting = allProxies.FirstOrDefault(s => s.TrimSafe().StartsWith("https="));
 
                 if (string.IsNullOrWhiteSpace(proxySetting))
                 {
-                    return;
+                    proxySetting = allProxies.FirstOrDefault(s => s.TrimSafe().StartsWith("http="));
                 }
-
-                if (proxySetting.IndexOf(';') != -1)
-                {
-                    var allProxies = proxySetting.Split(';');
-                    proxySetting = allProxies.FirstOrDefault(s => s.TrimSafe().StartsWith("https="))?.Replace("https=", string.Empty);
-
-                    if (string.IsNullOrWhiteSpace(proxySetting))
-                    {
-                        proxySetting = allProxies.FirstOrDefault(s => s.TrimSafe().StartsWith("http="))?.Replace("http=", string.Empty);
-                    }
-                }
-
-                config.Proxy.Location = proxySetting;
             }
+
+            if (proxySetting?.IndexOf('=') != -1 && !proxySetting.StartsWith("http"))
+            {
+                return;
+            }
+
+            config.Proxy.Location = proxySetting.Split('=').LastOrDefault();
         }
 
         private static void SetEnvironmentOptions(ChocolateyConfiguration config)
--- a/src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
+++ b/src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
@@ -463,15 +463,40 @@ namespace chocolatey.infrastructure.app.builders
         private static void SetProxyOptions(ChocolateyConfiguration config, Container container)
         {
             // Evaluation order of Proxy settings: System Set -> Environment Variable Set -> Chocolatey Configuration File Set -> CLI Passed in Argument
-            // If we don't yet have a Proxy Location, check if the system has one configured in the registry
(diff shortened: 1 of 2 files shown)


src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
```diff
@@ -463,15 +463,40 @@ private static void SetGlobalOptions(IList<string> args, ChocolateyConfiguration
...
+                    var allProxies = proxySetting.Split(';');
+                    proxySetting = allProxies.FirstOrDefault(s => s.TrimSafe().StartsWith("https="))?.Replace("https=", string.Empty);
+
+                    if (string.IsNullOrWhiteSpace(proxySetting))
+                    {
+                        proxySetting = allProxies.FirstOrDefault(s => s.TrimSafe().StartsWith("http="))?.Replace("http=", string.Empty);
+                    }
                 }
+
+                config.Proxy.Location = proxySetting;
```
vexx32, line 499: is it possible they might only have a single value in this registry key that still carries a `http=` or other prefix?

src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
```diff
@@ -463,15 +463,40 @@ private static void SetGlobalOptions(IList<string> args, ChocolateyConfiguration
...
+                if (string.IsNullOrWhiteSpace(proxySetting))
+                {
+                    return;
+                }
+
+                if (proxySetting.IndexOf(';') != -1)
                 {
-                    config.Proxy.Location = internetSettingsRegKey.GetValue("ProxyServer").ToStringSafe();
+                    var allProxies = proxySetting.Split(';');
+                    proxySetting = allProxies.FirstOrDefault(s => s.TrimSafe().StartsWith("https="))?.Replace("https=", string.Empty);
```
vexx32, line 491: Using .Replace here feels _weird_ to me just knowing it _could_ do a substitution somewhere other than the start of the string... I guess it should be fine, I'm curious if it would be a ton more complicated to do like a weird `IndexOf` and then `Substring` to get the remainder instead though... not sure it's worth doing that though.

vexx32: Ooh, wait. Maybe you could do `IndexOf("https=") == 0` instead of the StartsWith, and then do a `.Split('=')[1]`

I think that'd work, and _might_(?) be cheaper than a .Replace()?

tests/chocolatey-tests/features/Proxy.Tests.ps1
```diff
@@ -170,7 +170,7 @@ Describe "Proxy configuration (<Name>)" -Tag Proxy, ProxySkip -ForEach $TestCase
         $CliArgumentSet = "CliArgumentSetProxy"
 
         if ($ConfigurationsToTest.System) {
-            Set-ItemProperty -Path 'HKCU:/Software/Microsoft/Windows/CurrentVersion/Internet Settings' -name ProxyServer -Value $SystemSet
+            Set-ItemProperty -Path 'HKCU:/Software/Microsoft/Windows/CurrentVersion/Internet Settings' -name ProxyServer -Value "http=none;https=$SystemSet;"
```
vexx32, line 173: What if they have `http=$SystemSet;https=none`? We'd end up taking the `none` value if I read the code above right. Is this a valid registry setting to even do, having these `...=none` values in the string? 🤔

vexx32 (commented):
Some questions I had looking through it but on the whole it looks pretty solid! Great work :3

src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
corbob: Oh, that is definitely possible... Thanks for giving me more work to do 🤪

tests/chocolatey-tests/features/Proxy.Tests.ps1
corbob: In the case of someone setting https to an invalid proxy, I would expect their connections to fail. For the intents and purposes of this test, I think I might actually change it to just `https=$SystemSet` so that it tests the scenario you pointed out above where they only have https proxy configured...

PR #3140, opened 2023-04-28
(#508) Add warning about filtering of packages

gep13:
## Description Of Changes

This commit adds a warning, to inform the user that this may be happening.

## Motivation and Context

When running choco search, there are several options that would have previously caused packages to be filtered on the server side, before being returned to the client. The options were --approved-only, --download-cache, and --not-broken.  The decision was taken to no longer have these options passed to the server, since they can't (currently) be respected on a V3 feed.  Instead, the filtering is applied on the client side, within Chocolatey CLI.  The end result is that it is possible to get incomplete pages of data returned, as some packages may be filtered out on the client, and no additional packages will be requested.

## Testing

1. Run `choco search randompackage` and see that no warning is output
2. Run `choco search randompackage --approved-only` and see that the warning is output
3. Run `choco search randompackage --download-cache` and see that the warning is output
4. Run `choco search randompackage --not-broken` and see that the warning is output
5. Run `choco search randompackage --approved-only --limit-output` and see that no warning is output

### Operating Systems Testing

- Windows 10

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

- https://app.clickup.com/t/20540031/PROJ-599

### DIFF
--- a/src/chocolatey/infrastructure.app/services/NugetService.cs
+++ b/src/chocolatey/infrastructure.app/services/NugetService.cs
@@ -161,6 +161,22 @@ install them.
                 }
             }
 
+            if ((config.ListCommand.ApprovedOnly || config.ListCommand.DownloadCacheAvailable || config.ListCommand.NotBroken) && config.RegularOutput)
+            {
+                // This warning has been added here, to provide context to the user, and a follow up issue
+
+                // has been added here: https://github.com/chocolatey/choco/issues/3139 to address the actual
+                // issue that causes this warning to be required.
+                this.Log().Warn(@"
+Starting vith Chocolatey CLI v2.0.0, changes have been made to the
+`choco search` command which means that filtering of packages using the
+`--approved-only`, `--download-cache`, and `--not-broken` options are
+now performed within Chocolatey CLI. Previously, this filtering would
+have been performed on the Chocolatey Community Repository. As a result,
+it is possible that incomplete package lists are returned from a command
+that uses these options.");
+            }
+
             if (config.RegularOutput) this.Log().Debug(() => "Running list with the following filter = '{0}'".FormatWith(config.Input));
             if (config.RegularOutput) this.Log().Debug(ChocolateyLoggers.Verbose, () => "--- Start of List ---");
             foreach (var pkg in NugetList.GetPackages(config, _nugetLogger, _fileSystem))


src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -161,6 +161,20 @@ public virtual IEnumerable<PackageResult> List(ChocolateyConfiguration config)
                 }
             }
 
+            if ((config.ListCommand.ApprovedOnly || config.ListCommand.DownloadCacheAvailable || config.ListCommand.NotBroken) && config.RegularOutput)
+            {
+                // This warning has been added here, to provide context to the user, and a follow up issues
```
pauby, line 166:
```suggestion
                // This warning has been added here, to provide context to the user, and a follow up issue

```

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -161,6 +161,20 @@ public virtual IEnumerable<PackageResult> List(ChocolateyConfiguration config)
...
+                // has been added here: https://github.com/chocolatey/choco/issues/3139 to address the actual
+                // issue that causes this warning to be required.
+                this.Log().Warn(@"
+Starting vith Chocolatey CLI v2.0.0, changes have been made to the
+`choco search` command which means that filtering of packages using the
+--approved-only, --download-cache, and --not-broken options are not
+performed within Chocolatey CLI, rather than on the repository that is
+being queried. As a result, it is possible that incomplete pages of
+packages are returned from a command that uses these options.");
```
pauby, line 175:
```suggestion
Starting vith Chocolatey CLI v2.0.0, changes have been made to the
`choco search` command which means that filtering of packages using the
--approved-only, --download-cache, and --not-broken options are not
performed within Chocolatey CLI. These are instead performed on the
repository that is being queried. As a result, it is possible that incomplete
package lists are returned from a command that uses these options.");

```

pauby: Does this not just affect the Chocolatey Community Repository? If so, we should say 'performed by the Chocolatey Community Repository' instead of 'performed on the repository that is being queried'.

pauby (changes requested):

src/chocolatey/infrastructure.app/services/NugetService.cs
gep13: This is a great question.  Chocolatey Community Repository is the only repository server that I know of that returns these properties for a package, and there is no real reason for another repository to return them, unless they have implemented there own package moderation system, and caching mechanism.

I think it is a fairly safe bet that they only impact the Chocolatey Community Repository, so I will get this section updated.

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -161,6 +161,22 @@ public virtual IEnumerable<PackageResult> List(ChocolateyConfiguration config)
...
+                // has been added here: https://github.com/chocolatey/choco/issues/3139 to address the actual
+                // issue that causes this warning to be required.
+                this.Log().Warn(@"
+Starting vith Chocolatey CLI v2.0.0, changes have been made to the
+`choco search` command which means that filtering of packages using the
+--approved-only, --download-cache, and --not-broken options are not
+performed within Chocolatey CLI. These are instead performed by the
+Chocolatey Community Repository. As a result, it is possible that
+incomplete package lists are returned from a command that uses these
+options.");
```
AdmiringWorm, line 177: Isn't this the opposite?

It is now performed by the Chocolatey CLI, not by Chocolatey Community Repository?

gep13: Yes, I believe you are right.  I have fixed this up, can you take another look?

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -161,6 +161,22 @@ public virtual IEnumerable<PackageResult> List(ChocolateyConfiguration config)
...
 
+            if ((config.ListCommand.ApprovedOnly || config.ListCommand.DownloadCacheAvailable || config.ListCommand.NotBroken) && config.RegularOutput)
+            {
+                // This warning has been added here, to provide context to the user, and a follow up issue
+
+                // has been added here: https://github.com/chocolatey/choco/issues/3139 to address the actual
+                // issue that causes this warning to be required.
+                this.Log().Warn(@"
+Starting vith Chocolatey CLI v2.0.0, changes have been made to the
+`choco search` command which means that filtering of packages using the
```
pauby, line 172:
```suggestion
choco search command which means that filtering of packages using the

```

pauby: I would agree that backticks are used for this, but as it's going out in a warning message I'm not sure that makes sense? If it's consistent that we do this, then we should also add it to the options --approved-only, --download-cache, and --not-broken, mentioned below.

pauby (changes requested):

AdmiringWorm (approved):
LGTM, as there is currently a requested change comment I'll refrain from merging the changes until that is resolved.

src/chocolatey/infrastructure.app/services/NugetService.cs
gep13: I have had a look, and there is precedence for using backticks everywhere, so I have added them to the options.

pauby (approved):

gep13:
@AdmiringWorm and @pauby thank you for the reviews!

Issue #158, opened 2015-03-09
Change default behavior of `choco list` to be local only


findel:
The current default behaviour of `choco list` (or `clist`) is to list ALL of the packages available on chocolatey (currently 2596 in total). This is a little slow, and not really helpful as a default. 

Couldn't the default behavior be to display "local only" packages, displaying only packages that are installed locally, as when you do `choco list --localonly`?

Then require switches to get data from the repositories, rather than the other way around?

# Implementation Notes

When this work is undertaken, the deprecation notices that were add in this PR will need to be reviewed/removed:

https://github.com/chocolatey/choco/pull/2626

ferventcoder:
I love the idea of this - I want to get an idea of what it may require from other tools as this would be a breaking change.

alexchandel:
`choco list` should absolutely list only locally installed packages by default. We already have `choco search` with no arguments to list ALL packages.

ferventcoder:
Search and list are aliases

darrelmiller:
:+1:

christianrondeau:
Agreed on `list` being `--localonly` by default, and maybe `search` could not be an alias anymore? I also thought that `search` was server-based, and `list` was not. It may be interesting to run a quick survey if you're not sure.

alexchandel:
@christianrondeau I would instinctively assume `list` was local only, and only listed packages installed by default.

@ferventcoder Can this please happen in 0.9.9.7? It's intuitive (`list` gives your packages, `search` without any filter gives all packages), it parallels the behavior of other package managers, and it has broad backing. It's a breaking change, but it affects users who are exploring chocolatey packages, not scripts (compare with `choco install`), so its impact is limited.

ferventcoder:
@alexchandel it's intuitive for you but it is a breaking change, so that's why the milestone is 1.0.0. Folks need time to prepare for the break.

ferventcoder:
@alexchandel and it is used by applications that are building on top of chocolatey.

alexchandel:
@ferventcoder Is there any application known to use `choco list` instead of `choco search`, besides possibly ChocolateyGUI?

ferventcoder:
puppet choco provider / chef cookbook / boxstarter

alexchandel:
@ferventcoder So from what I saw, the three packages you mentioned all use `choco list` for local searches already (I should've specified that!) and won't be affected by this change. So it should be reasonably safe to implement.

RichiCoder1:
[...]

PR #3160, opened 2023-05-15
(#158) Restore --source for choco list

vexx32:
## Description Of Changes

- Restore `--source` param for `choco list`
- Add a handful of interfaces for the functionally different pieces of `ISourceRunner`
- Add appropriate interfaces for each currently available ISourceRunner
- Add pre-checks into ChocolateyPackageService which verifies which methods that the source runner it's trying to invoke supports and throws an error (or warning, in the case of `DryRun` / `--noop` operations) before ever trying to run it if the source runner doesn't support that operation.
- Update a few affected mocks in tests to ensure they still work correctly.

## Motivation and Context
Since alternate sources can list installed packages separately to our normal package list, we need to allow list to specify --source. This will have no effect for normal sources, but will select alternate source runners if specified.

Additionally, the early checks for each source runner as to which operations they support was implemented in a way that allows us to gut ISourceRunner in a later version and fall back to having the supported methods for each source runner enforced by the type system, so we have a clear indication as to which source runners support each kind of operation via an appropriate interface cast. I haven't added obsoletions to ISourceRunner's main methods just yet, but that might be something we want to do in future.

## Testing

- Check that `choco list --source windowsfeatures` works
- Check that `choco search --source windowsfeatures` works
- Check that `choco list --source cygwin` reports as not supported _without_ first installing cygwin if it's not yet installed
- Ensure that supported operations for list and for search for each source work as expected (currently afaik only the windowsfeatures source does any actual searching, the rest only do listing for local or remote packages).

### Operating Systems Testing

Win11

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

Restoring some functionality that was impacted by our changes when resolving #158

### DIFF
--- a/src/chocolatey/infrastructure.app/services/ISourceRunner.cs
+++ b/src/chocolatey/infrastructure.app/services/ISourceRunner.cs
@@ -25,99 +25,9 @@ namespace chocolatey.infrastructure.app.services
     using results;
 
     [MultiService]
-    public interface ISourceRunner
+    [Obsolete("This interface is deprecated and will be removed in v3.")]
+    public interface ISourceRunner : IBootstrappableSourceRunner, ICountSourceRunner, IListSourceRunner, IInstallSourceRunner, IUpgradeSourceRunner, IUninstallSourceRunner
     {
-        /// <summary>
-        ///   Gets the source type the source runner implements
-        /// </summary>
-        /// <value>
-        ///   The type of the source.
-        /// </value>
-        string SourceType { get; }
-
-        /// <summary>
-        ///   Ensures the application that controls a source is installed
-        /// </summary>
-        /// <param name="config">The configuration.</param>
-        /// <param name="ensureAction">The action to continue with as part of the install</param>
-        void EnsureSourceAppInstalled(ChocolateyConfiguration config, Action<PackageResult, ChocolateyConfiguration> ensureAction);
-
-        /// <summary>
-        ///     Retrieve the listed packages from the source feed cout
-        /// </summary>
-        /// <param name="config">The configuration.</param>
-        /// <returns>Packages count</returns>
-        int Count(ChocolateyConfiguration config);
-
-        /// <summary>
-        ///   Run list in noop mode
-        /// </summary>
-        /// <param name="config">The configuration.</param>
-        void ListDryRun(ChocolateyConfiguration config);
-
-        /// <summary>
-        ///   Lists/searches for packages from the source feed
-        /// </summary>
-        /// <param name="config">The configuration.</param>
-        /// <returns></returns>
-        IEnumerable<PackageResult> List(ChocolateyConfiguration config);
-
-        /// <summary>
-        ///   Run install in noop mode
-        /// </summary>
-        /// <param name="config">The configuration.</param>
-        /// <param name="continueAction">The action to continue with for each noop test install.</param>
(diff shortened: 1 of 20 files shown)


src/chocolatey/infrastructure.app/services/ISourceRunner.cs
```diff
@@ -25,122 +25,164 @@ namespace chocolatey.infrastructure.app.services
     using results;
 
     [MultiService]
-    public interface ISourceRunner
+    [Obsolete("This interface is deprecated and will be removed in v3.")]
+    public interface ISourceRunner : IBootstrappableSourceRunner, ICountSourceRunner, IListSourceRunner, IInstallSourceRunner, IUpgradeSourceRunner, IUninstallSourceRunner
```
vexx32, line 29: If this is being deprecated here, we should probably(?) make the NugetService / INugetService _also_ no longer use it? Currently I think it inherits off ISourceRunner directly; we should probably use the piece-wise interface for that as well.

src/chocolatey/infrastructure.app/services/ISourceRunner.cs
```diff
@@ -25,122 +25,164 @@ namespace chocolatey.infrastructure.app.services
...
+        [Obsolete("This overload is deprecated and will be removed in v3.")]
+        ConcurrentDictionary<string, PackageResult> upgrade_run(ChocolateyConfiguration config, Action<PackageResult, ChocolateyConfiguration> continueAction, Action<PackageResult, ChocolateyConfiguration> beforeUpgradeAction = null);
+        [Obsolete("This overload is deprecated and will be removed in v3.")]
+        void uninstall_noop(ChocolateyConfiguration config, Action<PackageResult, ChocolateyConfiguration> continueAction);
+        [Obsolete("This overload is deprecated and will be removed in v3.")]
+        ConcurrentDictionary<string, PackageResult> uninstall_run(ChocolateyConfiguration config, Action<PackageResult, ChocolateyConfiguration> continueAction, Action<PackageResult, ChocolateyConfiguration> beforeUninstallAction = null);
+#pragma warning restore IDE1006
+    }
+
+    public interface IAlternativeSourceRunner
```
vexx32, line 57: Also, since we're using this for all our sourcerunners, NugetService included (even if it's still using the main ISourceRunner, it gets this by virtue of the inherited interfaces) I feel like it's a bit weird to call this "Alternative".

If we specifically want a new name for the interface maybe we could go with something like `IPackageSource` or something along those lines?

Though, given the rest of the now-very-breaking changes added here, I think we could just skip the deprecation here anyway and replace ISourceRunner completely with this interface...

gep13: The`Alternative` here refers to Alternative Sources, which is the terminology used for things like Cygwin, Python, etc.  You can see that in our docs here: https://docs.chocolatey.org/en-us/choco/commands/search#alternative-sources

src/chocolatey/infrastructure.app/services/ISourceRunner.cs
gep13: I seem to remember this being discussed, and the long and short of it was we decided to leave this in place for now.  There was quite a bit of discussion on this at the time, but it was also in between lots of other discussions, so I don't recall the exact details. We can discuss further once everyone is back from vacation, etc.

For now, watching the automated tests, to see about moving forward with taking some release candidates for a spin...

PR #3175, opened 2023-05-31
(#3174) Apply version number normalizing consistently

### LINKED ISSUE #3174: Version normalization is not being applied everywhere
### Checklist

### What You Are Seeing?

When using `choco upgrade --noop 7zip` with 7zip installed, in some cases it seems to not be applying version normalization and shows two slightly different version numbers:

```ps
PS> choco upgrade --noop 7zip -r
7zip|22.1.0|22.1|false
```

### What is Expected?

The upgrade command should be normalizing both version numbers it shows so as to avoid confusion.

1. `choco install 7zip`
2. `choco upgrade --noop 7zip`

### System Details

* Operating System: Win11
* Chocolatey CLI Version: 2.0.0
* Chocolatey Licensed Extension version: 6.0.0
* Chocolatey License type: Business
* Terminal/Emulator: Microsoft Terminal

### Installed Packages

```bash
7zip 22.1.0
```

### Output Log

```bash
https://gist.github.com/vexx32/568fb18678c8accebd43f7f2fd04f238
```

### Additional Context

_No response_

vexx32:
## Description Of Changes

Went through all references to `ToStringSafe` which were being called on a NuGetVersion instance and changed them all to `ToNormalizedStringChecked`.

Additionally, found a place in the upgrade dry-run code that was just missing a call to this method and outputting a non-normalized version string and added it.

## Motivation and Context

We should be consistent in how we internally and externally represent versions as strings, so that folks relying on our output for consistent comparisons don't need to think too hard about it.

## Testing

1. `choco install 7zip --ignore-dependencies`
2. `choco upgrade --noop 7zip -r`
3. The output result should display `7zip|22.1.0|22.1.0|false` exactly (unless a new version of 7zip has been installed, so just verify the version numbers match exactly)

We _might_ be able to add a smoke test package for this case, but I've not had the time to look into it as yet. I will update here if I manage to get that done as well.

### Operating Systems Testing

WIn11

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

Fixes #3174

### DIFF
--- a/src/chocolatey/infrastructure.app/services/ChocolateyPackageInformationService.cs
+++ b/src/chocolatey/infrastructure.app/services/ChocolateyPackageInformationService.cs
@@ -225,7 +225,7 @@ A corrupt .registry file exists at {0}.
             {
                 var versionOverrideFile = _fileSystem.CombinePaths(pkgStorePath, VersionOverrideFile);
                 if (_fileSystem.FileExists(versionOverrideFile)) _fileSystem.DeleteFile(versionOverrideFile);
-                _fileSystem.WriteFile(versionOverrideFile, packageInformation.VersionOverride.ToStringSafe());
+                _fileSystem.WriteFile(versionOverrideFile, packageInformation.VersionOverride.ToNormalizedStringChecked());
             }
             else
             {
@@ -265,6 +265,9 @@ A corrupt .registry file exists at {0}.
             {
                 return preferredStorePath;
             }
+
+            // Legacy handling for old package versions that was installed prior to v2.0.0.
+            // Do not remove the call to `ToStringSafe`.
 
             var pkgStorePath = fileSystem.CombinePaths(ApplicationParameters.ChocolateyPackageInfoStoreLocation, "{0}.{1}".FormatWith(id, version.ToStringSafe()));
 
--- a/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
+++ b/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
@@ -1,4 +1,4 @@
-﻿// Copyright © 2017 - 2023 Chocolatey Software, Inc
+// Copyright © 2017 - 2023 Chocolatey Software, Inc
 // Copyright © 2011 - 2017 RealDimensions Software, LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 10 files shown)


src/chocolatey/infrastructure.app/services/ChocolateyPackageInformationService.cs
```diff
@@ -266,7 +266,7 @@ private static string GetStorePath(IFileSystem fileSystem, string id, NuGetVersi
                 return preferredStorePath;
             }
 
-            var pkgStorePath = fileSystem.CombinePaths(ApplicationParameters.ChocolateyPackageInfoStoreLocation, "{0}.{1}".FormatWith(id, version.ToStringSafe()));
+            var pkgStorePath = fileSystem.CombinePaths(ApplicationParameters.ChocolateyPackageInfoStoreLocation, "{0}.{1}".FormatWith(id, version.ToNormalizedStringChecked()));
```
AdmiringWorm, line 269: This one should not be `ToNormalizedStringChecked`. it uses `ToStringSafe` on purpose to catch legacy installations of a package.

src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
```diff
@@ -1314,7 +1314,7 @@ public virtual void HandlePackageUninstall(PackageResult packageResult, Chocolat
         {
             if (!_fileSystem.DirectoryExists(packageResult.InstallLocation))
             {
-                packageResult.InstallLocation += ".{0}".FormatWith(packageResult.PackageMetadata.Version.ToStringSafe());
+                packageResult.InstallLocation += ".{0}".FormatWith(packageResult.PackageMetadata.Version.ToNormalizedStringChecked());
             }
```
AdmiringWorm, line 1319: I think we can remove this entire if statement.

This looks to be related to side by side installations, which we no longer support.

/cc @gep13

AdmiringWorm (dismissed):
In addition to the requested changes, the failing tests need to be investigated as well.

src/chocolatey/infrastructure.app/services/ChocolateyPackageInformationService.cs
gep13: Can we add a comment in this area, so that this information isn't lost?

src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs
gep13: Yes, I would tend to agree.  There should be no need to have a versioned folder name going forwad.

AdmiringWorm: The question would be if we should remove it right now or not.

In theory (need to verify the behavior first), leaving it in could allow users to uninstall a package that was previously in v1 installed in side by side.

Removing it, will disable that ability immediately.

If we keep it, we should not call `ToNormalizedStringChecked` though.

AdmiringWorm: I have removed this statement for now, as it doesn't have any actual effect.

gep13 (approved):
LGTM!

gep13:
@AdmiringWorm @vexx32 thanks for getting this updated!

PR #3187, opened 2023-06-01
(maint) Fix issue and PR labels for GH Action

pauby:
## Description Of Changes
Fixed GH action so it only runs against issues that have the `0 - Waiting on User` label. And the PR label for Pending Closure was incorrect.

## Motivation and Context
Workflow is not running as it should.

## Testing

N/A

### Operating Systems Testing

N/A

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

This is a repository fix.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -20,11 +20,11 @@ jobs:
           exempt-all-assignees: false
           exempt-draft-pr: true
           stale-issue-label: "Pending Closure"
-          stale-pr-label: '0 - Waiting on User'
-          only-pr-labels: '0 - Waiting on User'
+          stale-pr-label: 'Pending Closure'
+          only-labels: '0 - Waiting on User'
           close-issue-label: "No Response / Stale"
           close-pr-label: "No Response / Stale"
-          exempt-issue-labels: 'Security / CVE,0 - Backlog,1 - Ready for work,2 - Working,3 - Review,4 - Done,5 - Push required'
+          exempt-issue-labels: 'Security / CVE'
           exempt-pr-labels: 'Security / CVE'
           labels-to-remove-when-unstale: '0 - Wating on User,Pending closure'
           stale-issue-message: |


gep13:
@AdmiringWorm I am not too familiar with how the stale workflow works, and I think you have used it in the past.  If so, can you have a look at this, and let me know?  The workflow has been running and marking issues as pending closure, when it shouldn't be.

pauby:
@AdmiringWorm @gep13 I don't see this being synced from the .github repository. It might be better to set that up instead so that if we make a change, we don't have to wait until a choco release for it to apply.

@AdmiringWorm thoughts on that?

gep13:
We wouldn't need to wait for a release to make this sort of change, we can merge it into develop as soon as we need to, and this will then be picked up and start working as.

In general though, having it in one place, and re-used from these, does make sense though.

.github/workflows/stale.yml
```diff
@@ -20,11 +20,11 @@ jobs:
           exempt-all-assignees: false
           exempt-draft-pr: true
           stale-issue-label: "Pending Closure"
-          stale-pr-label: '0 - Waiting on User'
-          only-pr-labels: '0 - Waiting on User'
+          stale-pr-label: 'Pending Closure'
+          only-labels: '0 - Waiting on User'
           close-issue-label: "No Response / Stale"
           close-pr-label: "No Response / Stale"
-          exempt-issue-labels: 'Security / CVE,0 - Backlog,1 - Ready for work,2 - Working,3 - Review,4 - Done,5 - Push required'
```
AdmiringWorm, line 27: Do we want to remove these labels as exempting labels?

There could be cases where we need more information from the user, but we know it will need to be looked at. If we have such a case, we can use these labels to prevent the issue from being closed even though it is waiting for the user.

AdmiringWorm:
> I don't see this being synced from the .github repository. It might be better to set that up instead so that if we make a change, we don't have to wait until a choco release for it to apply.

Even if we add it to the .github repository, it will still need to leave here in the repository as well. So there is no immediate need to add it to the .github repository yet until we want it synchronized with other repositories (once we get that properly set up).

.github/workflows/stale.yml
pauby: I feel that if we have added the 'waiting for user' label, we cannot proceed with the issue until they come back. If we want to keep the issue open regardless of whether the user comes back or not, we shouldn't add that label at all.

That is something we could add as a process.

AdmiringWorm: Fair enough.

PR #3229, opened 2023-06-25
(#3224) Do not try to use the Licensed Assembly when it is not loaded for compatibility reasons.

### LINKED ISSUE #3224: Chocolatey CLI tries to load the licensed environment settings when the licensed extension assembly is not supposed to be loaded
### Checklist

### What You Are Seeing?

When you have an incompatible version of Chocolatey Licensed Extension, the Environment is still attempted to be set by the licensed extension.

### What is Expected?

The Licensed Extension should not be called unless the user explicitly asks for it.

1. Build a custom Chocolatey Licensed Extension with a MinimumChocolateyVersion set higher than the CLI version.
2. Install this custom Licensed Extension
3. Run a Chocolatey command
4. See exceptions due to mismatch in strong name signing keys.

### System Details

* Operating System: Windows 10
* Chocolatey CLI Version: 2.0.0
* Chocolatey Licensed Extension version: Custom
* Chocolatey License type: Business
* Terminal/Emulator: conhost

### Installed Packages

```bash
chocolatey 2.0.0
chocolatey.extension 6.1.0-corys-20230622
```

### Output Log

```bash
N/A
```

### Additional Context

Here is a screenshot with a custom CLI as well:
image

corbob:
## Description Of Changes

Update the logic used to determine if Chocolatey should set the Licensed Environment to include if the Assembly is loaded.

## Motivation and Context

If the assembly has not been loaded, we should consider Chocolatey Licensed Extension unavailable.

## Testing

1. Follow the Manual Testing steps for an [Incompatible Chocolatey Licensed Extension](https://chocolatey.atlassian.net/wiki/spaces/DEV/pages/1562771487/Manual+Testing+Process+-+Chocolatey#Testing-Incompatible-Chocolatey-Licensed-Extension-with-Chocolatey).
2. Run some manual commands with `--skip-compatibility-check`.
3. Ensure that you get an exception about unable to load chocolatey.lib with the official strongname public key.
4. Enable the feature (`choco feature enable -n disableCompatibilityChecks`) to disable the check and run some manual commands.
5. Ensure that you get an exception about unable to load chocolatey.lib with the official strongname public key.

Examples of some commands to run: `choco list`, `choco search chocolatey`, `choco feature disable -n disableCompatibilityChecks`.

### Operating Systems Testing

- Windows Server 2019
- Windows Server 2016

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

Fixes #3224

### DIFF
--- a/src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
+++ b/src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
@@ -522,6 +522,7 @@ namespace chocolatey.infrastructure.app.builders
         private static void SetLicensedOptions(ChocolateyConfiguration config, ChocolateyLicense license, ConfigFileSettings configFileSettings)
         {
             config.Information.IsLicensedVersion = license.IsLicensedVersion();
+            config.Information.IsLicensedAssemblyLoaded = license.AssemblyLoaded;
             config.Information.LicenseType = license.LicenseType.DescriptionOrValue();
 
             if (license.AssemblyLoaded)
--- a/src/chocolatey/infrastructure.app/configuration/EnvironmentSettings.cs
+++ b/src/chocolatey/infrastructure.app/configuration/EnvironmentSettings.cs
@@ -136,7 +136,10 @@ namespace chocolatey.infrastructure.app.configuration
         {
             Environment.SetEnvironmentVariable("ChocolateyLicenseType", config.Information.LicenseType);
 
-            if (!(config.Information.IsLicensedVersion && config.Information.IsLicensedAssemblyLoaded)) return;
+            if (!(config.Information.IsLicensedVersion && config.Information.IsLicensedAssemblyLoaded))
+            {
+                return;
+            }
 
             var licenseAssembly = AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(a => a.GetName().Name.IsEqualTo("chocolatey.licensed"));
 
--- a/src/chocolatey/infrastructure.app/configuration/EnvironmentSettings.cs
+++ b/src/chocolatey/infrastructure.app/configuration/EnvironmentSettings.cs
@@ -1,4 +1,4 @@
-﻿// Copyright © 2017 - 2021 Chocolatey Software, Inc
+// Copyright © 2017 - 2021 Chocolatey Software, Inc
 // Copyright © 2011 - 2017 RealDimensions Software, LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/chocolatey/infrastructure.app/configuration/EnvironmentSettings.cs
+++ b/src/chocolatey/infrastructure.app/configuration/EnvironmentSettings.cs
@@ -134,9 +134,9 @@ namespace chocolatey.infrastructure.app.configuration
 
         private static void SetLicensedEnvironment(ChocolateyConfiguration config)
         {
-            if (!config.Information.IsLicensedVersion) return;
-
             Environment.SetEnvironmentVariable("ChocolateyLicenseType", config.Information.LicenseType);
+
+            if (!(config.Information.IsLicensedVersion && config.Information.IsLicensedAssemblyLoaded)) return;
 
             var licenseAssembly = AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(a => a.GetName().Name.IsEqualTo("chocolatey.licensed"));
 
--- a/tests/helpers/common/Chocolatey/Get-ChocolateyVersion.ps1
+++ b/tests/helpers/common/Chocolatey/Get-ChocolateyVersion.ps1
@@ -8,7 +8,7 @@
     [CmdletBinding()]
     param()
     if (-not $script:runningVersion) {
-        [NuGet.Versioning.NuGetVersion]$script:runningVersion = ((Invoke-Choco --version).Lines | Where-Object { $_ -NotMatch "please upgrade" }) -join '`r`n'
+        [NuGet.Versioning.NuGetVersion]$script:runningVersion = (Invoke-Choco --version).Lines | Where-Object { $_ -Match "^\d+\.[\d\.]+" } | Select-Object -First 1
     }
 
     $script:runningVersion
--- a/src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
+++ b/src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
@@ -527,6 +527,7 @@ NOTE: Hiding sensitive configuration data! Please double and triple
         public bool IsUserRemote { get; set; }
         public bool IsProcessElevated { get; set; }
         public bool IsLicensedVersion { get; set; }
+        public bool IsLicensedAssemblyLoaded { get; set; }
         public string LicenseType { get; set; }
         public string CurrentDirectory { get; set; }
     }


src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
```diff
@@ -510,7 +510,7 @@ private static void SetEnvironmentOptions(ChocolateyConfiguration config)
 
         private static void SetLicensedOptions(ChocolateyConfiguration config, ChocolateyLicense license, ConfigFileSettings configFileSettings)
         {
-            config.Information.IsLicensedVersion = license.IsLicensedVersion();
+            config.Information.IsLicensedVersion = license.IsLicensedVersion() && license.AssemblyLoaded;
```
AdmiringWorm, line 513: This is wrong.

Even if you are unable to load the extension, or if it isn't present, as long as you have a licensed file it is considered to be a licensed version of Chocolatey CLI.

Additionally, by setting `IsLicensedVersion` to false will prevent the installation of `chocolatey.extension` to happen.

For what you brought up internally, you need to update the `EnvironmentSettings` class that I linked to you to.

image
image

AdmiringWorm (changes requested):

src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
corbob: Ah, that would explain the failure installing the licensed extension 😅 . I erroneously thought the error was related to the strongname mismatch that I had forced.

My reasoning for trying to do it here was that we don't have a License object at the EnvironmentSettings, then thinking of it, it seemed like we might have similar checks that might fail elsewhere. So this seemed like it would solve the issue, as well as stave off other potential issues. Clearly that's not the case. I'll get this part backed out and add the check where we initially discussed (and do a quick check of other places we have similar checks).

tests/helpers/common/Chocolatey/Get-ChocolateyVersion.ps1
```diff
@@ -8,7 +8,7 @@
     [CmdletBinding()]
     param()
     if (-not $script:runningVersion) {
-        [NuGet.Versioning.NuGetVersion]$script:runningVersion = ((Invoke-Choco --version).Lines | Where-Object { $_ -NotMatch "please upgrade" }) -join '`r`n'
+        [NuGet.Versioning.NuGetVersion]$script:runningVersion = ((Invoke-Choco --version).Lines | Where-Object { $_ -NotMatch "please upgrade" } | Select-Object -Last 1) -join '`r`n'
```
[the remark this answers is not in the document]
corbob: It's needed because without it, we think the version contains the compatibility warning, and so our version checks don't actually work. See the below screenshot for what is output with the different scenarios.

corbob:
I believe I got all the instances where this check might need to update. The ones I saw and didn't change (with reason):

- https://github.com/chocolatey/choco/blob/10a91ef2e13249ec0f3ca65884ca46858e848df2/src/chocolatey/infrastructure.app/nuget/NugetList.cs#L289
  - I wasn't sure if this one is needed. The more I look at it, the more I think this should be included. If I'm reading the logic correctly this is a scenario where the user has asked for not broken, and if the package is in the CDN, and the user is licensed we assume they're not broken. But this would only be the case if the assembly is also loaded.
- https://github.com/chocolatey/choco/blob/10a91ef2e13249ec0f3ca65884ca46858e848df2/src/chocolatey/infrastructure.app/runners/GenericRunner.cs#L137
  - This is in the `FailOnMissingOrInvalidLicenseIfFeatureSet` method. It seems reasonable that this should not throw if the assembly isn't loaded so that they can set preferences as needed before actually installing the extension.
- https://github.com/chocolatey/choco/blob/10a91ef2e13249ec0f3ca65884ca46858e848df2/src/chocolatey/infrastructure.app/services/AutomaticUninstallerService.cs#L218
  - This just controls outputting a warning that the licensed versions are more effective at automatic uninstallation.
- https://github.com/chocolatey/choco/blob/10a91ef2e13249ec0f3ca65884ca46858e848df2/src/chocolatey/infrastructure.app/services/ChocolateyPackageService.cs#L423
  - This is the method to randomly notify the user about licensed features. It seems reasonable that a licensed user doesn't need to see this.

src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
corbob: Actually, this error occurs even with the check on the line mentioned in Slack. It looks like the `ChocolateyLicenseType` environment variable is set immediately after the updated check in EnvironmentSettings.

Even funner, the change to get the license causes all sorts of more output that is probably undesirable (Each time we run `License.ValidateLicense()` we output if the license file is there but assembly is not).



I've made some more changes, just testing them out locally before pushing up to GitHub, then will kick off some Test Kitchen runs both locally and in Team City.

src/chocolatey/infrastructure.app/configuration/EnvironmentSettings.cs
```diff
@@ -28,6 +28,7 @@ namespace chocolatey.infrastructure.app.configuration
     using logging;
     using nuget;
     using Environment = adapters.Environment;
+    using License = licensing.License;
```
gep13, line 31: I don't think that this is needed, and can be removed.

src/chocolatey/infrastructure.app/commands/ChocolateyUpgradeCommand.cs
```diff
@@ -23,6 +23,7 @@ namespace chocolatey.infrastructure.app.commands
     using commandline;
     using configuration;
     using infrastructure.commands;
+    using licensing;
```
gep13, line 26: I don't think that this is needed, and can be removed.

src/chocolatey/infrastructure.app/commands/ChocolateyInstallCommand.cs
```diff
@@ -23,6 +23,7 @@ namespace chocolatey.infrastructure.app.commands
     using commandline;
     using configuration;
     using infrastructure.commands;
+    using licensing;
```
gep13, line 26: I don't think this is needed, and can be removed.

gep13 (dismissed):

src/chocolatey/infrastructure.app/commands/ChocolateyInstallCommand.cs
```diff
@@ -233,7 +234,7 @@ public virtual void Validate(ChocolateyConfiguration configuration)
             if (!string.IsNullOrWhiteSpace(configuration.Input))
             {
                 var unparsedOptionsAndPackages = configuration.Input.Split(new[] { " " }, StringSplitOptions.RemoveEmptyEntries);
-                if (!configuration.Information.IsLicensedVersion)
+                if (!(configuration.Information.IsLicensedVersion && configuration.Information.IsLicensedAssemblyLoaded))
```
AdmiringWorm, line 237: I am not so sure about this change. Yes the Directory argument won't be available if they don't have the extension installed, but it is however available for a licensed customer, throwing an exception if the license installed doesn't sound appropriate.

At best I would say it should only output a warning in that case that they need to install a supported version of Chocolatey Licensed Extension.

@gep13 thoughts?

src/chocolatey/infrastructure.app/commands/ChocolateyUpgradeCommand.cs
```diff
@@ -264,7 +265,7 @@ public virtual void Validate(ChocolateyConfiguration configuration)
             if (!string.IsNullOrWhiteSpace(configuration.Input))
             {
                 var unparsedOptionsAndPackages = configuration.Input.Split(new[] { " " }, StringSplitOptions.RemoveEmptyEntries);
-                if (!configuration.Information.IsLicensedVersion)
+                if (!(configuration.Information.IsLicensedVersion && configuration.Information.IsLicensedAssemblyLoaded))
```
AdmiringWorm, line 268: Same as above for the `install` command.

src/chocolatey/infrastructure.app/configuration/EnvironmentSettings.cs
```diff
@@ -134,10 +135,10 @@ public static void SetEnvironmentVariables(ChocolateyConfiguration config)
 
         private static void SetLicensedEnvironment(ChocolateyConfiguration config)
         {
-            if (!config.Information.IsLicensedVersion) return;
-
             Environment.SetEnvironmentVariable("ChocolateyLicenseType", config.Information.LicenseType);
 
+            if (!(config.Information.IsLicensedVersion && config.Information.IsLicensedAssemblyLoaded)) return;
```
AdmiringWorm, line 140: Let us use curly braces for any new code

```suggestion
            if (!(config.Information.IsLicensedVersion && config.Information.IsLicensedAssemblyLoaded))
            {
                return;
            }

```

Also, any reason you changed this to happen after setting the environment variable?
This will most likely break some of our test kitchen tests (if memory serves me right).

tests/helpers/common/Chocolatey/Get-ChocolateyVersion.ps1
```diff
@@ -8,7 +8,7 @@
     [CmdletBinding()]
     param()
     if (-not $script:runningVersion) {
-        [NuGet.Versioning.NuGetVersion]$script:runningVersion = ((Invoke-Choco --version).Lines | Where-Object { $_ -NotMatch "please upgrade" }) -join '`r`n'
+        [NuGet.Versioning.NuGetVersion]$script:runningVersion = (Invoke-Choco --version).Lines | Where-Object { $_ -NotMatch "please upgrade" } | Select-Object -Last 1
```
AdmiringWorm, line 11: As you mentioned previously we only need the absolute last line so there is no longer a need for the Where-Object block AFAIK.
As such it can be changed to

```suggestion
        [NuGet.Versioning.NuGetVersion]$script:runningVersion = (Invoke-Choco --version).Lines | Select-Object -Last 1
```

or

```suggestion
        [NuGet.Versioning.NuGetVersion]$script:runningVersion = (Invoke-Choco --version).Lines | Where-Object { $_ -Match "^\d+\.[\d\.]+" } | Select-Object -First 1

```

AdmiringWorm (changes requested):

src/chocolatey/infrastructure.app/configuration/EnvironmentSettings.cs
corbob: I was going to tell you Visual Studio added it (which is true), but I took out the `License.ValidateLicense()` calls that needed it, so you're absolutely right this is not needed. I've rebased away the commit that added this.

src/chocolatey/infrastructure.app/commands/ChocolateyUpgradeCommand.cs
corbob: I was going to tell you Visual Studio added it (which is true), but I took out the `License.ValidateLicense()` calls that needed it, so you're absolutely right this is not needed. I've rebased away the commit that added this.

src/chocolatey/infrastructure.app/commands/ChocolateyInstallCommand.cs
corbob: I was going to tell you Visual Studio added it (which is true), but I took out the `License.ValidateLicense()` calls that needed it, so you're absolutely right this is not needed. I've rebased away the commit that added this.

src/chocolatey/infrastructure.app/configuration/EnvironmentSettings.cs
corbob: I moved it to after setting that particular environment variable because the `chocolateyInstall.ps1` script for the extension explicitly checks it and fails if it's not set.

As for reformatting the if statement, I left it without braces as that's the way I found it. But my preference is to have them as well, so I'll get that updated.

src/chocolatey/infrastructure.app/commands/ChocolateyInstallCommand.cs
corbob: As per our discussion on Zoom, I've backed out this change.

src/chocolatey/infrastructure.app/commands/ChocolateyUpgradeCommand.cs
corbob: As per our discussion on Zoom, I've backed out this change.

tests/helpers/common/Chocolatey/Get-ChocolateyVersion.ps1
corbob: I'm not entirely sure where the `NotMatch` on `please upgrade` was added/needed for, but my testing shows that the version check with the Regex works, so I went with that one.

src/chocolatey/infrastructure.app/configuration/EnvironmentSettings.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright © 2017 - 2021 Chocolatey Software, Inc
+// Copyright © 2017 - 2021 Chocolatey Software, Inc
```
AdmiringWorm, line 1: What changed here?
The removal of BOM or adding it?

If adding it, can we have that as part of a different commit?

tests/chocolatey-tests/chocolatey.Tests.ps1
```diff
@@ -257,6 +257,21 @@ exit $error.count
...
+
+        It "Should exit with Failure (1)" {
+            $Output.ExitCode | Should -Be 1 -Because $Output.String
+        }
+
+        It "Should report that is a licensed only option" {
+            $Output.Lines | Should -Contain "It appears you are attempting to use options that may be only available in licensed versions of Chocolatey ('--installdirectory='c:\temp''). There may be ways in the open source edition to achieve what you are looking to do. Please remove the argument and consult the documentation." -Because $Output.String
+        }
+    }
+
```
AdmiringWorm, line 274: Since the change about throwing the exception was backed out, this test is no longer valid.

AdmiringWorm (changes requested):
In addition to the requested changes, can you update the PR title to more accurately reflect the changes in this PR?

It is not that Chocolatey CLI will not be treated as licensed, as it is still a licensed edition. Just without the full functionality when the Licensed Extension is installed and loaded.

src/chocolatey/infrastructure.app/configuration/EnvironmentSettings.cs
corbob: Hmmm... Not sure why this changed. I'll take a look and add it as it's own commit if it should change.

corbob: Apparently Visual Studio decided this shouldn't have a BOM. Checking other cs files, it seems some have BOM, and some don't. I've added a maint commit to remove the BOM.

tests/chocolatey-tests/chocolatey.Tests.ps1
corbob: I've removed this addition.

src/chocolatey/infrastructure.app/configuration/EnvironmentSettings.cs
AdmiringWorm: We'll need to come to an agreement at some point of what files should have a BOM, and which should not.

I think my VS instance usually adds it back in (not sure), and older VS versions almost always added it in. But let us leave it at that for now.

AdmiringWorm:
@corbob the changes works as expected. Great work on getting this done 👍

PR #3262, opened 2023-07-11
(#3258) Expand logging for nuget resources errors

### LINKED ISSUE #3258: Expand logging around exception thrown when unable to communicate with source
### Checklist

There are times when Chocolatey CLI isn't able to communicate with a source.  When this happens, the warning that is shown at the command line is typically the following:

```
Unable to load the service index for source https://community.chocolatey.org/api/v2/.
```

From what we have seen, the reason that this can happen range from:

- attempting to run Chocolatey CLI on Windows 7/Windows Server 2008/R2
- issues with proxy configuration
- issues with VPN's
- etc

We need to provide more information in the Chocolatey log file when this happens, so that we can attempt to point people in the right direction.

When the exception:

```
Unable to load the service index for source https://community.chocolatey.org/api/v2/.
```

Is thrown, there is typically a number of inner exceptions, that provide additional detail on the problem.

We need to enumerate through these exceptions, and record the message from each of them in the log file (NOTE: We aren't planning on emitting these to the CLI), so that we can instruct folks on how to resolve the problem.

In addition to capturing this output in the log file, we will also include a ch0.co short link to the documentation that is linked below, so that folks can try to figure out what is going in.

### Additional Context

N/A

### Related Issues

- https://github.com/chocolatey/docs/issues/772
- https://github.com/chocolatey/choco/issues/3244
- https://github.com/chocolatey/choco/issues/3257

vexx32:
## Description Of Changes

- When resolving resources for a repository, take the initial warning we have and then inspect its inner exceptions.
- Log _all_ exception messages found this way to Debug, except for the last one which is logged as a warning again.

The reason the outermost and innermost exceptions are logged as warnings and the rest are relegated to debug is so that we have the best chance of giving users the most actionable information -- often the surface or the deepest exception contain the real problem, while the ones in the middle are often not as useful. However, should we encounter cases where the others are needed, users can inspect the log file for the additional messages, or use `--debug` to see them on a re-run.

## Motivation and Context

Previously all that was logged was the surface exception, which in quite a few cases we've seen not yielding any useful information.

Instead, log the entire exception chain, surfacing just the outermost and innermost exceptions, which we generally expect to contain the most relevant information for users.

## Testing

Reproducing the exact cases we've seen this thus far seemed non-trivial and needing quite a bit of setup to me, so I found this trivial case that illustrates the same behaviour:

1. Run `choco search chocolatey --source https://www.google.com/ --proxy www.google.com`
2. You should get two warnings; one that says `unable to load the service index`, and another that says `the remote server returned an error (405)`
3. Rerun the command with `--debug` and note there is an additional (in this case, not useful) exception logged between the two warnings that simply says `an error occurred while sending the request`

### Operating Systems Testing

Win11

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

* [x] Requires a change to the documentation. - the shortlink links to a doc page that is to-be-written

## Related Issue

Fixes #3258

### DIFF
--- a/src/chocolatey/chocolatey.csproj
+++ b/src/chocolatey/chocolatey.csproj
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="..\packages\Microsoft.CodeAnalysis.BannedApiAnalyzers.3.3.3\build\Microsoft.CodeAnalysis.BannedApiAnalyzers.props" Condition="Exists('..\packages\Microsoft.CodeAnalysis.BannedApiAnalyzers.3.3.3\build\Microsoft.CodeAnalysis.BannedApiAnalyzers.props')" />
   <PropertyGroup>
--- a/src/chocolatey/chocolatey.csproj
+++ b/src/chocolatey/chocolatey.csproj
@@ -200,6 +200,7 @@
       <Link>Properties\SolutionVersion.cs</Link>
     </Compile>
     <Compile Include="AssemblyExtensions.cs" />
+    <Compile Include="ExceptionExtensions.cs" />
     <Compile Include="infrastructure.app\attributes\MultiServiceAttribute.cs" />
     <Compile Include="infrastructure.app\commands\ChocolateyCacheCommand.cs" />
     <Compile Include="infrastructure.app\commands\ChocolateyListCommand.cs" />
--- a/src/chocolatey/infrastructure.app/nuget/NuGetEndpointResources.cs
+++ b/src/chocolatey/infrastructure.app/nuget/NuGetEndpointResources.cs
@@ -153,7 +153,32 @@ namespace chocolatey.infrastructure.app.nuget
             {
                 if (!_resolvingFailed)
                 {
-                    this.Log().Warn(ex.InnerException.Message);
+                    // Unwrap the AggregateException as its surface message is useless
+                    Exception error = ex.InnerException;
+                    this.Log().Warn(error.Message);
+
+                    // Enumerate the inner exceptions, log all but the last one in the list to debug
+                    string message = null;
+                    foreach (var err in error.InnerException.Enumerate())
+                    {
+                        if (message != null)
+                        {
+                            this.Log().Debug(message);
+                        }
+
+                        message = err.Message;
+                        error = err;
+                    }
+
+                    // If the last error in the list isn't the only one, write its message as a warning.
+                    // Typically the deepest/last error in the InnerExceptions will be a relevant and
+                    // actionable error.
+                    if (error != ex.InnerException && message != null)
+                    {
+                        this.Log().Warn(message);
+                    }
+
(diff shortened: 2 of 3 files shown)


src/chocolatey/infrastructure.app/nuget/NuGetEndpointResources.cs
```diff
@@ -153,7 +153,33 @@ private T ResolveResource<T>()
...
+                    // Typically the deepest/last error in the InnerExceptions will be a relevant and
+                    // actionable error.
+                    if (error != ex.InnerException && message != null)
+                    {
+                        this.Log().Warn(message);
+                    }
+
+                    this.Log().Warn("For more information on this issue and guidance in resolving the problem, see https://ch0.co/t/svcidx");
+
                     _resolvingFailed = true;
```
corbob, line 183:
```suggestion
                    this.Log().Warn("For more information on this issue and guidance in resolving the problem, see https://ch0.co/t/svcidx");
                    _resolvingFailed = true;
```

This empty line isn't needed is it?

src/chocolatey/infrastructure.app/nuget/NuGetEndpointResources.cs
```diff
@@ -153,7 +153,33 @@ private T ResolveResource<T>()
...
                 if (!_resolvingFailed)
                 {
-                    this.Log().Warn(ex.InnerException.Message);
+                    // Unwrap the AggregateException as its surface message is useless
+                    Exception error = ex.InnerException;
+                    this.Log().Warn(error.Message);
+
+                    // Enumerate the inner exceptions, log all but the last one in the list to debug
+                    string message = null;
+                    foreach (var err in error.InnerException.Enumerate())
```
corbob, line 162:
```suggestion
                    this.Log().Warn(error.Message);
                    // Enumerate the inner exceptions, log all but the last one in the list to debug
                    string message = null;

                    foreach (var err in error.InnerException.Enumerate())
```
This is more in line with the coding style right? Although, I feel like the other one is more readable...

corbob (changes requested):
Just a few formatting things.

src/chocolatey/chocolatey.csproj
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
corbob: Looking at the csproj files in VS Code, this is the only one that was UTF-8 without a BOM. Given that all of the other ones are that way, and everything worked on Team City, I'm going to leave this formatting change, but with a maint commit.

src/chocolatey/infrastructure.app/nuget/NuGetEndpointResources.cs
corbob: I've put this formatting change into the commit :)

src/chocolatey/infrastructure.app/nuget/NuGetEndpointResources.cs
corbob: The more I look at this, the more I don't like my suggestion here. The string is very much a part of the foreach loop, and the comment is the start of it. Disregard this requested change 👍

corbob (approved):

corbob:
Thanks for getting this sorted out @vexx32 👍

PR #3435, opened 2024-04-29
(#3433) Optimize queries used when resolving dependencies of exact versions

### LINKED ISSUE #3433: Improve how dependencies are looked up when an exact version is defined
### Checklist

When a package has dependencies and we are installing these, the dependencies are looked up by querying all of the available versions for those dependencies.

This causes unnecessary length of time and issues in some cases.

An additional problem is that when the package being installed has a dependency on a package in the moderation queue or Chocolatey Community Repository, where the package attempted to be installed is approved/exempted, but the dependency has not been approved/exempted, the installation will fail because of this. Even when a specific version is attempted.

We should update the dependency lookup to be a bit smarter.

When a version range with an inclusive upper limit is specified, then attempt to look up this exact version first. This would cover both inclusive upper limits, and exact version ranges so we do not have to do additional larger queries.

If there is no upper limit then we should attempt to look up the absolute latest version of the package (stable or pre-release dependending on configuration).

When an exclusive upper limit is used, or if the previous two checks fails it should fall back to the grabbing all the available versions of the dependency.

### Additional Context

N/A

### Related Issues

_No response_

AdmiringWorm:
## Description Of Changes

This updates the NuGetCommon class to better handle lookups of package dependencies when an exact version range is used as the dependency.

Additionally, some optimizations when looking up valid packages are also implemented, by only look up the dependencies of a dependency that is inside of the valid version range that the package specifies.

## Motivation and Context

Some optimizations, and better handling of exact version ranges. Additionally, without this change users of CCR will not be able to install a meta package and its dependency until both have been approved or exempted.

## Testing

Testing have been performed against an internal testing repository which will be called out as hermes.

1. `choco install hasbetadependency --version 1.1.0-beta --ignore-http-cache --verbose --source hermes`
2. Verify queries used for `hasbetadependency` and `isexactversiondependency` is looked up using a specific version.
3. Repeat 1 and 2, using `--pre` instead of `--version`.
4. Run `choco install toplevelhasexactversiondependency --ignore-http-cache --verbose --source hermes`
5. Verify queries used for `toplevelhasexactversiondependency` and `childdependencywithlooserversiondependency` is looked up using a specific verison.
6. Run `choco install toplevelwithnesteddependencies --ignore-http-cache --verbose --source hermes`
7. Verify queries used for `toplevelwithnesteddependencies` and `childdependencywithlooserversiondependency` is looked up using a specific version.
8. Run `choco install hasoutofrangedependency --version 2.0.2 --ignore-http-cache --verbose --source hermes`
9. Verify queries used for `hasoutofrangedependency` and `hasdependency` is looked up using a specific version.
10. Verify the query for `hasdependency` falled back to attempting the `FindPackagesById` endpoint.

### Operating Systems Testing

- Windows 10

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

* [ ] Tests to cover my changes, have been added. (Not really any tests that can be added)

## Related Issue

Fixes #3433

### DIFF
--- a/src/chocolatey/infrastructure.app/nuget/NugetCommon.cs
+++ b/src/chocolatey/infrastructure.app/nuget/NugetCommon.cs
@@ -371,21 +371,36 @@ namespace chocolatey.infrastructure.app.nuget
             ISet<PackageDependency> dependencyCache,
             ChocolateyConfiguration configuration)
         {
-            if (availablePackages.Contains(package))
+            foreach (var resource in resources)
             {
-                return;
-            }
+                var dependencyInfoResource = resource.DependencyInfoResource;
 
-            var dependencyInfoResources = resources.DependencyInfoResources();
+                if (dependencyInfoResource is null)
+                {
+                    // We can't lookup any dependencies using this resource.
+                    continue;
+                }
 
-            foreach (var dependencyInfoResource in dependencyInfoResources)
-            {
-                SourcePackageDependencyInfo dependencyInfo = null;
+                var dependencyInfo = availablePackages.FirstOrDefault(p => string.Equals(p.Id, package.Id, StringComparison.OrdinalIgnoreCase) && p.Version == package.Version && p.Source == resource.Source);
 
                 try
                 {
-                    dependencyInfo = await dependencyInfoResource.ResolvePackage(
-                        package, framework, cacheContext, logger, CancellationToken.None);
+                    if (dependencyInfo is null && !package.HasVersion)
+                    {
+                        var latestPackage = NugetList.FindPackage(package.Id, configuration, logger, cacheContext, new[] { resource });
+
+                        dependencyInfo = availablePackages.FirstOrDefault(p => string.Equals(p.Id, latestPackage.Identity.Id, StringComparison.OrdinalIgnoreCase) && p.Version == latestPackage.Identity.Version && p.Source == resource.Source);
+
+                        if (dependencyInfo is null)
+                        {
+                            dependencyInfo = await dependencyInfoResource.ResolvePackage(latestPackage.Identity, framework, cacheContext, logger, CancellationToken.None);
+                        }
+                    }
+                    else if (dependencyInfo is null)
+                    {
+                        dependencyInfo = await dependencyInfoResource.ResolvePackage(
+                            package, framework, cacheContext, logger, CancellationToken.None);
+                    }
                 }
                 catch (AggregateException ex) when (!(ex.InnerException is null))
                 {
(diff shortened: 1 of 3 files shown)


src/chocolatey/infrastructure.app/nuget/NugetCommon.cs
```diff
@@ -49,6 +49,7 @@
 using Console = chocolatey.infrastructure.adapters.Console;
 using Environment = chocolatey.infrastructure.adapters.Environment;
 using System.Collections.Concurrent;
+using System.Windows.Forms;
```
gep13, line 52: I don't think this is required, is it?

src/chocolatey/infrastructure.app/services/NugetService.cs
gep13, line 1: Looks like the line-endings on this file have changed.  If they need to be changed, can you do it as a serparte commit?

gep13 (changes requested):

src/chocolatey/infrastructure.app/services/NugetService.cs
AdmiringWorm: I have reverted the line endings changes, while I would like to keep them in, it would just cause us to go back and forth on changes until we have a gitattributes file (back and forth due to possibly people having different global git configurations).

src/chocolatey/infrastructure.app/nuget/NugetCommon.cs
AdmiringWorm: It is not, no.

I have removed this using statement.

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -641,7 +641,10 @@ public void InstallDryRun(ChocolateyConfiguration config, Action<PackageResult,
                     continue;
                 }
 
+                var oldPrerelease = config.Prerelease;
+                config.Prerelease = config.Prerelease || availablePackage.Identity.Version.IsPrerelease;
                 NugetCommon.GetPackageDependencies(availablePackage.Identity, NuGetFramework.AnyFramework, sourceCacheContext, _nugetLogger, remoteEndpoints, sourcePackageDependencyInfos, new HashSet<PackageDependency>(), config).GetAwaiter().GetResult();
+                oldPrerelease = config.Prerelease;
```
gep13, line 647: Shouldn't this be:

```
config.Prerelease = oldPrerelease
```

gep13: Change was made during pairing session.

PR #3568, opened 2024-11-19
(#3565 #2421) Avoid credential bleed from saved sources with the same hostname

### LINKED ISSUE #2421: Chocolatey cannot find package from the second source (different credentials)
### What You Are Seeing?
- **Chocolatey only find packages from the first source** (the source with the lowest priority "1") and don't find any packages from the other configured sources.
- **Installation of packages will fail** if the package is not on the first source (because Chocolatey don't find the package).
- **The command `choco list` only shows the package from the first source.**

#### Example: List/Searching a package
As you can see if I specify the source (`-s choco_ix-test_shared`) everything is working, but if I list/search overall sources Chocolatey cannot find the package:
image

### What is Expected?
- Chocolatey finds all packages with `choco list` no matter on which source the packages are.
- Chocolatey can install packages from other sources then only the first source.

- On jFrog Artifactory: create 2x NuGet repositories **with different users and passwords**.
- Upload a package on one of the newly created NuGet repositories and another packages on the other repository.
- Configure this two repositories as Chocolatey sources. One with "Priority=1" and the other with "Priority=2".
- Try to run `choco list` and you get only one package instead of both packages (the package from the source with the lower priority).

### Other Informations
- This issue only occurs if the two sources uses **different credentials**, unfortunately we have to use different credentials!
- In the jFrog Artifactory logs we can see Chocolatey tries to authenticate **always with the credentials from the first source instead of using the configured credentials from each source**.
- This issue don't occurs with Nexus as repository, but unfortunately we have to use jFrog Artifactory.
- This issue occurs with Chocolatey v0.10.15 & v0.11.2!
- With NugetCLI 5.11.0.10 everything works
image

_We already had a Chocolatey Case open for this issue, but this case does not help or fix this issue. Please fix this issue._

### LINKED ISSUE #3565: Credentials from configured source used for a non-configured source when the URL is similar
### Checklist

- [x] I confirm there are no unresolved issues reported on the [Chocolatey Status page](https://status.chocolatey.org).
- [x] I have verified no other issues exist related to my problem.
- [x] I have verified this is not an issue for a specific package.
- [x] I have verified this issue is not security related.
- [x] I confirm I am using **official**, and **not** unofficial, or modified, Chocolatey products.

### What You Are Seeing?

When performing package operations against a source, Chocolatey CLI may choose to use credentials stored against a different source.

### What is Expected?

If credentials are not provided, Chocolatey CLI should not use credentials stored against a different source.

1. Install Chocolatey on a machine\
   `Set-ExecutionPolicy Unrestricted Process -Force; irm https://ch0.co/go | iex`
2. Install Sonatype Nexus Repository ('Nexus') on a machine (note: also installing NexuShell)\
   `choco install nexus-repository nexushell --confirm`\
   a. Log in to Nexus\
   `Connect-NexusServer -Hostname LocalHost -Credential ([PSCredential]::new('admin', (Get-Content C:\ProgramData\sonatype-work\nexus3\admin.password | ConvertTo-SecureString -Force -AsPlainText)))`\
   b. Create ChocolateyInternal NuGet repository\
   `New-NexusRepository -Name ChocolateyInternal -Format nuget -Type hosted -DeploymentPolicy Allow`\
   c. Create AdminOnly NuGet repository\
   `New-NexusRepository -Name AdminOnly -Format nuget -Type hosted -DeploymentPolicy Allow`\
   d. Push a package to the AdminOnly repository\
   `$ApiKey = (Get-NexusNuGetApiKey -Credential ([PSCredential]::new('admin', (Get-Content C:\ProgramData\sonatype-work\nexus3\admin.password | ConvertTo-SecureString -Force -AsPlainText)))).apiKey; choco push $env:ChocolateyInstall\lib\nexus-repository\nexus-repository.nupkg --source http://localhost:8081/repository/adminOnly/index.json --api-key="$ApiKey"`\
   e. Turn off anonymous access (last, because otherwise you [can't push the package](https://github.com/chocolatey/choco/issues/2026))\
   `Set-NexusAnonymousAuth -Disabled`
3. Add the ChocolateyInternal repository to the Chocolatey install, including the valid username and password you are using for Nexus.\
   `choco source add -n=ChocoInternal --source=http://localhost:8081/repository/ChocolateyInternal/index.json -U=admin -P="$(Get-Content C:\ProgramData\sonatype-work\nexus3\admin.password)"`
4. `choco search --source $AdminOnlyFeedUrl`, where the AdminOnlyFeedUrl is the URL of the AdminOnly repository on Nexus, without specifying credentials.\
   `choco search --source http://localhost:8081/repository/adminOnly/index.json`

At the end of the configuration, the server should be configured as follows:

- Nexus installed
- Anonymous authentication turned off, an account ready to be used
- Two sources created
- A package pushed to AdminOnly
- The ChocolateyInternal authenticated source added to Chocolatey CLI
- The AdminOnly source _not_ added to Chocolatey CLI

### System Details

- Operating System: Windows Server 2022
- Windows PowerShell Version: 5.1.20348.2849
- Chocolatey CLI Version: 2.4.0
- Chocolatey Licensed Extension version: 6.2.1 (but seemingly N/A)
- Chocolatey License type (Professional / Business / ?): Business
- Terminal/Emulator: bare PowerShell.exe

### Installed Packages

```bash
N/A
```

### Output Log

See [GitLab issue](https://gitlab.com/chocolatey/solutions/support-issues/-/issues/107).

### Additional Context

- [Zendesk Ticket](https://chocolatey.zendesk.com/agent/tickets/24334)

vexx32:
## Description Of Changes

- Rework handling in ChocolateyNugetCredentialProvider to ensure we only match credentials from the full URL
- Add a config property so we can actually tell what the literal arguments passed to `--source` on the command line are, and start making use of it here to aid with looking up source credentials.
  - In discussions with Gary, we have agreed the behaviour of looking up a source based purely on the URL here is bad In General, because choco should probably just use the literal command line arguments and not assume you want it to look things up in the config when you did not ask, so this will later be expanded upon and some of the searching functionality here can be removed at a later date.
- Add unit tests for this credential provider

## Motivation and Context

See #3565 - the credentials are being reused in places that they shouldn't.

## Testing

Unit tests! :3 Including a regression test specifically for #3565

Open VS and run the tests in the `chocolatey.tests` project and make sure they pass.

Cory added: Ran `CredentialProvider` tag in Test Kitchen.

### Operating Systems Testing

Win10

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

Fixes #3565
Fixes #2421 (different symptom, same issue, likely _somewhat_ duplicate issue)

### DIFF
--- a/src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
+++ b/src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
@@ -70,10 +70,9 @@ namespace chocolatey.tests.infrastructure.app.nuget
                 };
             }
 
-            [OneTimeSetUp]
-            public async Task OneTimeSetup()
+            public override void Because()
             {
-                await With();
+                With().Wait();
             }
 
             public virtual async Task With()
@@ -83,91 +82,95 @@ namespace chocolatey.tests.infrastructure.app.nuget
             }
         }
 
-        public class When_using_explicit_credentials_and_source_param : ChocolateyNugetCredentialProviderSpecsBase
+        public class When_Using_Explicit_Credentials_And_Source_Param : ChocolateyNugetCredentialProviderSpecsBase
         {
-            public override void Because()
+            public override void Context()
             {
+                base.Context();
                 Configuration.Sources = Configuration.ExplicitSources = TargetSourceUrl;
                 Configuration.SourceCommand.Username = "user";
                 Configuration.SourceCommand.Password = "totally_secure_password!!!";
             }
 
             [Fact]
-            public void Creates_a_valid_credential()
+            public void Should_Create_Credentials()
             {
                 Result.Should().NotBeNull();
             }
 
             [Fact]
-            public void Provides_the_correct_username()
+            public void Should_Provide_The_Correct_Username()
             {
                 Result.UserName.Should().Be("user");
             }
 
             [Fact]
-            public void Provides_the_correct_password()
+            public void Should_Provide_The_Correct_Password()
             {
                 Result.Password.Should().Be("totally_secure_password!!!");
             }
         }
 
-        public class When_a_source_name_is_provided : ChocolateyNugetCredentialProviderSpecsBase
+        public class When_A_Source_Name_Is_Provided : ChocolateyNugetCredentialProviderSpecsBase
         {
-            public override void Because()
+            public override void Context()
             {
+                base.Context();
                 Configuration.Sources = TargetSourceUrl;
                 Configuration.ExplicitSources = TargetSourceName;
             }
 
             [Fact]
-            public void Finds_the_saved_source_and_returns_the_credential()
+            public void Should_Find_The_Saved_Source_And_Returns_The_Credential()
             {
                 Result.Should().NotBeNull();
             }
 
             [Fact]
-            public void Provides_the_correct_username()
+            public void Should_Provide_The_Correct_Username()
             {
                 Result.UserName.Should().Be(Username);
             }
 
             [Fact]
-            public void Provides_the_correct_password()
+            public void Should_Provide_The_Correct_Password()
             {
                 Result.Password.Should().Be(Password);
             }
         }
 
-        public class When_a_source_url_matching_a_saved_source_is_provided : ChocolateyNugetCredentialProviderSpecsBase
+        public class When_A_Source_Url_Matching_A_Saved_Source_Is_Provided : ChocolateyNugetCredentialProviderSpecsBase
         {
-            public override void Because()
+            public override void Context()
             {
+                base.Context();
                 Configuration.Sources = Configuration.ExplicitSources = TargetSourceUrl;
             }
 
             [Fact]
-            public void Finds_the_saved_source_and_returns_the_credential()
+            public void Should_Find_The_Saved_Source_And_Return_The_Credential()
             {
                 Result.Should().NotBeNull();
             }
 
             [Fact]
-            public void Provides_the_correct_username()
+            public void Should_Provide_The_Correct_Username()
             {
                 Result.UserName.Should().Be(Username);
             }
 
             [Fact]
-            public void Provides_the_correct_password()
+            public void Should_Provide_The_Correct_Password()
             {
                 Result.Password.Should().Be(Password);
             }
         }
 
-        public class Looks_up_source_url_when_name_and_credentials_is_provided : ChocolateyNugetCredentialProviderSpecsBase
+        public class Looks_Up_Source_Url_When_Name_And_Credentials_Is_Provided : ChocolateyNugetCredentialProviderSpecsBase
         {
-            public override void Because()
+            public override void Context()
             {
+                base.Context();
                 Configuration.Sources = TargetSourceUrl;
                 Configuration.ExplicitSources = TargetSourceName;
 
@@ -176,28 +179,29 @@ namespace chocolatey.tests.infrastructure.app.nuget
             }
 
             [Fact]
-            public void Creates_and_returns_the_credential()
+            public void Should_Create_And_Return_The_Credential()
             {
                 Result.Should().NotBeNull();
             }
 
             [Fact]
-            public void Provides_the_correct_username()
+            public void Should_Provide_The_Correct_Username()
             {
                 Result.UserName.Should().Be(Username);
             }
 
             [Fact]
-            public void Provides_the_correct_password()
+            public void Should_Provide_The_Correct_Password()
             {
                 Result.Password.Should().Be(Password);
             }
         }
 
-        public class When_no_matching_source_is_found_by_url : ChocolateyNugetCredentialProviderSpecsBase
+        public class When_No_Matching_Source_Is_Found_By_Url : ChocolateyNugetCredentialProviderSpecsBase
         {
-            public override void Because()
+            public override void Context()
             {
+                base.Context();
                 Configuration.Sources = Configuration.ExplicitSources = "https://unknownurl.com/api/v2/";
             }
 
@@ -208,98 +212,102 @@ namespace chocolatey.tests.infrastructure.app.nuget
             }
 
             [Fact]
-            public void Returns_the_default_network_credential()
+            public void Should_Return_The_Default_Network_Credential()
             {
                 Result.Should().Be(CredentialCache.DefaultNetworkCredentials);
             }
         }
 
-        public class When_multiple_named_sources_are_provided : ChocolateyNugetCredentialProviderSpecsBase
+        public class When_Multiple_Named_Sources_Are_Provided : ChocolateyNugetCredentialProviderSpecsBase
         {
-            public override void Because()
+            public override void Context()
             {
+                base.Context();
                 Configuration.Sources = Configuration.MachineSources.Select(s => s.Key).Join(";");
                 Configuration.ExplicitSources = Configuration.MachineSources.Select(s => s.Name).Join(";");
             }
 
             [Fact]
-            public void Finds_the_correct_saved_source_for_the_target_uri_and_returns_the_credential()
+            public void Should_Find_The_Correct_Saved_Source_For_The_Target_Uri_And_Return_The_Credential()
             {
                 Result.Should().NotBeNull();
             }
 
             [Fact]
-            public void Provides_the_correct_username()
+            public void Should_Provide_The_Correct_Username()
             {
                 Result.UserName.Should().Be(Username);
             }
 
             [Fact]
-            public void Provides_the_correct_password()
+            public void Should_Provide_The_Correct_Password()
             {
                 Result.Password.Should().Be(Password);
             }
         }
 
-        public class When_multiple_source_urls_are_provided : ChocolateyNugetCredentialProviderSpecsBase
+        public class When_Multiple_Source_Urls_Are_Provided : ChocolateyNugetCredentialProviderSpecsBase
         {
-            public override void Because()
+            public override void Context()
             {
+                base.Context();
                 Configuration.Sources = Configuration.ExplicitSources = $"https://unknownurl.com/api/v2/;{TargetSourceUrl}";
             }
 
             [Fact]
-            public void Finds_the_saved_source_and_returns_the_credential()
+            public void Should_Find_The_Saved_Source_And_Return_The_Credential()
             {
                 Result.Should().NotBeNull();
             }
 
             [Fact]
-            public void Provides_the_correct_username()
+            public void Should_Provide_The_Correct_Username()
             {
                 Result.UserName.Should().Be(Username);
             }
 
             [Fact]
-            public void Provides_the_correct_password()
+            public void Should_Provide_The_Correct_Password()
             {
                 Result.Password.Should().Be(Password);
             }
         }
 
-        public class When_a_mix_of_named_and_url_sources_are_provided : ChocolateyNugetCredentialProviderSpecsBase
+        public class When_A_Mix_Of_Named_And_Url_Sources_Are_Provided : ChocolateyNugetCredentialProviderSpecsBase
         {
-            public override void Because()
+            public override void Context()
             {
+                base.Context();
                 Configuration.Sources = $"https://unknownurl.com/api/v2/;{TargetSourceUrl}";
                 Configuration.ExplicitSources = $"https://unknownurl.com/api/v2/;{TargetSourceName}";
             }
 
             [Fact]
-            public void Finds_the_saved_source_and_returns_the_credential()
+            public void Should_Find_The_Saved_Source_And_Return_The_Credential()
             {
                 Result.Should().NotBeNull();
             }
 
             [Fact]
-            public void Provides_the_correct_username()
+            public void Should_Provide_The_Correct_Username()
             {
                 Result.UserName.Should().Be(Username);
             }
 
             [Fact]
-            public void Provides_the_correct_password()
+            public void Should_Provide_The_Correct_Password()
             {
                 Result.Password.Should().Be(Password);
             }
         }
 
         // This is a regression test for issue #3565
-        public class When_a_url_matching_the_hostname_only_of_a_saved_source_is_provided : ChocolateyNugetCredentialProviderSpecsBase
+        public class When_A_Url_Matching_The_Hostname_Only_Of_A_Saved_Source_Is_Provided : ChocolateyNugetCredentialProviderSpecsBase
         {
             private Uri _otherRepoUri;
-            public override void Because()
+            public override void Context()
             {
+                base.Context();
                 _otherRepoUri = new Uri(TargetSourceUri.GetComponents(UriComponents.SchemeAndServer, UriFormat.Unescaped) + "/other_path/repository/");
                 Configuration.Sources = Configuration.ExplicitSources = _otherRepoUri.AbsoluteUri;
             }
@@ -311,7 +319,7 @@ namespace chocolatey.tests.infrastructure.app.nuget
             }
 
             [Fact]
-            public void Returns_the_default_network_credential()
+            public void Should_Return_The_Default_Network_Credential()
             {
                 Result.Should().Be(CredentialCache.DefaultNetworkCredentials);
             }
--- a/src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
+++ b/src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
@@ -1,0 +1,320 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using chocolatey.infrastructure.app.configuration;
+using chocolatey.infrastructure.app.nuget;
+using FluentAssertions;
+using Moq;
+using NuGet.Configuration;
+using NUnit.Framework;
+
+namespace chocolatey.tests.infrastructure.app.nuget
+{
+    public class ChocolateyNugetCredentialProviderSpecs
+    {
+        public abstract class ChocolateyNugetCredentialProviderSpecsBase : TinySpec
+        {
+            protected ChocolateyConfiguration Configuration;
+            protected ChocolateyNugetCredentialProvider Provider;
+
+            protected const string Username = "user";
+            protected const string Password = "totally_secure_password!!!";
+            protected const string TargetSourceName = "testsource";
+            protected const string TargetSourceUrl = "https://testserver.org/repository/test-repository";
+            protected Uri TargetSourceUri = new Uri(TargetSourceUrl);
+
+            protected NetworkCredential Result;
+
+            private readonly CancellationTokenSource _tokenSource = new CancellationTokenSource();
+            protected CancellationToken CancellationToken
+            {
+                get
+                {
+                    return _tokenSource.Token;
+                }
+            }
+
+            public override void Context()
+            {
+                Configuration = new ChocolateyConfiguration();
+                Provider = new ChocolateyNugetCredentialProvider(Configuration);
+
+                Configuration.Information.IsInteractive = false;
+                Configuration.MachineSources = new List<MachineSourceConfiguration>
+                {
+                    new MachineSourceConfiguration
+                    {
+                        AllowSelfService = true,
+                        VisibleToAdminsOnly = false,
+                        EncryptedPassword = NugetEncryptionUtility.EncryptString("otherPassword"),
+                        Username = "otherUserName",
+                        Key = "https://someotherplace.com/repository/things/",
+                        Name = "not-this-one",
+                        Priority = 1,
+                    },
+                    new MachineSourceConfiguration
+                    {
+                        AllowSelfService = true,
+                        VisibleToAdminsOnly = false,
+                        EncryptedPassword = NugetEncryptionUtility.EncryptString(Password),
+                        Username = Username,
+                        Key = TargetSourceUrl,
+                        Name = TargetSourceName,
+                        Priority = 1,
+                    },
+                };
+            }
+
+            [OneTimeSetUp]
+            public async Task OneTimeSetup()
+            {
+                await With();
+            }
+
+            public virtual async Task With()
+            {
+                var result = await Provider.GetAsync(TargetSourceUri, proxy: null, CredentialRequestType.Unauthorized, message: string.Empty, isRetry: false, nonInteractive: true, CancellationToken);
+                Result = result.Credentials as NetworkCredential;
+            }
+        }
+
+        public class When_using_explicit_credentials_and_source_param : ChocolateyNugetCredentialProviderSpecsBase
+        {
+            public override void Because()
+            {
+                Configuration.Sources = Configuration.ExplicitSources = TargetSourceUrl;
+                Configuration.SourceCommand.Username = "user";
+                Configuration.SourceCommand.Password = "totally_secure_password!!!";
+            }
+
+            [Fact]
+            public void Creates_a_valid_credential()
+            {
+                Result.Should().NotBeNull();
+            }
+
+            [Fact]
+            public void Provides_the_correct_username()
+            {
+                Result.UserName.Should().Be("user");
+            }
+
+            [Fact]
+            public void Provides_the_correct_password()
+            {
+                Result.Password.Should().Be("totally_secure_password!!!");
+            }
+        }
+
+        public class When_a_source_name_is_provided : ChocolateyNugetCredentialProviderSpecsBase
+        {
+            public override void Because()
+            {
+                Configuration.Sources = TargetSourceUrl;
+                Configuration.ExplicitSources = TargetSourceName;
+            }
+
+            [Fact]
+            public void Finds_the_saved_source_and_returns_the_credential()
+            {
+                Result.Should().NotBeNull();
+            }
+
+            [Fact]
+            public void Provides_the_correct_username()
+            {
+                Result.UserName.Should().Be(Username);
+            }
+
+            [Fact]
+            public void Provides_the_correct_password()
+            {
+                Result.Password.Should().Be(Password);
+            }
+        }
+
+        public class When_a_source_url_matching_a_saved_source_is_provided : ChocolateyNugetCredentialProviderSpecsBase
+        {
+            public override void Because()
+            {
+                Configuration.Sources = Configuration.ExplicitSources = TargetSourceUrl;
+            }
+
+            [Fact]
+            public void Finds_the_saved_source_and_returns_the_credential()
+            {
+                Result.Should().NotBeNull();
+            }
+
+            [Fact]
+            public void Provides_the_correct_username()
+            {
+                Result.UserName.Should().Be(Username);
+            }
+
+            [Fact]
+            public void Provides_the_correct_password()
+            {
+                Result.Password.Should().Be(Password);
+            }
+        }
+
+        public class Looks_up_source_url_when_name_and_credentials_is_provided : ChocolateyNugetCredentialProviderSpecsBase
+        {
+            public override void Because()
+            {
+                Configuration.Sources = TargetSourceUrl;
+                Configuration.ExplicitSources = TargetSourceName;
+
+                Configuration.SourceCommand.Username = "user";
+                Configuration.SourceCommand.Password = "totally_secure_password!!!";
+            }
+
+            [Fact]
+            public void Creates_and_returns_the_credential()
+            {
+                Result.Should().NotBeNull();
+            }
+
+            [Fact]
+            public void Provides_the_correct_username()
+            {
+                Result.UserName.Should().Be(Username);
+            }
+
+            [Fact]
+            public void Provides_the_correct_password()
+            {
+                Result.Password.Should().Be(Password);
+            }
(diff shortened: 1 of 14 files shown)


src/chocolatey/infrastructure.app/commands/ChocolateyInfoCommand.cs
```diff
@@ -38,7 +38,10 @@ public override void ConfigureArgumentParser(OptionSet optionSet, ChocolateyConf
                 .Add(
                     "s=|source=",
                     "Source - Source location for install. Can use special 'windowsfeatures', 'ruby', 'cygwin', or 'python' sources. Defaults to configured sources.",
-                    option => configuration.Sources = option.UnquoteSafe())
+                    option => {
+                        configuration.Sources = configuration.ExplicitSources = option.UnquoteSafe();
+                        configuration.ListCommand.ExplicitSource = true;
```
corbob, line 43: This is not for this PR, but I need to capture it so I don't forget: Chocolatey CLI 1.x does not have an `ExplicitSource` property. We will want to consider if that part is needed for fixing it in the support branch.

corbob:
I've updated the Pester tests once I got them sorted in Test Kitchen. This is once again ready for review.

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
```diff
@@ -0,0 +1,320 @@
...
+            }
+        }
+
+        public class When_a_source_name_is_provided : ChocolateyNugetCredentialProviderSpecsBase
+        {
+            public override void Because()
+            {
+                Configuration.Sources = TargetSourceUrl;
+                Configuration.ExplicitSources = TargetSourceName;
+            }
```
AdmiringWorm, line 120: This should be part of the context (ie the setup), not the because method.

```suggestion
            public override void Context()
            {
                base.Context();
                Configuration.Sources = TargetSourceUrl;
                Configuration.ExplicitSources = TargetSourceName;
            }

```

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
```diff
@@ -0,0 +1,320 @@
...
+                Result.Password.Should().Be(Password);
+            }
+        }
+
+        public class When_a_source_url_matching_a_saved_source_is_provided : ChocolateyNugetCredentialProviderSpecsBase
+        {
+            public override void Because()
+            {
+                Configuration.Sources = Configuration.ExplicitSources = TargetSourceUrl;
+            }
```
AdmiringWorm, line 146: Again, should be Context.

```suggestion
            public override void Context()
            {
                base.Context();
                Configuration.Sources = Configuration.ExplicitSources = TargetSourceUrl;
            }

```

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
```diff
@@ -0,0 +1,320 @@
...
+        public class Looks_up_source_url_when_name_and_credentials_is_provided : ChocolateyNugetCredentialProviderSpecsBase
+        {
+            public override void Because()
+            {
+                Configuration.Sources = TargetSourceUrl;
+                Configuration.ExplicitSources = TargetSourceName;
+
+                Configuration.SourceCommand.Username = "user";
+                Configuration.SourceCommand.Password = "totally_secure_password!!!";
+            }
```
AdmiringWorm, line 176:
```suggestion
            public override void Context()
            {
                base.Context();
                Configuration.Sources = TargetSourceUrl;
                Configuration.ExplicitSources = TargetSourceName;

                Configuration.SourceCommand.Username = "user";
                Configuration.SourceCommand.Password = "totally_secure_password!!!";
            }

```

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        public class When_multiple_named_sources_are_provided : ChocolateyNugetCredentialProviderSpecsBase
+        {
+            public override void Because()
+            {
+                Configuration.Sources = Configuration.MachineSources.Select(s => s.Key).Join(";");
+                Configuration.ExplicitSources = Configuration.MachineSources.Select(s => s.Name).Join(";");
+            }
```
AdmiringWorm, line 223:
```suggestion
            public override void Context()
            {
                base.Context();
                Configuration.Sources = Configuration.MachineSources.Select(s => s.Key).Join(";");
                Configuration.ExplicitSources = Configuration.MachineSources.Select(s => s.Name).Join(";");
            }

```

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
```diff
@@ -0,0 +1,320 @@
...
+                Result.Password.Should().Be(Password);
+            }
+        }
+
+        public class When_multiple_source_urls_are_provided : ChocolateyNugetCredentialProviderSpecsBase
+        {
+            public override void Because()
+            {
+                Configuration.Sources = Configuration.ExplicitSources = $"https://unknownurl.com/api/v2/;{TargetSourceUrl}";
+            }
```
AdmiringWorm, line 249:
```suggestion
            public override void Context()
            {
                base.Context();
                Configuration.Sources = Configuration.ExplicitSources = $"https://unknownurl.com/api/v2/;{TargetSourceUrl}";
            }

```

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        public class When_a_mix_of_named_and_url_sources_are_provided : ChocolateyNugetCredentialProviderSpecsBase
+        {
+            public override void Because()
+            {
+                Configuration.Sources = $"https://unknownurl.com/api/v2/;{TargetSourceUrl}";
+                Configuration.ExplicitSources = $"https://unknownurl.com/api/v2/;{TargetSourceName}";
+            }
```
AdmiringWorm, line 276:
```suggestion
            public override void Context()
            {
                base.Context();
                Configuration.Sources = $"https://unknownurl.com/api/v2/;{TargetSourceUrl}";
                Configuration.ExplicitSources = $"https://unknownurl.com/api/v2/;{TargetSourceName}";
            }

```

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
```diff
@@ -0,0 +1,320 @@
...
+
+        // This is a regression test for issue #3565
+        public class When_a_url_matching_the_hostname_only_of_a_saved_source_is_provided : ChocolateyNugetCredentialProviderSpecsBase
+        {
+            private Uri _otherRepoUri;
+            public override void Because()
+            {
+                _otherRepoUri = new Uri(TargetSourceUri.GetComponents(UriComponents.SchemeAndServer, UriFormat.Unescaped) + "/other_path/repository/");
+                Configuration.Sources = Configuration.ExplicitSources = _otherRepoUri.AbsoluteUri;
+            }
```
AdmiringWorm, line 305:
```suggestion
            public override void Context()
            {
                base.Context();
                _otherRepoUri = new Uri(TargetSourceUri.GetComponents(UriComponents.SchemeAndServer, UriFormat.Unescaped) + "/other_path/repository/");
                Configuration.Sources = Configuration.ExplicitSources = _otherRepoUri.AbsoluteUri;
            }

```

src/chocolatey/infrastructure.app/commands/ChocolateyInfoCommand.cs
AdmiringWorm, line 43: Why are we storing a `Source` and an `ExplicitSource`, at the same time that we are setting a boolean property called `ExplicitSource`?

Also, what happens if the source that is passed is the named source of a configuration? Or one of the alternative sources?
Should that also be considered an explicit source?

src/chocolatey/infrastructure.app/nuget/ChocolateyNugetCredentialProvider.cs
```diff
@@ -86,67 +86,62 @@ public Task<CredentialResponse> GetAsync(Uri uri, IWebProxy proxy, CredentialReq
...
             MachineSourceConfiguration source = null;
 
-
-            if (candidateSources.Count == 1)
+            // If the user has specified --source with a *named* source and not a URL, try to find the matching one
+            // with the correct URL for this credential request.
+            var namedExplicitSources = _config.ExplicitSources?.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries)
+                .Where(s => !Uri.IsWellFormedUriString(s, UriKind.Absolute))
+                .ToList();
+            if (namedExplicitSources?.Count > 0)
```
AdmiringWorm, line 98: Should be a newline here.

```suggestion
                .ToList();

            if (namedExplicitSources?.Count > 0)

```

src/chocolatey/infrastructure.app/nuget/ChocolateyNugetCredentialProvider.cs
```diff
@@ -86,67 +86,62 @@ public Task<CredentialResponse> GetAsync(Uri uri, IWebProxy proxy, CredentialReq
...
             {
-                // only one match, use it
-                source = candidateSources.FirstOrDefault();
+                // Uri.Equals() and == operator compare hostnames case-insensitively and the remainder of the url case-sensitively
+                // while ignoring #fragments on the URLs, but does care about trailing slashes, which we do not here.
+                source = _config.MachineSources
+                    .Where(s => namedExplicitSources.Contains(s.Name) && new Uri(s.Key.TrimEnd('/')) == trimmedTargetUri)
+                    .FirstOrDefault();
```
AdmiringWorm, line 104: Why are we doing case sensitive checking here?

The `Contains` method is case sensitive, and as the comment says it does a case sensitive match on the url (just not the hostname).
That do not make sense, and I would think it is a case insensitive match we would want.

src/chocolatey/infrastructure.app/nuget/ChocolateyNugetCredentialProvider.cs
```diff
@@ -86,67 +86,62 @@ public Task<CredentialResponse> GetAsync(Uri uri, IWebProxy proxy, CredentialReq
...
+
+            if (source is null)
             {
-                // find the source that is the closest match
-                foreach (var candidateSource in candidateSources.OrEmpty())
+                // Could not find a valid source by name, or the source(s) specified were all URLs.
+                // Try to look up the target URL in the saved machine sources to attempt to match credentials.
+                //
+                // Note: This behaviour remains as removing it would be a breaking change, but we may want
+                // to remove this in a future version, as specifying an explicit URL should potentially
```
AdmiringWorm, line 113: If there has been a decision to remove this, or potentially removing it, we would need an issue to be created to keep track of this.

src/chocolatey/infrastructure.app/nuget/ChocolateyNugetCredentialProvider.cs
```diff
@@ -86,67 +86,62 @@ public Task<CredentialResponse> GetAsync(Uri uri, IWebProxy proxy, CredentialReq
...
+                // not go looking in the configuration file for saved credentials anyway.
+                var candidateSources = _config.MachineSources
+                    .Where(s => !string.IsNullOrWhiteSpace(s.Username)
+                        && !string.IsNullOrWhiteSpace(s.EncryptedPassword)
+                        && Uri.TryCreate(s.Key.TrimEnd('/'), UriKind.Absolute, out var trimmedSourceUri)
+                        && trimmedSourceUri == trimmedTargetUri)
```
AdmiringWorm, line 119: Coming back to the comment further up, if this is doing a case-sensitive match on the URL (except the hostname), then this type of comparison is probably not what we would want.

src/chocolatey/infrastructure.app/nuget/ChocolateyNugetCredentialProvider.cs
corbob: The reason for doing a case sensitive match on everything but the hostname is that the hostname is handled by DNS and is case insensitive, but the rest of the URL is up to the responding server which may or may not be case insensitive. For example: `https://docs.chocolatey.org/en-us/central-management/` is valid while `https://docs.chocolatey.org/en-us/central-managemenT/` is not. But if you were to case-insensitive compare them they are identical.

src/chocolatey/infrastructure.app/commands/ChocolateyInfoCommand.cs
corbob: In the credential provider, the first thing it does it try to look up any explicitly set named sources, and then the remainder as URIs. I if it's an alternative source it'll be handled before getting to the credential provider and so will not matter if it's set as an explicit source.

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
```diff
@@ -0,0 +1,320 @@
...
+                        Priority = 1,
+                    },
+                };
+            }
+
+            [OneTimeSetUp]
+            public async Task OneTimeSetup()
+            {
+                await With();
+            }
```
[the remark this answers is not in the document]
corbob: This has been updated.

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        public class When_using_explicit_credentials_and_source_param : ChocolateyNugetCredentialProviderSpecsBase
+        {
+            public override void Because()
+            {
+                Configuration.Sources = Configuration.ExplicitSources = TargetSourceUrl;
+                Configuration.SourceCommand.Username = "user";
+                Configuration.SourceCommand.Password = "totally_secure_password!!!";
+            }
```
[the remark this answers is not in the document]
corbob: This has been updated.

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
```diff
@@ -0,0 +1,320 @@
...
+
+            [Fact]
+            public void Creates_a_valid_credential()
+            {
+                Result.Should().NotBeNull();
+            }
```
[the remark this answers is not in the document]
corbob: This has been updated.

A note on the naming and casing: should it not be PascalCase as well?

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
corbob: This has been updated.

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
corbob: This has been updated.

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
corbob: This has been updated.

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
corbob: This has been updated.

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
corbob: This has been updated.

src/chocolatey/infrastructure.app/nuget/ChocolateyNugetCredentialProvider.cs
corbob: This has been updated.

src/chocolatey/infrastructure.app/commands/ChocolateyInfoCommand.cs
gep13: I am ok with the introduction of the `ExplicitSources` property on the top-level configuration object, but that does bring up the question of whether the `ExplicitSource` boolean is needed anymore on the ListCommand object.  Should we not replace the usage of `ExplicitSource` in the `ChocolateyListCommand` class, and replace if with having anything in the `ExplicitSources` property?  Perhaps mark `ExplicitSource` as obsolete, and then we can remove it in future breaking change?

src/chocolatey/infrastructure.app/nuget/ChocolateyNugetCredentialProvider.cs
gep13: I think we agree that we would add an issue to track this discussion, and point to it in this comment in the code, right?

tests/helpers/common/Chocolatey/Get-ChocolateySource.ps1
```diff
@@ -0,0 +1,14 @@
+﻿function Get-ChocolateySource {
+    [CmdletBinding()]
+    param(
+        [Parameter()]
+        [string]$Name = "*",
+
+        [Parameter()]
+        [switch]$All
```
gep13, line 8: Am I missing where this is being used?

tests/pester-tests/features/CredentialProvider.Tests.ps1
```diff
@@ -0,0 +1,62 @@
...
+        Disable-ChocolateySource -All
+        Enable-ChocolateySource -Name 'hermes'
+        $SetupSource = Get-ChocolateySource -Name 'hermes-setup'
+        Remove-Item download -force -recurse
+    }
+
+    # Skip the download command if chocolatey.extension is not installed.
+    Context 'Command (<Command>)' -Skip:($Command -eq 'download' -and -not $HasLicensedExtension) {
+        BeforeAll {
+            # Picked a package that is on `hermes-setup` but not on `hermes`.
```
gep13, line 42: How are we going to enforce this going forward?

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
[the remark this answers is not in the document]
AdmiringWorm: > A note on the naming and casing: should it not be PascalCase as well?

Not for tests, as the names of tests can become large, it will be easier to read these names when the words are separated by an underscore.

src/chocolatey/infrastructure.app/nuget/ChocolateyNugetCredentialProvider.cs
AdmiringWorm: While that is true for the documentation site we have, that is actually quite uncommon.

If we take the api we have, the URL `https://community.chocolatey.org/api/v2/` returns the exact same information as `https://community.chocolatey.org/API/v2/`.

But even if they were not returning the same information, it would not be the point.

The point is, we are looking up information about a URL in the machine sources configuration, and it is not guaranteed that a user will remember the exact case of the URI that was configured but still intend to use it. By not allowing the case to be different, we are preventing this configured URL to be picked up. We should also use the URL as specified in the machine configuration (and if I am reading everything correctly, we are doing that), so even if the server would accept one case but not the other, it wouldn't be a problem since it already uses the case that is configured on the machine object.

There is also the case that with this change, a user can not use a named source that differs in casing either. Ie, they can not pass the source `Chocolatey` when the name is configured to be `chocolatey`. This is something that is changing the behavior of what is currently available, as currently using a named source with a different case will resolve to a configured URL.

src/chocolatey.tests/infrastructure.app/nuget/ChocolateyNugetCredentialProviderSpecs.cs
corbob: This has been updated.

src/chocolatey/infrastructure.app/commands/ChocolateyInfoCommand.cs
corbob: > Should we not replace the usage of `ExplicitSource` in the `ChocolateyListCommand` class, and replace if with having anything in the `ExplicitSources` property? Perhaps mark `ExplicitSource` as obsolete, and then we can remove it in future breaking change?

If I recall my conversation with Rain correctly, one thing they thought of for `ExplicitSource` on the `ChocolateyListCommand` class would be to do it as a calculated property and return true if the top-level `ExplicitSources` is set. But that change can't be introduced right now.

corbob: That makes sense. Neither Extension nor GUI actually make use of the `ListCommand.ExplicitSource` setting. So if I'm understanding correctly, the idea is to: mark `ListCommand.ExplicitSource` Obsolete, and change where it's used in CLI code.

src/chocolatey/infrastructure.app/nuget/ChocolateyNugetCredentialProvider.cs
corbob: I think I have talked myself into what you're suggesting here. On the topic of the URL case sensitivity, I'll say this: in my experiences, casing differences in URLs outside of the domain name are typically up to the host operating system. I also know that Nexus is partially case sensitive for instance `http://nexus/repository/my-repo/` and `http://nexus/repository/My-RePO/` are both treated the same, but `http://nexus/Repository/my-repo/` results in a 404 error.

But getting to the point of allowing the user to put the source in at the terminal and not have to worry about the casing of it, I think I can get behind that for you reason you said about we're pulling the URL from the machine config, so even if they got the casing wrong, we're still going to use the correct casing.

The named source is still being picked up in a case insensitive matter (at least in my limited testing it was).

tests/helpers/common/Chocolatey/Get-ChocolateySource.ps1
corbob: Nope... I copied it from the Enable/Disable functions... But it doesn't make sense on a `Get` function. `Get-ChildItem` gets me all the child items (except hidden, but that's a different story, and that uses force anyways). I'll take it off of here.

src/chocolatey/infrastructure.app/commands/ChocolateyInfoCommand.cs
gep13: Yes, that is what I am suggesting.

src/chocolatey/infrastructure.app/nuget/ChocolateyNugetCredentialProvider.cs
vexx32: 🤔 I'm not sure how this would be deprecated in any meaningful way, honestly? It's kind of an edge case implementation detail, but. Yeah we can hammer that out in the issue when we get to it. 👍

vexx32: Cory already raised it this morning looks like, so added some comments to it. https://github.com/chocolatey/choco/issues/3573

AdmiringWorm: > 🤔 I'm not sure how this would be deprecated in any meaningful way, honestly? It's kind of an edge case implementation detail, but. Yeah we can hammer that out in the issue when we get to it. 👍

What I meant by deprecating it, would probably be to output a message to the user that the functionality of loading settings for absolute URLs will be removed in the future. (Assuming that is the decision that will be made).

> Cory already raised it this morning looks like, so added some comments to it. https://github.com/chocolatey/choco/issues/3573

👍

tests/pester-tests/features/CredentialProvider.Tests.ps1
corbob: I've updated this comment. The package used doesn't really matter, but it is best if we use one that might be found, so I used `chocolatey`.

tests/helpers/common/Chocolatey/Get-ChocolateySource.ps1
corbob: I took this out since it's not used.

src/chocolatey/infrastructure.app/nuget/ChocolateyNugetCredentialProvider.cs
vexx32: Small wrench to throw in Kim's logic here: this credential provider _does not_ provide the URL back to NuGet. NuGet is still potentially going to be querying an incorrectly-cased URL with the saved credentials, because we don't do this kind of lookup anywhere else when evaluating which sources to query. So this is a narrower case but still potentially the same as the original bug: querying route A with credentials intended for route B on the same server.

That said, given this is still an improvement and it would take a long time to hash that out I'm content to leave that to a follow up issue / let it be until we remove this weird lookup logic entirely (I'll probably put some details around things Gary and I have talked about in the related issue @corbob raised and we can figure out how we want to split things out from there, if needed).

corbob: I have updated this to do a case sensitive comparison of the lower case name (because we can't tell `.Contains` to do a case insensitive compare, and using `string.Join` to join into a string leaves potential for a substring matching where it shouldn't). And then to do a case insensitive compare on the URI Components that the `==` compares on.

src/chocolatey/infrastructure.app/commands/ChocolateyInfoCommand.cs
corbob: I have removed the additions where we were setting this property in the new places, and marked the property obsolete.

tests/pester-tests/features/CredentialProvider.Tests.ps1
corbob: I've re-run the tests. The package needs to be one that is not already installed (otherwise `choco install won't try to install it), but may be found on the repository. As such, I've switched it back to `chocolatey-compatibility.extension` so that the tests all pass.

src/chocolatey/infrastructure.app/nuget/ChocolateyNugetCredentialProvider.cs
```diff
@@ -86,67 +86,67 @@ public Task<CredentialResponse> GetAsync(Uri uri, IWebProxy proxy, CredentialReq
...
+                .Where(s => !Uri.IsWellFormedUriString(s, UriKind.Absolute))
+                .ToList();
 
-            if (candidateSources.Count == 1)
+            if (namedExplicitSources?.Count > 0)
             {
-                // only one match, use it
-                source = candidateSources.FirstOrDefault();
+                // Uri.Equals() and == operator compare hostnames case-insensitively and the remainder of the url case-sensitively
+                // while ignoring #fragments on the URLs, but does care about trailing slashes, which we do not here.
```
vexx32, line 103: Given we're not using .Equals / == here anymore, we should probably remove this comment.

vexx32 (commented):
Looks good to me apart from one minor comment. Thank for picking this up, @corbob :3

tests/pester-tests/features/CredentialProvider.Tests.ps1
```diff
@@ -0,0 +1,63 @@
+﻿Describe 'Ensuring credentials do not bleed from configured sources' -Tag CredentialProvider -ForEach @(
```
gep13, line 1: I personally think that we need some additional words here as to exactly that this test is trying to assert.

Why are we expecting 0 ExitCodes for some commands, and not for others?

If we are expecting that Chocolatey CLI should prompt for credentials, since it is no longer able to be provided credentials from another defined source, why is `choco outdated` not returning an error exit code?  Same question goes for `choco info`.

I think having the extra description about what this test is trying to assert would help to understand this test, without having to read through the whole thing.

AdmiringWorm (approved):
LGTM, but I think the comment from Gary needs to be answered (updated) before anything is merged.

tests/pester-tests/features/CredentialProvider.Tests.ps1
corbob: I've updated this to have some comments about what it's testing and why the exit codes for info and outdated are 0. I also re-ordered them to put info and outdated beside each other to avoid duplicating the comment.

tests/pester-tests/features/CredentialProvider.Tests.ps1
```diff
@@ -0,0 +1,70 @@
+﻿# These tests are to ensure that credentials from one configured and enabled source are not
+# picked up and used when a URL is matching based on the hostname. These tests use an authenticated
+# source without explicitly providing a username/password. It is expected that Chocolatey with prompt for
```
gep13, line 3:
```suggestion
# source without explicitly providing a username/password. It is expected that Chocolatey will prompt for

```

gep13 (changes requested):
Minor nit pick, sorry.

gep13 (approved):
LGTM!

PR #3682, opened 2025-06-04
(#3344) Adds ParameterBlock Handling to Chocolatey Script Runner

### LINKED ISSUE #3344: Allow use of param() blocks in Chocolatey* Scripts
### Checklist

As a PowerShell developer, it would be great if we could support the commonly-used param-block usage seen in functions and scripts.

```PowerShell
param(
    # The path to install to
    [string]$InstallDir = "C:\Python$($env:ChocolateyPackageVersion -replace "^(\d+\.\d+).*", "`$1")"
)
$ErrorActionPreference = 'Stop'
$toolsDir   = Split-Path $MyInvocation.MyCommand.Definition -Parent

<# ...Do install things... #>
```

When writing a PowerShell script, I expect this to work - but there is no nice way to pass arguments into Chocolatey run scripts like this. 

Instead, we have folk using something like this:

```PowerShell
$pp = Get-PackageParameters
$twoPartVersion = $Env:ChocolateyPackageVersion -replace "^(\d+\.\d+).*", "`$1"
$defaultFolder = '{0}\Python{1}' -f $Env:SystemDrive, ($twoPartVersion -replace '\.')
if ( $pp.InstallDir ) {
  $installDir = $pp.InstallDir
  if ($installDir.StartsWith("'") -or $installDir.StartsWith('"')) { $installDir = $installDir -replace '^.|.$' }
  mkdir -force $installDir -ea 0 | out-null
}
else {
  $installDir = $defaultFolder
}
```

And there are _many_ different ways people can write this handling logic.

`Get-PackageParameters` exists, but everyone has to reimplement the handling for parameters in their package, which is less simple than it might be. If we had a first-class handling for param-blocks (or some similar modern method), we could advise folk to handle parameters in a specific way, which would allow us to do potentially interesting things.

For instance, by enabling package maintainers to use param blocks:

- They could (possibly) use standard PowerShell parameter validation and other fun attributes
- They could add comment help to the parameters
    - This could then be analyzed and exposed in various ways (e.g. automatically displaying supported parameters and help for them on CCR, or within the CLI).
- It leads into a known method and place to handle variables that can be changed, which will help community moderators and users reading install-scripts understand them more easily
- It reduces the amount of package maintainers having to rewrite the (parameter handling) wheel every package

It could also ease testing of install scripts, though I am not convinced that is particularly useful at this point for various reasons.

### Additional Context

I had a play with adding a very basic transformation attribute to parameters within a `chocolateyInstall.ps1` script.

```PowerShell
class PackageParameterAttribute : System.Management.Automation.ArgumentTransformationAttribute {
    [string]$TargetParameter

    static [string] GetPackageParameterValue ($Name) {
        $PP = Get-PackageParameters
        if ($PP.ContainsKey($Name)) {
            return $PP[$Name]
        } else {
            return $null
        }
    }

    [object] Transform([System.Management.Automation.EngineIntrinsics]$engineIntrinsics, [object]$inputData) {
        if ([PackageParameterAttribute]::GetPackageParameterValue($this.TargetParameter)) {
            return [PackageParameterAttribute]::GetPackageParameterValue($this.TargetParameter)
        } else {
            return $inputData
        }
    }

    # PackageParameterAttribute() {
    #     # Can't figure out a way to get the name of the parameter
    #     $this.TargetParameter = "Test"
    # }

    PackageParameterAttribute([string]$Target) {
        $this.TargetParameter = $Target
    }
}
```

This seems to work pretty nicely when added to the `chocolateyInstaller` helper functions, and handles using the default value in a script if a user doesn't pass a matching `--package-parameter` (whilst overriding it if they do).

As an example, by adding `[PackageParameter("NameOfPackageParameter")]` to a parameter, we can see the default, the help, and easily have a user provide input:

```PowerShell
[CmdletBinding()]
param(
    # The path to extract the files to
    [PackageParameter("InstallPath")]
    $InstallPath = $(Split-Path $MyInvocation.MyCommand.Definition -Parent),

    # A message to output (completely arbitrary example)
    [PackageParameter("Message")]
    [ValidateNotNullOrEmpty()]
    $MessageOutput = "There was no additional message provided."
)

Write-Host "Installing package to '$($InstallPath)'"

Write-Host $MessageOutput
```

image

It has a few potential disadvantages:

- Classes like this are only supported in PowerShell 5+ (but we could use `Add-Type` to bring it back to PowerShell 3+, or possibly use a compiled module to make it available everywhere?)
- Adding parameter attributes like this will cause package scripts to fail if the attribute isn't available, which is obviously rubbish. Adding the attribute _does_ seem to work both from helpers _and_ from extensions, though, so we could have it available in Choco-latest and a compatibility package.
- ArgumentTransformation is only triggered on parameters that have a value, it seems, so you need to provide a default value to any parameters that use it. It's possible we could implement this as a different attribute type and still modify the value?
- It could be inefficient if someone had an obscene number of parameters (though I think the number would have to be silly, and this could be easily solved - again, this is just some POC fun to inspire discussion, here)

Example of Add-Type equivalent:
```
Add-Type @'
using System.Management.Automation;

public sealed class PackageParameterAttribute : ArgumentTransformationAttribute {
    string _targetParameter;

    string _getPackageParameterValueScript {
        get {
            return string.Format(
                @"
                    $PP = Get-PackageParameters
                    if ($PP.ContainsKey('{0}')) {{
                        return $PP['{0}']
                    }} else {{
                        return $null
                    }}
                ",
                _targetParameter
            );
        }
    }

    // // This currently doesn't work, as we can't retrieve the ParameterName easily
    // public PackageParameterAttribute() {
    //   _targetParameter = '???'
    // }

    public PackageParameterAttribute(string packageParameterName) {
        _targetParameter = packageParameterName;
    }

    public override object Transform(EngineIntrinsics engineIntrinsics, object inputData) {
        var result = engineIntrinsics.InvokeCommand.InvokeScript(_getPackageParameterValueScript)[0];
        if (null != result) {
            return result;
        }
        return inputData;
    }
}
'@
```

An alternative to adding an attribute like this would be rewriting the `chocolateyScriptRunner.ps1` to pass in parameters where parameters are found, but that would involve a fair bit of calculation (or a requirement for an ignored parameter with ValueFromRemainingArguments to swallow unwanted splatting on all supporting scripts, perhaps?) and I've quite enjoyed this method so far.

### Related Issues

_No response_

JPRuskin:
## Description Of Changes

This PR introduces the ability for ChocolateyInstall, ChocolateyBeforeModify, and ChocolateyUninstall scripts to have and utilise PowerShell `param()` blocks to handle parameters.

It does this by adding a function (`Get-PackageScriptParameters`), which calculates what parameters a script has and passes any parameters found by the existing function `Get-PackageParameters` that match the parameter names to the script. Optionally, we could handle aliases, but this implementation does not yet.

The new function is called within the ChocolateyScriptRunner before invoking the script, and the resultant hashtable is "splatted" with the call. When no matching parameters are found (on the script _and_ in packageparameters), this results in no change to behaviour.

It's worth noting that though this does allow for usage of parameter attributes and validation, such as `[Parameter(Mandatory)]`, `[ValidateSet()]` and the like, Chocolatey can currently swallow some useful output (e.g. with SomeString being a mandatory parameter):
image
vs the PowerShell default:
image
This could possibly be improved in the future.

It adds the functionality, but doesn't remove the existing parameter handling - so all existing packages will continue working.

### SMA Update
As a part of this, because of the use of the `System.Management.Automation.Language` to assess the param block, the `System.Management.Automation` assembly present in the `lib/powershell` folder was updated from a version matching the version shipped with PowerShell 2.0 (best guess) to a version from our earliest currently supported version of Windows - 2012.

This DLL isn't shipped in any of our builds; it's just used to ensure our code uses methods et al that are supported by the specific DLL we have embedded in the repository.

Specifically, I grabbed this from the uncertain/fun provenance of a Windows Server 2012 VM, in the folder `C:\Windows\Microsoft.NET\assembly\GAC_MSIL\System.Management.Automation\v4.0_3.0.0.0__31bf3856ad364e35`. There's not a similar version on nuget.org, annoyingly, which only has versions from 2018 onwards starting at 6.0.4.

There is an argument that this is a breaking change, as it may no longer work on machines we don't support - but that "break" has already happened. We _do not_ support Windows versions prior to 2012 (and that only in Azure). Everything else is additive-only, and backwards compatible via a "lovely" new extension I'm looking to publish.

## Motivation and Context

This allows for standardisation of how package maintainers handle parameters within their scripts, and helps Chocolatey Community Repository users _and_ moderators to more easily understand what parameters can be used, what default values are used, and what (if any) validation is in place.

## Testing

- Grab https://chocolatey.org/install.ps1 and the modified version of the Chocolatey nupkg package, copy them to your test environment.
- `Set-ExecutionPolicy RemoteSigned Process -Force` and Install with `.\Install.ps1 -ChocolateyDownloadUrl .\Path\To\chocolatey.nupkg`
- Optional: install a random package without parameters to ensure I've not broken the existing functionality `choco install cloudflared --confirm`
- Optional: install a random package with parameters to ensure I've not broken the existing functionality `choco install Temurin21jre --params="/ADDLOCAL=FeatureMain,FeatureEnvironment,FeatureJarFileRunWith,FeatureJavaHome" --confirm`
- Grab a package with param blocks available, e.g. [testextension.1.0.0.zip](https://github.com/user-attachments/files/20596910/testextension.1.0.0.zip) (attached as a zip, as nupkg is not a supported file type) and copy it to the test environment (or have it in a repository)
- `choco install testextension --params='/SomeString:A String Value' --confirm` (some massaging required for source)
    image

> NB: testextension will fail to install if the PowerShell parameter value "SomeString" provided doesn't match the package parameter found by `Get-PackageParameters`. The install script is partially shown below:

```powershell
param(
    [string]$SomeString
)

$PackageParameters = Get-PackageParameters

if ($PackageParameters.SomeString -ne $SomeString) {
    throw "'$SomeString' was not equal to '$($PackageParameters.SomeString)'"
}
```

### Operating Systems Testing

- Windows Server 2012
    image
- Windows Server 2022
    image
- Windows 11 (Sandbox)
    image

I also have an extension package that adds the same functionality to any other extension-supporting version of Chocolatey CLI. This hopefully means that this can be an acceptable thing to use for Community packages.

image

I am very excited about this, despite the code being _utterly despicable_.

I have had to slightly revise my backwards compatibility happiness - it doesn't work with the earliest mention of Chocolatey extensions (0.9.10) because `Get-PackageParameters` doesn't exist at that point. I could fix this by packing a backup version fo the function in the extension, but I feel like supporting back to the 0.10.8 is enough for any unreasonable expectation:
image

## Change Types Made

* ~[ ] Bug fix (non-breaking change).~
* ~[ ] Breaking change (fix or feature that could cause existing functionality to change).~ See SMA update.
* ~[ ] Documentation changes.~

## Change Checklist

* ~[ ] Requires a change to the documentation.~
* ~[ ] Documentation has been updated.~
* [x] PowerShell code changes: PowerShell v3 compatibility checked?

## Related Issue

Resolves #3344


src/Chocolatey.PowerShell/Helpers/PackageParameter.cs
```diff
@@ -0,0 +1,93 @@
...
+                var sensitivePackageParameters = EnvironmentHelper.GetVariable(
+                    cmdlet,
+                    "ChocolateyPackageParametersSensitive",
+                    EnvironmentVariableTarget.Process);
+                if (!string.IsNullOrEmpty(sensitivePackageParameters))
+                {
+                    logParams = false;
+                    paramStrings.Add(sensitivePackageParameters);
+                }
+            }
```
JPRuskin: I suspect this should be removed, here, though I brought it over from the conversion WIP branch (where it may have been intended for CLE?) so left it for discussion.

src/Chocolatey.PowerShell/Commands/GetPackageScriptParametersCommand.cs
```diff
@@ -0,0 +1,69 @@
...
+            WriteVerbose($"Found {scriptParameters.Count()} parameter(s) in '{ScriptPath}'");
+
+            // For each of those in PackageParameters, add it to the splat
+            foreach (var parameter in scriptParameters)
+            {
+                if (packageParameters.ContainsKey(parameter))
+                {
+                    splatHash.Add(parameter, packageParameters[parameter]);
+                }
+            }
```
vexx32, line 65: Is there a specific reason this wasn't pulled into a helper class/method as well?

src/Chocolatey.PowerShell/Helpers/PackageParameter.cs
vexx32: I think it's fine to put here. The environment variable itself is set in CLE, and never in CLI, so this will essentially do nothing in CLI but IMO there is no harm doing it like this. IIRC `Start-ChocolateyProcessAsAdmin` also has a precedent for this kind of thing, exposing a parameter for `-SensitiveStatements` but it's never used in CLI and isn't supported except in CLE.

src/Chocolatey.PowerShell/Helpers/PackageParameter.cs
```diff
@@ -0,0 +1,93 @@
...
+                    }
+                    else
+                    {
+                        value = (object)true;
+                    }
+
+                    if (logParams)
+                    {
+                        cmdlet.WriteDebug($"Adding package param '{name}'='{value}'");
+                    }
```
vexx32: I'm guessing this logic was kind of pre-existing but looking at this I _do_ wonder if we can split things out at some point to log the non-sensitive things and just skip logging sensitive data, maybe.

vexx32 (commented):
I'll have to come back and review the tests added in a bit, but some initial comments. :3

src/Chocolatey.PowerShell/Helpers/PackageParameter.cs
JPRuskin: I was wondering the same! I'd happily modify it to do that. (It would also make my sideline in log analysis potentially easier!)

src/Chocolatey.PowerShell/Commands/GetPackageScriptParametersCommand.cs
JPRuskin: Honestly, I figured that until we ever wanted to use it in a second place (which may end up being CLE) it made as much sense like this. I can do it, though, if preferred.

vexx32: I've tended to just put everything into helpers almost preemptively, but I'm unsure if CLE needs any of this directly, I suppose? (And tbf I've been working through a lot of the more baseline/reused cmdlets so most of them _have_ needed to be put into helpers lol)

I guess if we can't envision a scenario where we might want it called from other cmdlets at present, it's alright to leave here. I might suggest at least just putting the bulk of the logic into a private method on the cmdlet, if so, then; just makes it a smidge easier to pull out into a helper class if we ever need to.

src/Chocolatey.PowerShell/Helpers/PackageParameter.cs
JPRuskin: I'm not sure I love the way I've done this. Phrase of the year for me, really.

JPRuskin:
I believe this is no longer a breaking change (as it doesn't remove or change any PowerShell helpers), makes all changes to helper functionality in C#, and doesn't break any business promises.

I suspect I've missed the boat on getting it merged into 2.5.0, though! Shocking.

PR #3688, opened 2025-06-09
(#3294, #1664) Handle "|" and "&" in environment variable value

### LINKED ISSUE #1664: refreshenv eats percent symbols in environment variable values
### What You Are Seeing?
refreshenv.cmd doesn't correctly handle environment variables with a value that includes a "%" character.

I'm using Windows 10 and choco v0.10.8.

### What is Expected?
I would like the environment variable value after a refreshenv to be the same as if I create a new process.

 - start new cmd.exe
 - ```> setx test test%20value``` (N.B. This creates a persistent environment variable. After testing, you may wish to remove it with the GUI or registry command as described [here](https://stackoverflow.com/questions/13222724/command-line-to-remove-an-environment-variable-from-the-os-level-configuration).)
 - start new cmd.exe 
 - ```> set``` 
 - Observe that the value of variable ```test``` is ```test%20value```
 - ```> refreshenv```
 - ```> set``` 
 - Observe that the value of variable ```test``` is ```test0value```

### Output Log
I note that the ```refreshenv.cmd``` uses a temporary file ```%TEMP%\_env.cmd```
The script does delete some of the files that it creates in this folder, but not ```_env.cmd```, so it is possible to observe the contents of the file after running ```refreshenv```. In my case, the relevant line is 

```set "test=test%20value"```

This seems ok, but the command processor is interpreting the ```%2``` as a command line argument, so the final value of the variable is ```test0value```.

### Thoughts about possible fixes
One possibility would be to replace all occurrences of % in the _env.cmd file with %% before calling the cmd file. It does seem to solve the current problem, but it appears to cause a problem: when I edit my path variable using the windows GUI, I see one entry which is ```%JAVA_HOME%\bin```. This variable is expanded by the system at some point, because when I run ```set``` in my ```cmd.exe``` window, the result does not show ```%JAVA_HOME%\bin```, it shows ```C:\Program Files\Java\jre1.8.0_181\bin```. If I naively replace all single % with %%, then the resulting path does not have that variable replaced. In my case, it might be possible to replace % with %% in all the non-path variables. But it's not clear that would work for other people, who may want variable expansion in non-path variables.

I'm not sure if there is a sensible way to distinguish between cases where the % should be quoted, and cases where it should not.

Another workaround I am considering is to store my environment variable value with %%:
    ```setx test test%%20value``` 
Then, when I look at the value in a fresh cmd.exe process, it will be ```test%%20value```. After ```refreshenv```, the value will be ```test%20value```.
My application that uses the value could then replace occurrences of %% with % before processing the value further.
For my application, where the value I'm storing is an encoded URL, I don't need to represent a real value that could have %% in it literally, so the replacement should be safe. But this doesn't seem safe in general.

Interested in peoples thoughts.

### LINKED ISSUE #3294: RefreshEnv.cmd can't handle `"|` in environment variables
### Checklist

- [X] I confirm I am using **official**, and **not** unofficial, or modified, Chocolatey products.

### What You Are Seeing?

When an environment variable is set that contains double quotes followed by a pipe (`"|`), `RefreshEnv.cmd` will fail. This came about due to a user's `JAVA_OPTS` environment variable looking something like the below:

```JAVA_OPTS = '"-Dhttp.proxyPort=80" "-Dhttps.proxyPort=80" "-Dhttp.nonProxyHosts=localhost|127.*|[::1]"'```

The failure looks like the below:

```
C:\Users\cmacleod17.CORP>C:\Users\cmacleod17.CORP\RefreshEnv.cmd
Refreshing environment variables from registry for cmd.exe. Please wait...'127.*' is not recognized as an internal or external command,
operable program or batch file.
```

### What is Expected?

That `RefreshEnv.cmd` puts `JAVA_OPTS` into the environment without raising an error.

Minimal reproducible case:

1. Set an environment variable to `"|1"`

image

2. Run `RefreshEnv.cmd`
3. Observe the below error:

```
Refreshing environment variables from registry for cmd.exe. Please wait...'1""' is not recognized as an internal or external command,
operable program or batch file.
```

### System Details

* Operating System: Windows 10
* Windows PowerShell version: N/A (run on CMD)
* Chocolatey CLI Version: Tested with https://github.com/chocolatey/choco/blob/develop/src/chocolatey.resources/redirects/RefreshEnv.cmd
* Chocolatey Licensed Extension version: N/A
* Chocolatey License type: N/A
* Terminal/Emulator: Windows Terminal

### Installed Packages

```bash
N/A
```

### Output Log

```bash
N/A
```

### Additional Context

_No response_

gep13:
## Description Of Changes

This commit adds this replacement, in addition to moving the endlocal further down the file, so that the EnableDelayedExpansion is in place when the operation is done.

## Motivation and Context

When an environment variable contains a "|" character, during the process of hydrating the values, the "|" causes an error, as it is attempting to do "something" when executing the "set" command, from the _env.cmd file.  We suspect that the underlying problem is the combination of the " which is added to the environment variable by Windows itself, and usage of the | as a logical OR operation.  To get around this problem, we needed to escape the pipe character using a ^ character.  This is done using a replacement of a single "|" with "^|".

## Testing

1. Apply the changes in this PR to your local `refreshenv.cmd` file
2. Create an environment variable with the value of `|1`
3. Open a Windows Command Prompt
4. Run `set` to verify you can see the created environment variable with the value of `|1`
5. Run `refreshenv`
6. Verify that there is not an error
7. Run `set` to verify you can see the created environment variable with the same value as before
8. Repeat for `&` and `%` instead of `|`

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

- Fixes #3294
- Fixes #1664

### DIFF
--- a/tests/pester-tests/features/RefreshEnvCmd.Tests.ps1
+++ b/tests/pester-tests/features/RefreshEnvCmd.Tests.ps1
@@ -1,0 +1,71 @@
+﻿Describe "Ensuring RefreshEnv.cmd updates environment variables" -Tag EnvironmentVariables, RefreshEnv, Chocolatey {
+    BeforeDiscovery {
+        $testVariables = @(
+            @{
+                Name = "test"
+                Value = "test%20value"
+            }
+            @{
+                Name = "zsdblamp"
+                Value = '"&1'
+            }
+            @{
+                Name = "zsdblpercent"
+                Value = '"%1'
+            }
+            @{
+                Name = "zsdblpipe"
+                Value = '"|1'
+            }
+            @{
+                Name = "zsnoneamp"
+                Value = "&1"
+            }
+            @{
+                Name = "zspercent"
+                Value = "%1"
+            }
+            @{
+                Name = "zsnonepipe"
+                Value = "|1"
+            }
+            @{
+                Name = "zssingleamp"
+                Value = "'&1"
+            }
+            @{
+                Name = "zssinglepercent"
+                Value = "'%1"
+            }
+            @{
+                Name = "zssinglepipe"
+                Value = "'|1"
+            }
+        )
+    }
+
+    BeforeAll {
+        Initialize-ChocolateyTestInstall
+        New-ChocolateyInstallSnapshot
+    }
+
+    AfterAll {
+        Remove-ChocolateyTestInstall
+    }
+
+    Context "When refreshing environment variables with special characters (<Name>=<Value>)" -ForEach $testVariables {
+        BeforeAll {
+            [System.Environment]::SetEnvironmentVariable($Name, $Value, 'User')
+
+            $Output = & "cmd.exe" "/c" "`"$env:ChocolateyInstall\bin\RefreshEnv.cmd`" & set"
+        }
+
+        AfterAll {
+            [System.Environment]::SetEnvironmentVariable($Name, $null, 'User')
+        }
+
+        It "Should have expected value in refreshed variable" {
+            $Output | Should -Contain "$Name=$Value" -Because ($Output -join "`n")
+        }
+    }
+}
--- a/src/chocolatey.resources/redirects/RefreshEnv.cmd
+++ b/src/chocolatey.resources/redirects/RefreshEnv.cmd
@@ -59,7 +59,6 @@ if !IsExternal!==1 (
     exit 1
 )
 
-endlocal
 :: End code from @beatcracker
 @echo off
 ::
@@ -80,7 +79,18 @@ goto main
 :SetFromReg
     "%WinDir%\System32\Reg" QUERY "%~1" /v "%~2" > "%TEMP%\_envset.tmp" 2>NUL
     for /f "usebackq skip=2 tokens=2,*" %%A IN ("%TEMP%\_envset.tmp") do (
-        echo/set "%~3=%%B"
+        set "tempvar=%%B"
+        REM Remove double quotes from temporary value
+        set "__tempvar=!tempvar:"=!"
+
+        REM If the dequoted string differs from the original string, the variable contains double quotes.
+        REM Escape the | and & in the variable to avoid errors.
+        if NOT "!__tempvar!" == "!tempvar!" (
+            set "tempvar=!tempvar:|=^|!"
+            set "tempvar=!tempvar:&=^&!"
+        )
+
+        echo/set "%~3=!tempvar!"
     )
     goto :EOF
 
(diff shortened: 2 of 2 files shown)


corbob:
@gep13 I'm not able to reproduce getting an error with your specific testing steps... But I do get an error with different steps. I also get undesirable results with your testing steps. I'll try to explain:

For ease of testing I created a new file called `newrefreshenv.cmd` in C:\ProgramData\Chocolatey\bin\`

1. Through the Windows GUI set a variable named `quoted` to `"|1"`, and one named `unquoted` to `|1`
2. In `cmd.exe` Run `refreshenv`.
3. Note the error is received.
4. Run `newrefreshenv`.
5. Note no errors.
6. Run `set` and look at your environment variables.
7. Note that `quoted` is correctly set to `"|1"`, but `unquoted` is set to `^|1`.

I think this is just a matter of figuring out if the variable is quoted and only escaping the `|` if it's quoted.

AdmiringWorm:
The fix from PR #1678 has been added to this PR, this for the most part due to it having to be rebased and re-implemented.

AdmiringWorm (dismissed):
LGTM, but should have another approval as I made part of the changes.

corbob (dismissed):

tests/pester-tests/features/RefreshEnvCmd.Tests.ps1
```diff
@@ -0,0 +1,71 @@
...
+
+    AfterAll {
+        Remove-ChocolateyTestInstall
+    }
+
+    Context "When refreshing environment variables with special characters (<Name>=<Value>)" -ForEach $testVariables {
+        BeforeAll {
+            [System.Environment]::SetEnvironmentVariable($Name, $Value, 'User')
+
+            $Output = & "cmd.exe" "/c" "`"E:\orgs\chocolatey\choco\src\chocolatey.resources\redirects\RefreshEnv.cmd`" & set"
```
gep13: Can you remove the hard-coded path here?

tests/pester-tests/features/RefreshEnvCmd.Tests.ps1
```diff
@@ -0,0 +1,71 @@
...
+
+            $Output = & "cmd.exe" "/c" "`"$env:ChocolateyInstall\bin\RefreshEnv.cmd`" & set"
+        }
+
+        AfterAll {
+            [System.Environment]::SetEnvironmentVariable($Name, $null, 'User')
+        }
+
+        It "Should have expected value in refreshed variable" {
+            $Output | Should -Contain "$Name=$($Value)" -Because $Output
```
corbob:
```suggestion
            $Output | Should -Contain "$Name=$($Value)" -Because ($Output -join [System.Environment]::NewLine)
```
This will give us easier to read output if there is an error.

AdmiringWorm: CHanged this to use ``"`n"`` instead of `[System.Environment]::NewLine`

corbob (approved):
LGTM

AdmiringWorm:
The failure on the ubuntu build has to do with the setup, and since the previous run was successful I will get this PR merged now.

PR #3689, opened 2025-06-10
(#1713) Previous package version as env variable

### LINKED ISSUE #1713: Provide version of previously installed package version
This is a feature request.

Currently the old package version is available in `chocolateyBeforeModify.ps1` with environment variable `$env:chocolateyPackageVersion` (because at that time the package has not yet been upgraded or removed). From time to time, we made some mistake in packaging, and would like to release a fix migrating user from bad state to good one. It would be great if we have access to old version in `chocolateyInstall.ps1`, through `$env:chocolateyPackageOldVersion` for example.

gep13:
## Description Of Changes

This commit adds the ability to output the previous package version, during say a `choco upgrade` command, as an environment variable. Prior to this commit, it was only the package version that was currently being installed/upgraded that was presented as an environment variable.

The method that is responsible for writing the environment variables is the PreparePowerShellEnvironment method, which is defined by the IPowershellService interface.  Rather than make a change to the interface to pass in this information, it was decided to add a property to the ChocolateyConfiguration class, which is currently being passed through to this method.  This means that there is some maintainability in place, where additional information can be passed through as environment variables in the future, should the need arise.  This would be done by simply adding another item into the Dictionary.

Pester tests have been updated to allow the testing for this new environment variable during the process of upgrading a package version.

## Motivation and Context

This is needed since there are times that during an upgrade you want to do something based on which version is being upgraded from. For example, you might have to do "something" when upgrading from version 1.0.0 and "something else" when upgrading from version 2.0.0.  Being able to inspect the previous package version as an environment variable, in the same way as you can the package version being currently installed is _very_ useful, and also consistent with how package maintainers get other information.

## Testing

1. Clone this repository, and check out PR branch
2. Open project in Visual Studio, and set a breakpoint on line 1719 of the `NugetService.cs` class
3. Run `install vscode --version 1.100.2 -n`
4. The above breakpoint should not be hit, as the install is a different code path
5. Run `upgrade vscode -n`
6. The above breakpoint should be hit for first the `vscode.install` package, with the previous version of 1.100.2 and the current version 1.100.3
7. Continue code execution
8. The above breakpoint should be hit for first the `vscode` package, with the previous version of 1.100.2 and the current version 1.100.3
9. Each time, the `config.Information.EnvironmentVariables` should either have a Count of 0, or have a null value for the `ChocolateyPreviousPackageVersion` entry.

### Operating Systems Testing

- Windows 11

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

Fixes #1713

### DIFF
--- a/src/chocolatey/StringResources.cs
+++ b/src/chocolatey/StringResources.cs
@@ -175,6 +175,11 @@ namespace chocolatey
                 public const string OsVersion = "OS_VERSION";
 
                 /// <summary>
+                /// The normalized version of the currently installed package version - set during package upgrade.
+                /// </summary>
+                public const string ChocolateyPreviousPackageVersion = nameof(ChocolateyPreviousPackageVersion);
+
+                /// <summary>
                 /// The location where temporary files should be stored during PowerShell execution.
                 /// </summary>
                 /// <remarks>
--- a/src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
+++ b/src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
@@ -530,6 +530,11 @@ NOTE: Hiding sensitive configuration data! Please double and triple
     [Serializable]
     public sealed class InformationCommandConfiguration
     {
+        public InformationCommandConfiguration()
+        {
+            EnvironmentVariables = new Dictionary<string, string>();
+        }
+
         // application set variables
         public PlatformType PlatformType { get; set; }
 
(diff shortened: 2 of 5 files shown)


gep13:
@AdmiringWorm as a reminder here, for both you, and me, we need to update the docs page with the new environment variable that is being created here.

AdmiringWorm (approved):
LGTM, and passed all expected tests internally.

AdmiringWorm:
@gep13 great work on getting this added 👍

PR #3727, opened 2025-06-26
(#3717) Add `--ignore-pinned` option to choco list

### LINKED ISSUE #3717: Add option to `choco list` to ignore packages that are pinned
### Checklist

- [x] I have verified no other issues exist related to my request.

It should be possible to easily retrieve a list of currently installed packages that are not pinned.  In order to get that information today, you would need to run `choco list` and then `choco outdated`, and get the information from the output of the second command, and remove the entries from the first command.

The suggestion instead is doing the following:

```
choco list --ignore-pinned
```

This will then ignore any pinned packages from the list that is provided.

There are times we want to be able to do "stuff" on only the currently installed packages where the package is not pinned.  Some existing Chocolatey CLI already _only_ work on packages that are _not_ currenly pinned, for example `choco upgrade`.  However, there are times when retrieving a list of packages that are not currently installed could be useful.

### Additional Context

There is no requirement for something like:

```
choco list --only-pinned
```

Since this information can be provided via the following command:

```
choco pin list
```

### Related Issues

N/A

gep13:
## Description Of Changes

This commit makes it possible to list all the currently installed packages, but ignore (i.e. skip over listing) of any packages that are currently pinned.

This commit also adds the needed unit tests to verify that the new option appears in the OptionSet for the ChocolateyListCommand, as well as extending the Pester tests to verify that when listing packages, using the --ignore-pinned option means that pinned packages are no longer shown.

## Motivation and Context

This can be useful information to have when deciding what operations can/should be performed on the currently installed packages.  Without this option in place, it would have been necessary to combine the output from choco list and choco pin list, to get this information.  Now, this is much simpler by running: choco list--ignore-pinned.

## Testing

1. Run `choco list` to verify that you have some packages
2. Run `choco pin list` to verify that you have a pinned package (if you don't, run `choco pin` to pin one of the packages in the list from step 1
3. Run `choco list` again, to verify that all packages are still shown
4. Run `choco list --ignore-pinned` to verify that the pinned package is no longer shown

In addition, a Test-Kitchen test will be completed for this PR.

### Operating Systems Testing

- Windows 11

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

* [x] PowerShell code changes: PowerShell v3 compatibility checked?

## Related Issue

Fixes #3717


src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
```diff
@@ -624,6 +624,7 @@ public ListCommandConfiguration()
         public bool ByIdOnly { get; set; }
         public bool ByTagOnly { get; set; }
         public bool IdStartsWith { get; set; }
+        public bool IgnorePinned { get; set; }
```
AdmiringWorm, line 628: We spoke about earlier today that we would try do avoid double negatives for the Configuration file itself.

Since this is a new option, would it maybe make sense to flip this to something like `ListPinnedPackage` and default it to `true` instead?

src/chocolatey/infrastructure.app/commands/ChocolateyListCommand.cs
```diff
@@ -135,7 +135,10 @@ public virtual void ConfigureArgumentParser(OptionSet optionSet, ChocolateyConfi
                      option => configuration.ListCommand.IdStartsWith = option != null)
                  .Add("detail|detailed",
                      "Detailed - Alias for verbose.",
-                     option => configuration.Verbose = option != null);
+                     option => configuration.Verbose = option != null)
+                 .Add("ignore-pinned",
+                     "Ignore Pinned - Ignore pinned packages. Defaults to false.",
+                     option => configuration.ListCommand.IgnorePinned = option != null);
```
AdmiringWorm, line 141: If going with my other comment, this could maybe be changed to:

```suggestion
                     option => configuration.ListCommand.ListPinnedPackages = option is null);

```

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -327,12 +333,19 @@ Package url{6}
...
                     }
                 }
                 else
-                {
-                    this.Log().Debug(() => "{0}{1}".FormatWith(package.Identity.Id, config.ListCommand.IdOnly ? string.Empty : " {0}".FormatWith(package.Identity.Version.ToFullStringChecked())));
+                {   
+                    if (!(packageInfo != null && packageInfo.IsPinned && config.ListCommand.IgnorePinned))
+                    {
+                        this.Log().Debug(() => "{0}{1}".FormatWith(package.Identity.Id, config.ListCommand.IdOnly ? string.Empty : " {0}".FormatWith(package.Identity.Version.ToFullStringChecked())));
+                    }
```
AdmiringWorm: Why do we need this part of the if/else statement?

IF we go into this else statement, it would mean that `LocalOnly` is not requested, and there is no local metadata available.

Do we want to hide these packages even for remote searches?

AdmiringWorm (changes requested):
A few questions, but in general this looks good. Haven't had a chance to do a test run yet though.

src/chocolatey/infrastructure.app/services/NugetService.cs
gep13: Looking at this again, I don't think that this if is needed.  So I will get it removed.

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
gep13: Since we are already using `IgnorePinned` elsewhere, I am going to suggest that we stick with it for now, and look to rename everywhere, under a breaking change release.

src/chocolatey/infrastructure.app/commands/ChocolateyListCommand.cs
gep13: See comment here: https://github.com/chocolatey/choco/pull/3727#discussion_r2176924772

gep13:
@AdmiringWorm I have pushed a change to this PR, and left comments on the other conversations.  Can you have another look at this one?  Thanks!

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
AdmiringWorm: We already discussed this internally, and I agree to keep it so we are consistent.

It may however be a point to change it everywhere in the future.

tests/pester-tests/commands/choco-list.Tests.ps1
```diff
@@ -37,6 +37,26 @@ Describe "choco list" -Tag Chocolatey, ListCommand {
...
+        It "Exits with Success (0)" {
+            $Output.ExitCode | Should -Be 0 -Because $Output.String
+        }
+
+        It "Should contain non-pinned packages" {
+            $Output.Lines | Should -Contain "upgradepackage 1.0.0"
+        }
+
+        It "Should not contain pinned packages" {
+            $Output.Lines | Should -Contain "installpackage 1.0.0"
```
AdmiringWorm: This test is wrong.

The test expects the `installpackage` to be listed, but in this case it shouldn't be listed.

AdmiringWorm (changes requested):
I am failing to see anything in the code that is excluding pinned packages, and while running it locally I am still seeing pinned packages listed even when using `--ignore-pinned`.

Did the logic about removing pinned packages get removed by some reason by a mistake?

gep13:
@AdmiringWorm as discussed, I have added an additional commit to this PR to cover a `maint` commit for updating the sln file.

tests/pester-tests/commands/choco-list.Tests.ps1
gep13: Agreed, updated this during a pairing session.

gep13:
I have also kicked off a new Test-Kitchen build for this PR.

AdmiringWorm (approved):
LGTM, and local testing works as expected. Not merging this yet as it is still running through Test Kitchen.

AdmiringWorm:
@gep13 great work on getting this added 👍

Issue #3718, opened 2025-06-22
Add ability to always ignore the HTTP request cache used by Chocolatey CLI commands


gep13:
### Checklist

- [x] I have verified this is the correct repository for opening this issue.
- [x] I have verified no other issues exist related to my request.

### Is Your Feature Request Related To A Problem? Please describe.

Starting with Chocolatey CLI, a number of HTTP requests are cached, and the results reused in subsequent queries.  It is possible to bypass this cache using the `--ignore-http-cache` option, but you have to explicitly set this option each time, there is no way to set this globally. It should be possible to opt into this globlaly, similar to what can be done with the `allowGlobalConfirmation` feature.

### Describe The Solution. Why is it needed?

A new feature named `alwaysIgnoreHttpCache` should be added, which will do the same thing as the `--ignore-http-cache` option, but it means that it won't need to be provided to each command that is being executed.

### Additional Context

There already exists a common option `--ignore-http-cache` that can be used to bypass the cache, but enabling the `alwaysIgnoreHttpCache` would remove the need to always explicitly opt into this.

A decision has been made to _not_ include the reciprocal option for `--use-http-cache`, as that is deemed as not required.  Again, as per the decision here, if there is enough demand, we can look to include this option, but for now, the workaround would be to disable the `alwaysIgnoreHttpCache` feautre.

### Related Issues

- https://github.com/chocolatey/choco/issues/3193 - this is where the `--ignore-http-cache` option was introduced

gep13:
@mwallner you might be interested in this one 😄

PR #3729, opened 2025-06-30
(#3718) Add ability to always ignore HTTP cache

gep13:
## Description Of Changes

This commit adds the ability to enable a new alwaysIgnoreHttpCahe feature, which does the same thing as using the '--ignore-http-cache' option, just allowing it to _always_ be set that way, rather than having to apply it to each command.

Integration tests have been added to exercise the new feature to always ignore the HTTP Cache.  The assertions in the tests are basically verifying that there are either fewer or more HTTP requests based on whether the feature is enabled or disabled.

## Motivation and Context

This has been a requested feature from the Chocolatey Community for a while now.

## Testing

Integration tests have been added to validate this change.

1. Set a breakpoint on line 37 of the `ChocolateySourceCacheContext.cs` file
2. Run `choco feature enable alwaysIgnoreHttpCache`
3. Run `choco search windirstat`
4. When the breakpoint above is hit, make sure that the if block is entered
5. Run `choco feature disable alwaysIgnoreHttpCache`
6. Run `choco search windirstat`
7. When the breakpoint above is hit, make sure that the if block is _not_ entered

### Operating Systems Testing

* Windows 11

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

#3718


src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
```diff
@@ -463,15 +464,9 @@ private static void SetGlobalOptions(IList<string> args, ChocolateyConfiguration
...
-                            option =>
-                            {
-                                if (option != null)
-                                {
-                                    config.CacheExpirationInMinutes = -1;
-                                }
-                            });
+                        .Add(StringResources.Options.IgnoreHttpCache,
+                            StringResources.OptionDescriptions.IgnoreHttpCache,
+                            option => config.IgnoreHttpCache = true);
```
AdmiringWorm: To be safe, I think you also need to set the `CacheExpirationInMinutes` to a negative value.

Otherwise it may cause problems elsewhere in the codebase (maybe even for the CLE download command).

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
```diff
@@ -352,6 +352,7 @@ private void AppendOutput(StringBuilder propertyValues, string append)
         public bool ApplyPackageParametersToDependencies { get; set; }
         public bool ApplyInstallArgumentsToDependencies { get; set; }
         public bool IgnoreDependencies { get; set; }
+        public bool IgnoreHttpCache { get; set; }
```
AdmiringWorm: This is again a double negative here. Same as #3727, we should probably turn this the other way around.
Ie, `IncludeHttpCache`, or `UseHttpCache`, or maybe `EnableHttpCache`.

src/chocolatey/infrastructure.app/ApplicationParameters.cs
```diff
@@ -276,6 +276,7 @@ public static class Features
             public static readonly string DisableCompatibilityChecks = "disableCompatibilityChecks";
             public static readonly string UsePackageHashValidation = "usePackageHashValidation";
             public static readonly string AlwaysIncludeHeaders = "alwaysIncludeHeaders";
+            public static readonly string AlwaysIgnoreHttpCache = "alwaysIgnoreHttpCache";
```
AdmiringWorm: Not sure if I like this name, but not requesting any changes here.
Just wanted to note it.

src/chocolatey/infrastructure.app/builders/ConfigurationBuilder.cs
AdmiringWorm: I also learned about some hidden options in our parsing command.
Instead of blindly setting this to `true`, check if the `option` is null or not, and only enable it if it is not null.

gep13: I will get this updated.

src/chocolatey/infrastructure.app/configuration/ChocolateyConfiguration.cs
gep13: I have done as you suggested here.

gep13:
@AdmiringWorm this should be ready for another review.  Can you please have a careful review of this to make sure I haven't messed something up with the switch to using a non double negative!  Thanks!

AdmiringWorm:
> @AdmiringWorm this should be ready for another review. Can you please have a careful review of this to make sure I haven't messed something up with the switch to using a non double negative! Thanks!

Oh, you mean I have to run it with limited output only 😛

AdmiringWorm:
Additional test made after the listed tests:

1. Reset the feature `useHttpCache` back to the default (no command for that, so removing the entry in the configuration file).
2. Run `choco search windirstat` and verify `MaxAge` in the `ChocolateySourceCacheContext` is set to the cache expiration in minutes.
3. Open `chocolatey.config` and verify the listed feature `useHttpCache` is set to `true` and not explicitly set.

AdmiringWorm:
@gep13 thanks for getting this feature added. Great work 👍

Issue #3694, opened 2025-06-11
Unable to list template information for the NewFileInstaller template


gep13:
### Checklist

- [x] I confirm there are no unresolved issues reported on the [Chocolatey Status page](https://status.chocolatey.org).
- [x] I have verified this is the correct repository for opening this issue.
- [x] I have verified no other issues exist related to my problem.
- [x] I have verified this is not an issue for a specific package.
- [x] I have verified this issue is not security related.
- [x] I confirm I am using **official**, and **not** unofficial, or modified, Chocolatey products.

### What You Are Seeing?

When attempting to get information about the installed `NewFileInstaller` template, an error output is shown saying that the template cannot be found.

![Image](https://github.com/user-attachments/assets/beeebbfe-75d9-41ed-aa8b-dbcfe4f633ff)

If I change the name of the template folder from `NewFileInstaller` to `newfileinstaller` it is found correctly:

![Image](https://github.com/user-attachments/assets/523d77c2-ad83-4200-b1e6-2e2e2e6772df)

### What is Expected?

The installed template should be found, without the need to change the casing of the name of the folder

### How Did You Get This To Happen?

1. With CLE installed, run the following command `choco new --file <path-to-any-exe>`
2. This will generate a new `NewFileInstaller` folder in the `c:/programdata/chocolatey/templates` folder
3. Run `choco template list`
4. You will see the `NewFileInstaller` template listed as being available
5. Run `choco template info --name NewFileInstaller`
6. You will get an error stating `Unable to find requested template 'newfileinstaller'`

### System Details

* Operating System: Windows 11
* Windows PowerShell version: 7.5.1
* Chocolatey CLI Version: 2.4.3
* Chocolatey Licensed Extension version: 6.3.1
* Chocolatey License type: Business
* Terminal/Emulator: Windows Terminal

### Installed Packages

```bash
N/A
```

### Output Log

```bash
2025-06-11 10:46:49,961 24092 [INFO ] - ============================================================
2025-06-11 10:46:50,014 24092 [DEBUG] - User may be subject to UAC, checking for a split token (not 100%
 effective).
2025-06-11 10:46:50,209 24092 [INFO ] - Chocolatey v2.4.3 Business
2025-06-11 10:46:50,212 24092 [DEBUG] - Chocolatey is running on Windows v 10.0.26100.0
2025-06-11 10:46:50,217 24092 [DEBUG] - Attempting to delete file "C:/ProgramData/chocolatey/choco.exe.old".
[...]

PR #3736, opened 2025-07-03
(#3694) Fix edge case with template naming 

gep13:
## Description Of Changes

A change was made to try to make sure that everything related to
template, the name, the directories, etc are all lower case.  That way,
in the absence of a StringComparer, a simple comparison can be made.
However, one of our automated tests caught a problem with this approach
when the code is checking for the defaultTemplate and whether it had
been overridden or not.  This code path hadn't been ToLower'ed and as a
result, when there are uppercase letters in the path, for example the
"C:\ProgramData" location, the check would fail, and it would no longer
output that the default template had been overridden.

This commit addresses that problem by first ToLower'ing these values
ahead of the check.  It also switches to using the case-insensitive
IsEqualTo method, which makes the code slightly less complicated, and
easier to follow.

## Motivation and Context

Some Test-Kitchen tests were failing, which made us investigate this problem.

## Testing

Tested locally, but will also re-run Test-Kitchen tests from this PR.

### Operating Systems Testing

- Windows 11

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

Relates to #3694


AdmiringWorm (approved):
LGTM, but are waiting on internal Test Kitchen to finish before this PR can be merged.

gep13:
With @AdmiringWorm approval in place, and the Test-Kitchen tests completed, I am going to go ahead and get this merged.

PR #3775, opened 2025-08-23
(#1276) List command with last updated date

### LINKED ISSUE #1276: Improving choco list with options to show dependencies and sort by date (and more)
**Feature request:**
* Add features to `choco list`:
    * Show the installed apps with their dependencies (maybe as a tree?)
    * Sort the output list by installed date, last updated
    * Check which arguments were used to install an app
    * Show exact installed apps - only apps I actively installed (without dependencies). Useful for saving the list of my installed apps.

I believe it would be very useful 😄

swaran18999:
## Description Of Changes
- Add --show-last-updated-date option to choco list to display the last date a package was installed/updated, formatted as yyyy-MM-dd HH:mm:ss. Falls back to “Last updated not available” when metadata is missing.
- Add --order-by-last-updated-date option to choco list to order results by the package’s last installed/updated date by mapping to PackageOrder.LastPublished.
- Update PowerShell tab completion (src/chocolatey.resources/helpers/ChocolateyTabExpansion.ps1) to include both new switches in the list command’s completion options.

## Motivation and Context
- Administrators and users often need quick visibility into when packages were last installed or updated, and to sort lists by recency for auditing and maintenance workflows. Previously, choco list did not expose this information nor allow sorting by it.
- These additions improve usability by displaying last-change timestamps inline with list results.
- This work addresses the needs discussed in issue #1276

## Testing
- Flag parsing for --show-last-updated-date sets ShowLastUpdatedDate = true; defaults to false when omitted. Output includes formatted date yyyy-MM-dd HH:mm:ss when Published exists.
- Sorting places packages in the expected order by last updated date. 
- Correct order when --order-by-last-updated-date is used.
- Behavior is consistent with --id-only/--by-id-only (date not appended, sorting still applied).

### Operating Systems Testing
- Windows 10

## Change Types Made
* [ ] Documentation changes.

## Change Checklist

## Related Issue
Partially fixes #1276


src/chocolatey.resources/helpers/ChocolateyTabExpansion.ps1
```diff
@@ -49,7 +49,7 @@ $commandOptions = @{
...
     license   = ""
-    list      = "--by-id-only --by-tag-only --detail --exact --id-only --id-starts-with --ignore-pinned --include-programs --page='' --page-size='' --prerelease --source='' --version=''"
+    list      = "--by-id-only --by-tag-only --detail --exact --id-only --id-starts-with --ignore-pinned --include-programs --page='' --page-size='' --prerelease --order-by-last-updated-date --show-last-updated-date --source='' --version=''"
```
vexx32: I think most of our flags are given in alphabetical order in tab completion; could we put these in before `--page`?

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -276,11 +276,12 @@ it is possible that incomplete package lists are returned from a command
                     {
                         if (!(packageInfo != null && packageInfo.IsPinned && config.ListCommand.IgnorePinned))
                         {
-                            this.Log().Info(logger, () => "{0}{1}".FormatWith(package.Identity.Id, config.ListCommand.IdOnly ? string.Empty : " {0}{1}{2}{3}".FormatWith(
+                            this.Log().Info(logger, () => "{0}{1}".FormatWith(package.Identity.Id, config.ListCommand.IdOnly ? string.Empty : " {0}{1}{2}{3} {4}".FormatWith(
```
vexx32: Looking at how the other information is added here, it looks like the space before each arg is added below rather than here; I'm guessing that's probably due to not wanting to add trailing space and also to keep it simple to extend later on. For consistency's sake, can we do the same here? i.e., no space in this format string, but adding the space if the user is asking to show the last updated date -- similar to how the Approved / download cache information is done.

I *think* you can do that just by adding a space to the ToString() format string on line 284 and the "Last updated not available" string..

vexx32 (changes requested):
Looks pretty solid, nice self-contained change, just a couple minor comments.

You will need to rebase this branch against `develop` when you're done, just so the merge history is a bit tidier.

src/chocolatey.resources/helpers/ChocolateyTabExpansion.ps1
swaran18999: change has been made

src/chocolatey/infrastructure.app/services/NugetService.cs
swaran18999: change has been made

vexx32:
Alright, I've given it a run through and I do have some comments/concerns with this current implementation. It looks like the `Published` date is not the correct thing to use here, primarily because it only looks at the file's LastModifiedTime as a UTC time. This isn't a sufficiently reliable metric for this purpose; there are cases where `choco` may modify the nupkg file, so this date isn't guaranteed to be the same as the date the package was actually installed or updated.

What we're going to need to do is a bit more involved if we want to do this. Probably the simplest approach is going to be taking notes from how the `.deploymentLocation` file works.

First, we have a static property that just stores the filename as a constant. You'll need to add another, call it `.lastUpdated`

https://github.com/chocolatey/choco/blob/97bc0d4b06864ec17788c892c7cbbdbf1f0befeb/src/chocolatey/infrastructure.app/services/ChocolateyPackageInformationService.cs#L47

Second, it will need to have read/write logic in this file to save and load this data (the install/upgrade/etc methods in NugetService call into here, so you can just add logic into here in order to save/load this data), similar to this and around the same areas of the file:

https://github.com/chocolatey/choco/blob/97bc0d4b06864ec17788c892c7cbbdbf1f0befeb/src/chocolatey/infrastructure.app/services/ChocolateyPackageInformationService.cs#L184-L196

https://github.com/chocolatey/choco/blob/97bc0d4b06864ec17788c892c7cbbdbf1f0befeb/src/chocolatey/infrastructure.app/services/ChocolateyPackageInformationService.cs#L299-L308

When saving this information I think it would make the most sense to grab the value of `DateTime.Now` and just store that.

And then, once you have those two pieces in place, the current place you've added the logic to display it into will need to be amended. Specifically you'll need to copy what's done for deploymentLocation here and do similar to retrieve the last updated date:

https://github.com/chocolatey/choco/blob/97bc0d4b06864ec17788c892c7cbbdbf1f0befeb/src/chocolatey/infrastructure.app/services/NugetService.cs#L220C24-L241

This can then be displayed in the list output if needed.

swaran18999:
@vexx32 Can you please review this? The requested changes have been made and tested as well.

src/chocolatey/infrastructure.app/services/ChocolateyPackageInformationService.cs
```diff
@@ -353,7 +385,7 @@ public void Remove(IPackageMetadata package)
             _fileSystem.DeleteDirectoryChecked(pkgStorePath, recursive: true);
         }
 
-        private static string GetStorePath(IFileSystem fileSystem, string id, NuGetVersion version)
+        public static string GetStorePath(IFileSystem fileSystem, string id, NuGetVersion version)
```
vexx32: I haven't found where this is used yet, but as a general rule we are *very* cautious making any APIs `public`. If this can be `internal` or something else non-public, that would be preferred. Given it's a static method, it also might make more sense to move it to a more easily-accessible place if we're needing to use it in multiple places.

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -240,7 +241,7 @@ it is possible that incomplete package lists are returned from a command
                     packageInfo = _packageInfoService.Get(packageLocalMetadata);
                     deploymentlocation = packageInfo.DeploymentLocation;
                     sourceInstalledFrom = packageInfo.SourceInstalledFrom;
-
+                    lastUpdated = packageInfo.LastUpdated;
                     if (config.ListCommand.IncludeVersionOverrides)
```
vexx32, line 246:
```suggestion
                    lastUpdated = packageInfo.LastUpdated;

                    if (config.ListCommand.IncludeVersionOverrides)
```

vexx32 (changes requested):
Just about there in terms of the overall code, although we probably want to avoid making that method `public` if we can.

We'll also need to look at what tests can be added here to validate the behaviour automatically. If you'd like some pointers on that, there are some existing tests in `tests/chocolatey-tests/choco-list.Tests.ps1` that may be useful to reference, and we can discuss further as needed as well.

PR #3783, opened 2025-09-04
(maint) Finish the 2.5.1 release by merging master to develop

gep13:
This back-merges the changes made to master during the release into develop.

There was a merge conflict when doing this, so it was necessary to create an intermediary branch from master to first merge develop in (resolving conflicts), and then merge that branch into develop via this PR.


gep13:
No, I think that is worth calling out.  Those were files where there was a merge conflict that had to be addressed, so I can only assume that those line endings were changed during that.  I have added another commit to put the line endings back to how they were.

Can you re-review?  Thanks!

AdmiringWorm (changes requested):
In general the changes looks good, however as @st3phhays had noted internally, the credits.pdf file is currently invalid and won't open.

This needs to be re-generated, or corrected. Leaving a request for a change for that until it has been fixed.

PR #3801, opened 2025-10-14
(#3804, #1901) Add source locations to package information

### LINKED ISSUE #1901: Include software install path in `choco info` output
Following [this tweet](https://twitter.com/fsfarimani/status/1163768514739625985?s=20), it would be great if the installation folder could be inquired somehow. For example, the command `choco info packageName` could also report the path.

### LINKED ISSUE #3804: Record and display the source location used to install package
I don't think the platform already does this, but it would be nice to see the source location of the .nupkg file.  Whether it is URL, UNC or local disk.
I think it would be best to leave off the nupkg name so the path could be used directly or if someone want to programmatically reconstruct the full path they could do so on their own.

This would _optionally_ allow for optional chocolatey platform features like checking that source before repo urls.  Not saying this a good idea, just saying things like that would be possible.

Would also allow auditing whether a package has been installed from a proper source or just finding the original .nupkg to see if it is still there or the right version.

* [GitLab issue](https://gitlab.com/chocolatey/collaborators/choco-licensed/-/issues/755)

corbob:
## Description Of Changes

Add the source repository and deployment location to the stored Chocolatey package information, displaying it as part of `choco info --local-only` and `choco list --verbose`.

## Motivation and Context

See #3804 and #1901 

## Testing

1. Run `choco install <packageName>`
2. Check `$env:ChocolateyInstall/.chocolatey/<packageName>.<packageVersion>/.sourceInstalledFrom` that it exists, and includes the repository that was used for installation.
3. Create a local Chocolatey package.
4. Run `choco install <packageName> --source .`
5. Check `$env:ChocolateyInstall/.chocolatey/<packageName>.<packageVersion>/.sourceInstalledFrom` that it exists, and includes the local path that was used for installation.
6. Run `choco info <packageName> --local-only` and ensure that both the deployment location and the source installed from location are listed.
7. Run `choco list --verbose` and ensure that both the deployment location and the source installed from location are listed.

### Operating Systems Testing

- Windows Server 2016
- Windows Server 2019
- Windows 11

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

## Related Issue

- Fixes #3804
- Fixes #1901

### DIFF
--- a/tests/helpers/common/Chocolatey/Initialize-ChocolateyTestInstall.ps1
+++ b/tests/helpers/common/Chocolatey/Initialize-ChocolateyTestInstall.ps1
@@ -28,6 +28,9 @@
         $null = robocopy $env:ChocolateyInstall/lib/chocolatey $Directory/lib/chocolatey /MIR
         $null = robocopy $env:ChocolateyInstall/lib/chocolatey.extension $Directory/lib/chocolatey.extension /MIR
         $null = robocopy $env:ChocolateyInstall/lib/chocolatey-agent $Directory/lib/chocolatey-agent /MIR
+        Get-ChildItem $env:ChocolateyInstall/lib -Filter chocolatey-license-* | ForEach-Object {
+            $null = robocopy "$($_.FullName)" "$Directory/lib/$($_.Name)" /MIR
+        }
 
         $env:ChocolateyInstall = $Directory
         Set-ChocolateyTestLocation -Directory $Directory


src/chocolatey/infrastructure.app/domain/ChocolateyPackageInformation.cs
```diff
@@ -37,5 +37,6 @@ public ChocolateyPackageInformation(IPackageMetadata package)
         public bool IsPinned { get; set; }
         public string ExtraInformation { get; set; }
         public string DeploymentLocation { get; set; }
+        public string PackageInstalledFrom {  get; set; }
```
corbob: Naming things is hard... Should this perhaps be `SourceInstalledFrom`? We can't necessarily use Uri as it may not be a Uri? Perhaps `PackageOriginationSource`?

src/chocolatey/infrastructure.app/services/ChocolateyPackageInformationService.cs
```diff
@@ -45,6 +45,7 @@ public class ChocolateyPackageInformationService : IChocolateyPackageInformation
         private const string ExtraFile = ".extra";
         private const string VersionOverrideFile = ".version";
         private const string DeploymentLocationFile = ".deploymentLocation";
+        private const string PackageInstalledFromFile = ".packageInstalledFrom";
```
corbob: See previous comment regarding naming things.

src/chocolatey/infrastructure/results/PackageResult.cs
```diff
@@ -103,7 +103,12 @@ public PackageResult(IPackageSearchMetadata packageSearch, string installLocatio
             */
         }
 
-        public PackageResult(IPackageMetadata packageMetadata, IPackageSearchMetadata packageSearch, string installLocation, string source = null) : this(packageMetadata.Id, packageMetadata.Version.ToNormalizedStringChecked(), installLocation)
+        [Obsolete("This overload is deprecated and will be removed in v3.")]
+        public PackageResult(IPackageMetadata packageMetadata, IPackageSearchMetadata packageSearch, string installLocation, string source = null) : this(packageMetadata, packageSearch, installLocation, source, null)
+        {
+        }
+
+        public PackageResult(IPackageMetadata packageMetadata, IPackageSearchMetadata packageSearch, string installLocation, string source, string sourceUri) : this(packageMetadata.Id, packageMetadata.Version.ToNormalizedStringChecked(), installLocation)
```
corbob: @gep13 we had discussed making these optional, I found when I did that, I could no longer build CLE with these artifacts as it couldn't determine which method signature to use. Making them both mandatory solved that compile time issue for me. More than happy to be shown a better way to obsolete things here.

src/chocolatey/infrastructure.app/domain/ChocolateyPackageInformation.cs
gep13: @pauby can you provide any input here?  This is with regard to capturing which specific source a package was downloaded and installed from.  Thanks!

src/chocolatey/infrastructure/results/PackageResult.cs
gep13: To keep things working with an optional parameter, I think you would need to do the following:

```
public PackageResult(IPackageMetadata packageMetadata, IPackageSearchMetadata packageSearch, string installLocation, string sourceUri, string source = null)
```

But since we are creating a new constructor, I don't see any harm in making them non optional.

@AdmiringWorm might have some thoughts on this as well, so happy to let him chime in here.

src/chocolatey/infrastructure/results/PackageResult.cs
```diff
@@ -123,6 +128,7 @@ public PackageResult(IPackageMetadata packageMetadata, IPackageSearchMetadata pa
             }
 
             Source = sources.FirstOrDefault(uri => uri.IsFile || uri.IsUnc).ToStringSafe();
+            SourceUri = sourceUri;
```
gep13: Depending on the naming that is decided for the above parameter, we might want to make this one more explicit as well.

src/chocolatey/infrastructure/results/PackageResult.cs
gep13: It would be could to add some documentation for this method as well, especially around why there are two parameters related to Source, i.e. one to capture _all_ the sources used during the operation, and one for the source that was _actually_ used to perform the operation.

AdmiringWorm: Using an overload would be better than using an optional parameter.

Adding more optional parameters would require a new compilation of CLE for it to work against it.
An optional parameter is only optional during compilation, but during runtime it is still a required to be passed, so when CLE is compiled against this constructor with the optional parameter, it will pass along `null` as the value for the `source` parameter, but as this wouldn't be present without a recompile, then the method signature will be different than what CLE expects.

At least, those are my understandings.

src/chocolatey/infrastructure.app/domain/ChocolateyPackageInformation.cs
pauby: `SourceInstalledFrom` is clear what it contains, so I'd suggest that.

src/chocolatey/infrastructure/results/PackageResult.cs
corbob: Are you suggesting that we add a new public property called `SourceInstalledFrom` and use it instead of the existing `SourceUri`?

(for now I will proceed with that being the idea, we can adjust it easily enough if not...)

src/chocolatey/infrastructure/results/PackageResult.cs
corbob: > It would be could to add some documentation for this method as well, especially around why there are two parameters related to Source, i.e. one to capture _all_ the sources used during the operation, and one for the source that was _actually_ used to perform the operation.

I'm not 100% sure on this one, but I have added some details to this particular overload.

vexx32:
From a functionality standpoint I'm happy to approve this. I'll leave the final approval on naming and such to Gary and Paul before we call it all good, though. :3

corbob:
I noticed some missed renames after @vexx32 left their review. I've added one last commit, and run the Chocolatey Agent Test Kitchen suite here: https://sra-soteria-web.ch0.co/buildConfiguration/TestKitchen_ChocolateyAgent/38694. I will review the results in the morning.

src/chocolatey/infrastructure/results/PackageResult.cs
gep13: Hmm, that wasn't what I was suggesting no, as I hadn't noticed (or rather, I had forgotten that `SourceUri` was already a property on this class.  I think what you have done here is fine, I would go one step further though and say that if `SourceUri` is not being used anywhere, we mark it as `Obsolete`, so that going forward, there is no ambiguity about what is going on with this class.  `Source` will be used (when needed) to store all the sources that were _available_ when the operation took place, and `SourceInstalledFrom` used to store where the package was _actually_ installed from.

Thoughts?

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -217,6 +217,9 @@ it is possible that incomplete package lists are returned from a command
 
                 ChocolateyPackageMetadata packageLocalMetadata;
                 string packageInstallLocation = null;
+                string deploymentlocation = null;
+                string packageInstalledFrom = null;
```
gep13: Can we get this updated to the name that we went for?

src/chocolatey/infrastructure.app/services/NugetService.cs
```diff
@@ -329,6 +334,8 @@ Package url{6}
                                     package.Summary != null && !string.IsNullOrWhiteSpace(package.Summary.ToStringSafe()) ? "\r\n Summary: {0}".FormatWith(package.Summary.EscapeCurlyBraces().ToStringSafe()) : string.Empty,
                                     package.Description.EscapeCurlyBraces().Replace("\n    ", "\n").Replace("\n", "\n  "),
                                     !string.IsNullOrWhiteSpace(package.ReleaseNotes.ToStringSafe()) ? "{0} Release Notes: {1}".FormatWith(Environment.NewLine, package.ReleaseNotes.EscapeCurlyBraces().Replace("\n    ", "\n").Replace("\n", "\n  ")) : string.Empty,
+                                    !string.IsNullOrWhiteSpace(deploymentlocation) ? "{0} Deployed to: '{1}'".FormatWith(Environment.NewLine, deploymentlocation) :string.Empty,
+                                    !string.IsNullOrWhiteSpace(packageInstalledFrom) ? "{0} Package Installed From: '{1}'".FormatWith(Environment.NewLine, packageInstalledFrom) : string.Empty,
```
gep13:
```suggestion
                                    !string.IsNullOrWhiteSpace(packageInstalledFrom) ? "{0} Package installed from: '{1}'".FormatWith(Environment.NewLine, packageInstalledFrom) : string.Empty,
```

I don't "know" is this is right or not.  Perhaps a final check from @pauby here?

tests/pester-tests/commands/choco-info.Tests.ps1
```diff
@@ -34,6 +34,27 @@
...
+            $Setup.ExitCode | Should -Be 0 -Because $Setup.String
+
+            $Output = Invoke-Choco info installpackage --local-only
+        }
+
+        It "Exits with Success (0)" {
+            $Output.ExitCode | Should -Be 0 -Because $Output.String
+        }
+
+        It "Should contain a summary" {
```
gep13: Is this a copy/pasta or intended?

gep13 (changes requested):
This all looks good to me, just a couple minor nitpicks, and an overall suggestion on how to finish this off.

src/chocolatey/infrastructure/results/PackageResult.cs
```diff
@@ -103,7 +104,20 @@ public PackageResult(IPackageSearchMetadata packageSearch, string installLocatio
...
+        public PackageResult(IPackageMetadata packageMetadata, IPackageSearchMetadata packageSearch, string installLocation, string source = null) : this(packageMetadata, packageSearch, installLocation, source, null)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PackageResult"/> class.
+        /// </summary>
+        /// <param name="packageMetadata"></param>
+        /// <param name="packageSearch"></param>
+        /// <param name="installLocation"></param>
```
gep13: I think we should fill these out while we are here as well.

gep13 (dismissed):

tests/pester-tests/commands/choco-info.Tests.ps1
corbob: It is a copy/pasta... But it could probably be clearer about what summary it's looking for...

src/chocolatey/infrastructure/results/PackageResult.cs
corbob: I've added help to all of the constructors and properties.

src/chocolatey/infrastructure.app/services/NugetService.cs
corbob: Spoke with Gary about this, he and I were looking at different parts of this suggestion. We've settled on the wording "Source package was installed from" as it matches the name of the variable and is more clear what it's referring to.

src/chocolatey/infrastructure/results/PackageResult.cs
corbob: These have been filled out.

vexx32 (approved):
This looks good to me. Great work, Cory! 💜

PR #3917, opened 2026-06-28
[DO NOT MERGE] Pester 6.0.0 release candidate update

nohwnd:
Hi — Pester 6.0.0 is almost ready, and before I release it I'm trying it out
against real-world Pester v5 test suites.

This PR updates Pester to the 6.0.0 release candidate and makes the changes
needed for your tests to run on it. I opened it to learn two things:

- whether migrating from v5 to v6 is straightforward, and
- where v6 breaks or behaves differently, so I can fix those things before the
  final release instead of after.

You don't need to merge this, and I'd suggest you don't — it's only meant to
show what a v6 update would involve and to give me feedback. Please leave it
open, though: if more release candidates come out I'll push those updates here
too, and I'll close it myself afterwards.

This PR was co-authored by GitHub Copilot. If you'd rather not receive
AI-assisted PRs, just close it and I won't open one for the next release.

Thanks for the CI time this run used, and for maintaining a suite I could test
against.

— Jakub & Copilot

### DIFF
--- a/Invoke-AuthedTests.ps1
+++ b/Invoke-AuthedTests.ps1
@@ -1,4 +1,4 @@
-﻿#Requires -Module @{ ModuleName = 'pester'; ModuleVersion = '5.3.1' }
+﻿#Requires -Module @{ ModuleName = 'pester'; ModuleVersion = '6.0.0' }
 #Requires -RunAsAdministrator
 <#
     .SYNOPSIS
--- a/Invoke-Tests.ps1
+++ b/Invoke-Tests.ps1
@@ -1,4 +1,4 @@
-#Requires -Module @{ ModuleName = 'pester'; ModuleVersion = '5.3.1' }
+#Requires -Module @{ ModuleName = 'pester'; ModuleVersion = '6.0.0' }
 #Requires -RunAsAdministrator
 <#
     .SYNOPSIS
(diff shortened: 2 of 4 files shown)


corbob:
@nohwnd: I've approved the GitHub Actions to run, although I'm not sure that they will provide you with much in the way of confirming changes you're making here.

Because the tests here are testing Chocolatey CLI, I would generally recommend testing with the provided Vagrant environment. Let me know if you have any questions or if there's anything I can help you with.

nohwnd:
Thanks, this helps as well. Version changed to latest, tests still pass.

corbob:
> Thanks, this helps as well. Version changed to latest, tests still pass.

Just to make sure we're both on the same page, when you say tests still pass, do you mean in the GitHub Actions? Or when running things locally? I am as sure as I can be that some of the tests don't currently pass when run locally, but I would call it a success if the numbers of tests run and passing and failing and such match before and after the change.

nohwnd:
I was talking about the status check on this PR.

corbob:
Ah yeah, unfortunately those checks don't actually run any of the Pester tests. If I recall, when I was working on https://github.com/pester/pester/pull/2607, I pulled my changes into our Test Kitchen environment that does run the Pester tests, there was a failure due to an empty ForEach object, but the rest of them passed. I'll see about running the vagrant environment here to get you information about the tests, but it might not be until next week.

nohwnd:
Okay got it, my plan is to release on Tuesday, other repos that ran tests show that we are in good shape. If there are any problems you encounter later when you decide to migrate to v6 please report them back to Pester.

PR #3936, opened 2026-09-17
(#3645) Improve HttpCache validation

### LINKED ISSUE #3645: Chocolatey CLI is incorrectly reporting bad permissions on the HttpCache folder and is not setting them correctly
### Checklist

- [x] I confirm there are no unresolved issues reported on the [Chocolatey Status page](https://status.chocolatey.org).
- [x] I have verified no other issues exist related to my problem.
- [x] I have verified this is not an issue for a specific package.
- [x] I have verified this issue is not security related.
- [x] I confirm I am using **official**, and **not** unofficial, or modified, Chocolatey products.

### What You Are Seeing?

When the permissions on the `httpcache` folder are changed, Chocolatey CLI shows the following validation warning: 

```
System Cache directory is not locked down to administrators. Remove the directory 'C:\ProgramData\ChocolateyHttpCache' to have Chocolatey CLI create it with the proper permissions. 
```

When trying to resolve this warning, the user ran an Advanced Deployment Step in Chocolatey Central Management to remove the `httpcache` folder and have Chocolatey CLI recreate it. Chocolatey CLI creates the cache folder, with all local machine users having write access to the folder. This permission change causes the recreated `httpcache` folder to report the same issue of the folder not being locked down to just Administrator user access.

### What is Expected?

Chocolatey should be able to recreate this folder with the correct permissions. Chocolatey should be setting the cache folder's ACL before every run of Chocolatey.

1. Install Chocolatey for Business and license it. Note the current permissions on `C:\ProgamData\ChocolateyHttpCache` folder.
2. Go to the `C:\ProgamData\ChocolateyHttpCache` folder and edit the permissions to add `LocalMachine\Users` object to include Write access.
3. Run `choco list` and the validation warning mentioned above, should be shown.
4. Undo the folder permissions change done in Step 2.
5. Run the following [exported Chocolatey Central Management deployment](https://gist.github.com/ryanrichter94/19f56ff66d106144ed76c9b5c69f3dd2) against your test machine.
6. You should notice that the `C:\ProgamData\ChocolateyHttpCache` folder has been re-created. However, `Users` now have `Write` access, and the `CREATOR OWNER` has `Full Control` rights. If you run `choco list` again, you get the same validation warning message.

### System Details

- Operating System: 20348.3091 (Windows Server 2022)
- Windows PowerShell Version: 5.1.20348.2849
- Chocolatey CLI Version: 2.4.2
- Chocolatey Licensed Extension version: 6.3.1
- Chocolatey Agent version: 2.2.2
- Chocolatey License type (Professional / Business / ?): Business
- Terminal/Emulator: Windows PowerShell & Chocolatey Central Management

### Installed Packages

```bash
N/A
```

### Output Log

```bash
[chocolatey.log](https://gist.github.com/ryanrichter94/1cf7cba454db54ec7d48664842e65012)
```

### Additional Context

- [GitLab issue](https://gitlab.com/chocolatey/solutions/support-issues/-/issues/122).
- [Zendesk Ticket](https://chocolatey.zendesk.com/agent/tickets/25162).

Adding the following to the PowerShell of the Advanced Deployment Step in the reproduction steps of this issue will programmatically recreate the cache folder, mirroring the ACLs from `$env:ChocolateyInstall` onto it.

```
if (Test-Path -Path "C:\ProgramData\ChocolateyHttpCache"){
	Write-Error "Chocolatey Cache directory still exists!"
} else { 
    Write-Output "Recreating Chocolatey Cache Directory"
    New-Item -Path "C:\ProgramData\" -Name "ChocolateyHttpCache" -ItemType "directory"

    Write-Output "Setting correct permissions for Chocolatey Cache Directory"
    $ChocolateyACL = Get-Acl -Path $env:ChocolateyInstall
    Set-Acl -Path "C:\ProgramData\ChocolateyHttpCache" -AclObject $ChocolateyAcl
}
```

┆Issue is synchronized with this [Clickup task](https://app.clickup.com/t/868m3p4kg) by [Unito](https://www.unito.io)

vexx32:
## Description Of Changes
- Improve validation rule for HttpCache, allowing choco to fail out if something is wrong and it needs to use the HttpCache
- Ensure validation checks are being performed by the public API entry points and not just the CLI entry point
- Add unit tests 

## Motivation and Context

The cache folder validation was a bit fallible and could run into weird issues if we don't handle it correctly. Additionally, in some scenarios using the API, the cache directory could be created with incorrect permissions.

## Testing

1. Run VS as admin
2. Delete your `C:\ProgramData\ChocolateyHttpCache` folder
3. Create the folder again yourself
4. Run CLI from VS with `search vlc`
5. Observe you get a validation warning stating the cache directory had to be purged, but things proceed as normal
6. Check the cache folder permissions and validate that it has been locked down to only allow writing from SYSTEM or Administrators (Users get read access too)
7. Repeat steps 2-4, this time additionally creating a file after step 3 and setting it to Read Only
8. You should get a validation *error* stating that the cache directory could not be locked down or deleted, and Chocolatey fails the operation completely
9. Go back and unset the read-only on the file and rerun choco to fix your local cache directory.

As an additional verification, I would also suggest building CLE and Agent against this PR and installing the resulting packages on a VM to test the background service behaves as expected. On my own VM (*cough* reusing a local Test Kitchen VM) I had to set the following in order to get it to run a `download` in background mode:

```
choco feature enable useBackgroundService --allow-unofficial
choco feature disable useBackgroundServiceWithNonAdministratorsOnly
choco feature disable useBackgroundServiceWithSelfServiceSourcesOnly
choco config set --name backgroundServiceAllowedCommands --value "install,upgrade,uninstall,download"
```

The added Pester tests can be run through Test Kitchen as well, though you will need to target the `HttpCache` tag at least.

I have also pulled the build through into CLE and GUI to validate that the API changes have not immediately broken anything unexpectedly there as well.

In GUI, this does produce an unhandled exception if the cache directory is in an unrecoverable state (weird permissions and readonly files is generally the way that happens). This could probably be improved in GUI in a follow up issue, but I think given this prevents weird behaviour if corrupt cache files get put in there, it may be the lesser of two evils.

### Operating Systems Testing

Win11, Server 2022

## Change Types Made

* [ ] Documentation changes.

## Change Checklist

* [ ] All items are complete on the [Definition of Done](https://github.com/chocolatey/home/blob/main/definition-of-done.md).

## Related Issue

Fixes #3645
PROJ-3152


vexx32:
I discussed the ChocolateyGUI case with Gary and we agreed that this is an acceptable change and ChocolateyGUI will need to fix-forward the edge cases where the Chocolatey API actually throws an exception. I've opened https://github.com/chocolatey/ChocolateyGUI/issues/1155 to track that follow up work.

src/chocolatey.tests/infrastructure.app/validations/CacheFolderLockdownValidationSpecs.cs
```diff
@@ -0,0 +1,307 @@
...
+            }
+        }
+        public sealed class When_not_elevated : CacheFolderLockdownValidationSpecsBase
+        {
+            [SetUp]
+            public void CheckElevation()
+            {
+                if (IsElevated)
+                {
+                    Assert.Ignore();
```
AdmiringWorm: Can we add a message on why it is being ignored (should flow into the test results).

For istance


```suggestion
                    Assert.Ignore("Ignored due to running in elevated mode.");
```

src/chocolatey.tests/infrastructure.app/validations/CacheFolderLockdownValidationSpecs.cs
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        public abstract class ElevatedCacheFolderLockdownValidationSpecsBase : CacheFolderLockdownValidationSpecsBase
+        {
+            [SetUp]
+            public void CheckElevation()
+            {
+                if (!IsElevated)
+                {
+                    Assert.Ignore();
```
AdmiringWorm: Can we passa message to the `Assert.Ignore` so the message can flow into the test results?

For instance:


```suggestion
                    Assert.Ignore("Ignored due to running in non-elevated mode.");
```

src/chocolatey/infrastructure.app/validations/CacheFolderLockdownValidation.cs
```diff
@@ -68,25 +80,60 @@ public ICollection<ValidationResult> Validate(ChocolateyConfiguration config)
...
+                        }
+                        else
+                        {
+                            result.Add(new ValidationResult
+                            {
+                                ExitCode = 0,
+                                Message = $"System Cache directory exists, but could not be locked down to administrators or deleted: {deleteError.Message}.".SplitOnSpace(linePrefix: "   "),
+                                Status = ValidationStatus.Warning
+                            });
+                        }
```
AdmiringWorm, line 128: From the look of it, you have done this under the wrong assumption.

When the `UseHttpCache` is set to false, it isn't really that you aren't using the HTTP Cache location, it is just that it won't re-use any existing HTTP Cache files that was created before the process runs.

But it will still write files to the directories (and re-use the ones that was created during this process).

src/chocolatey.tests/infrastructure.app/validations/CacheFolderLockdownValidationSpecs.cs
AdmiringWorm: Additionally, no single run can cover all of these. Our CI reports 4 skipped out of 1206, so
it is running elevated: the 14 elevated tests run and the 2 non-elevated ones are skipped. A
non-elevated machine gives the reverse, 2 run and 14 skipped.

That means the non-elevated path is never exercised on CI, which is unfortunate given that is
where the no-profile case I raised on `Should_succeed_with_user_cache_directory` lives. No
test added there can cover it while elevation is read from the ambient process.

We should probably need to think of perhaps doing some kind of injection of the elevated (or
allowing it to be overridden in the cache lockdown validation) if we want to properly test it.

src/chocolatey.tests/infrastructure.app/validations/CacheFolderLockdownValidationSpecs.cs
```diff
@@ -0,0 +1,307 @@
...
+            }
+
+            [Fact]
+            public void Should_succeed_with_user_cache_directory()
+            {
+                var result = Results.First();
+                result.Status.Should().Be(ValidationStatus.Success);
+                result.Message.Should().Be("User Cache directory is valid.");
+                result.ExitCode.Should().Be(0);
+            }
```
AdmiringWorm, line 81: But what if there is no user cache in a situation where the running user do not have a user profile on disk?

In that case, this shouldn't be valid since it can't even use a User Cache directory.

src/chocolatey/GetChocolatey.cs
```diff
@@ -349,7 +349,10 @@ public IEnumerable<T> List<T>()
                 (config) =>
                 {
                     var runner = new GenericRunner();
-                    return runner.List<T>(config, _container, isConsole: false, parseArgs: null);
+                    return runner.List<T>(config, _container, isConsole: false, command =>
+                    {
+                        PreRunValidationChecks.Validate(command, config, _container);
+                    });
```
AdmiringWorm, line 356: This could potentially be considered a breaking change from someone calling into the API.

Previously no validations were made on this call and have the potential to break anyone using the API.

I understand it was considered to be fine under the Chocolatey GUI situation, this is still in a call that we consider the public API that anyone using the Chocolatey library can use.

src/chocolatey/infrastructure.app/validations/PreRunValidationChecks.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <param name="config">The current <see cref="ChocolateyConfiguration"/>.</param>
+        /// <param name="container">The currently in-use <see cref="Container"/>.</param>
+        /// <returns>A list of validation results that should be inspected for any secondary failures, to be handled by the caller.</returns>
+        /// <exception cref="ApplicationException">Thrown if the core command and config validation fails.</exception>
+        internal static IList<ValidationResult> GetValidationResults(ChocolateyConfiguration config, Container container)
+        {
+            "chocolatey".Log().Debug(() => "Performing validation checks...");
+            var validationResults = new List<ValidationResult>();
+            var validationChecks = container.GetAllInstances<IValidation>();
+            foreach (var validationCheck in validationChecks)
```
AdmiringWorm, line 44:
```suggestion
            var validationChecks = container.GetAllInstances<IValidation>();

            foreach (var validationCheck in validationChecks)
```

src/chocolatey/infrastructure.app/validations/PreRunValidationChecks.cs
```diff
@@ -0,0 +1,112 @@
...
+                "chocolatey".Log().Info("");
+                var errorMessage = new StringBuilder("Validation Errors:").AppendLine();
+
+                foreach (var error in errors.OrEmpty())
+                {
+                    errorMessage.AppendLine(" - {0}".FormatWith(error.Message));
+                }
+
+                throw new ApplicationException(errorMessage.ToString());
+            }
```
AdmiringWorm, line 110: Previously in the old `ReportValidationSummary`, it logged each error through `Log().Error` (similiar to the previous warning reporting).

Are we sure we only want to have it part of the exception being thrown, and not in the standard output similiar to the warning being logged?

src/chocolatey/infrastructure.app/validations/CacheFolderLockdownValidation.cs
```diff
@@ -68,25 +80,60 @@ public ICollection<ValidationResult> Validate(ChocolateyConfiguration config)
...
-                        ExitCode = 0,
-                        Message = "System Cache directory is not locked down to administrators.\nRemove the directory '{0}' to have Chocolatey CLI create it with the proper permissions.".FormatWith(cacheFolderPath).SplitOnSpace(linePrefix: "   "),
-                        Status = ValidationStatus.Warning
-                    });
+                        _fileSystem.DeleteDirectory(cacheFolderPath, recursive: true);
+                        result.Add(new ValidationResult
+                        {
+                            Message = "The HTTP Cache was not correctly locked down, purging the cache before continuing.".SplitOnSpace(linePrefix: "   "),
+                            Status = ValidationStatus.Warning,
+                        });
```
AdmiringWorm, line 107: You are setting the exit code on every other ValidationResult (even when 0), why not set it on this one as well?

src/chocolatey/infrastructure.app/validations/CacheFolderLockdownValidation.cs
```diff
@@ -68,25 +80,60 @@ public ICollection<ValidationResult> Validate(ChocolateyConfiguration config)
...
+                    }
+                    catch (Exception deleteError)
```
AdmiringWorm, line 109: Catching absolute all exceptions doesn't sound appropriate to me.

Should we maybe just cache the `IOException` and `UnauthorizedAccessException` instead?

src/chocolatey/infrastructure.app/validations/CacheFolderLockdownValidation.cs
```diff
@@ -68,25 +80,60 @@ public ICollection<ValidationResult> Validate(ChocolateyConfiguration config)
...
+                                Message = $"System Cache directory exists, but could not be locked down to administrators or deleted: {deleteError.Message.TrimEnd('.')}. Remove the '{cacheFolderPath}' directory and try again, or provide the '--ignore-http-cache' option to temporarily ignore this error.".SplitOnSpace(linePrefix: "   "),
+                                Status = ValidationStatus.Error
+                            });
+                        }
+                        else
+                        {
+                            result.Add(new ValidationResult
+                            {
+                                ExitCode = 0,
+                                Message = $"System Cache directory exists, but could not be locked down to administrators or deleted: {deleteError.Message}.".SplitOnSpace(linePrefix: "   "),
```
AdmiringWorm: Since you are trimming the ending `.` on the erorr message in the other location, should it maybe be done here as well?


```suggestion
                                Message = $"System Cache directory exists, but could not be locked down to administrators or deleted: {deleteError.Message.TrimEnd('.')}.".SplitOnSpace(linePrefix: "   "),
```

src/chocolatey/infrastructure.app/validations/PreRunValidationChecks.cs
```diff
@@ -0,0 +1,112 @@
...
+{
+    internal static class PreRunValidationChecks
+    {
+        /// <summary>
+        /// Retrieve the full list of validation results from any secondary validations.
+        /// </summary>
+        /// <param name="config">The current <see cref="ChocolateyConfiguration"/>.</param>
+        /// <param name="container">The currently in-use <see cref="Container"/>.</param>
+        /// <returns>A list of validation results that should be inspected for any secondary failures, to be handled by the caller.</returns>
+        /// <exception cref="ApplicationException">Thrown if the core command and config validation fails.</exception>
```
AdmiringWorm, line 37: Is this exception handling a copy paste error from somewhere else?

src/chocolatey/PreRunValidationChecks.cs
```diff
@@ -0,0 +1,112 @@
...
+using chocolatey.infrastructure.commands;
+using chocolatey.infrastructure.logging;
+using chocolatey.infrastructure.validations;
+using SimpleInjector;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace chocolatey
```
AdmiringWorm: Having this class in the root of the application doesn't sound like the right place for it IMO.

Should it maybe be located in the `infrastructure.app.validations` instead?

AdmiringWorm (changes requested):
I have looked through the code here now, and gone through the testing steps documented (only for CLI at the moment).

I believe there are a few things to fix up, and maybe that needs additional consideration.

src/chocolatey.tests/infrastructure.app/validations/CacheFolderLockdownValidationSpecs.cs
vexx32: Yeah, it's certainly not *ideal* :/

I'm not totally against having some funky dependency injection things going on for the ProcessInformation, but that is a much wider-reaching change.

Personally I think the user-mode tests are less important, and not sufficiently important to warrant that degree of change; that cache folder is *always* considered insecure to begin with, which is why it's not used for installations typically, unless they're using Chocolatey in an unsupported configuration to begin with.

src/chocolatey/infrastructure.app/validations/CacheFolderLockdownValidation.cs
vexx32: Yeah, it does write files. It just won't use any files that are already there, so that's sufficient (IMO) to avoid any potential hazards already placed in the cache directory, no? 

Or are you suggesting there's still a vector for some odd timing based attack where a cache file is modified during a `choco` execution after it's created?

src/chocolatey.tests/infrastructure.app/validations/CacheFolderLockdownValidationSpecs.cs
vexx32: Yes, but I don't know how to test this case without refactoring the ProcessInformation class completely, and that would be a bit of a large change here IMO.

And I would *probably* consider a user-mode run where there is no user profile directory at all to be an unsupported / Undefined Behaviour type scenario to begin with.

src/chocolatey/GetChocolatey.cs
vexx32: Yep, it could be. I have discussed this with @gep13, and while we do acknowledge that I think for the API, we agree this is a better behaviour than silently failing out and only emitting to the error log.

It's not *ideal* and it is a change, but it's the lesser of two evils to hand back a coherent exception (especially since this is a pre-run validation, not just something that happens during install), rather than burying validation errors in the log file and (potentially) an application using the API not even knowing about the errors unless they're manually inspecting the log file.

And given these errors will look a bit different to errors normally given back from a choco execution, I'm inclined to not want to rely purely on just emitting the error to the log file and quietly returning as though nothing's wrong.

src/chocolatey/infrastructure.app/validations/PreRunValidationChecks.cs
vexx32: When running choco normally, it does still emit the error to the log file and console. We *could* optionally also have it emit that error to the log file silently, but it will  just result in duplicate log file entries.

src/chocolatey/infrastructure.app/validations/CacheFolderLockdownValidation.cs
vexx32: I think I opted to catch everything here as I don't actually know what the established behavioural expectations for Validations is. From looking at the existing ones, my read was that we don't expect running the validations to be able to throw themselves, only hand back error results, and the throwing (if any) is left to the caller, no?

src/chocolatey/infrastructure.app/validations/PreRunValidationChecks.cs
vexx32: No, that's referring to the `command.Validate()` call, that can throw an ApplicationException as well.

src/chocolatey/PreRunValidationChecks.cs
vexx32: Yeah I wasn't too sure exactly where it should live. It's not *itself* an IValidation so it felt a bit weird to put it in there, but I also don't know if there's any other better spot for it, so. Yeah, it can go in there.

src/chocolatey.tests/infrastructure.app/validations/CacheFolderLockdownValidationSpecs.cs
vexx32: If we want to add a validation for that and fail out in that scenario I can, but I don't see a useful way to test that without totally redesigning ProcessInformation :/

And since it's public API we can't change it completely, I'd have to add something that duplicates the code or references it :/

src/chocolatey/GetChocolatey.cs
vexx32: Also, ref: https://github.com/chocolatey/ChocolateyGUI/issues/1156

Chocolatey's API can already throw exceptions in some cases, so this isn't *completely* new -- although that is an unusual case, it isn't unheard of that the API can throw exceptions.

